PR #21, opened 2019-06-24
Add garage plugin.

c650:
Adds cheerio dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
   "author": "joey",
   "license": "ISC",
   "dependencies": {
+    "@types/cheerio": "^0.22.11",
     "axios": ">=0.18.1",
     "bottlejs": "^1.7.2",
     "cheerio": "^1.0.0-rc.3",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
   "author": "joey",
   "license": "ISC",
   "dependencies": {
-    "@types/cheerio": "^0.22.11",
     "axios": ">=0.18.1",
     "bottlejs": "^1.7.2",
     "cheerio": "^1.0.0-rc.3",
@@ -21,6 +20,7 @@
     "typescript": "^3.4.5"
   },
   "devDependencies": {
+    "@types/cheerio": "^0.22.11",
     "@types/dotenv": "^6.1.1",
     "@typescript-eslint/eslint-plugin": "^1.9.0",
     "@typescript-eslint/parser": "^1.9.0",
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "dependencies": {
     "axios": ">=0.18.1",
     "bottlejs": "^1.7.2",
+    "cheerio": "^1.0.0-rc.3",
     "discord.js": "^11.5.0",
     "dotenv": "^8.0.0",
     "typescript": "^3.4.5"
--- a/src/app/plugins/garage.plugin.ts
+++ b/src/app/plugins/garage.plugin.ts
@@ -1,0 +1,86 @@
+import { Plugin } from '../../common/plugin';
+import { IContainer, IMessage, ChannelType, IHttpResponse } from '../../common/types';
+
+const cheerio = require("cheerio")
+
+class Garage {
+	public name : string = "";
+	public available: number = 0;
+	public saturation: number = 0;
+	public capacity: number = 0;
+	public percent_full: number = 0;
+	public percent_avail: number = 0;
+}
+
+export class GaragePlugin extends Plugin {
+	public name: string = 'Garage Plugin';
+	public description: string = 'Get garage status';
+	public usage: string = 'garage [which garage]';
+	public permission: ChannelType = ChannelType.Public;
+
+	private _API_URL: string = "http://secure.parking.ucf.edu/GarageCount/iframe.aspx";
+	private title_message: string = "**Current UCF Garage Saturation**"
+
+	public process_response(text: String): Garage[] {
+		const $ = cheerio.load(text)
+		const garages = $(".dxgv")
+
+		var last = ""
+		var ret: Garage[] = []
+
+		garages.map((idx: number, elem: Object) => {
+			if (idx % 3 == 0)
+				last = $(elem).text().replace("\n","")
+			else if (idx % 3 == 1) {
+				const tok = $(elem).text().trim().split("/")
+				var gar = {
+					name: last,
+					available: +tok[0],
+					saturation: 0,
+					capacity: +tok[1],
+					percent_full: 0.0,
+					percent_avail: 0.0
+				}
+				gar.percent_avail = 100.0 * gar.available / gar.capacity
+				gar.saturation = gar.capacity - gar.available
+				if (gar.saturation < 0) gar.saturation = 0
+				gar.percent_full = 100.00 * gar.saturation / gar.capacity
+				ret.push(gar)
+			}
+		})
+
+		return ret
+	}
+
+	constructor(public container: IContainer) {
+		super();
+	}
+
+	public validate(message: IMessage, args: string[]): boolean {
+		return true;
+	}
+
+	public hasPermission(message: IMessage): boolean {
+		const channelName = this.container.messageService.getChannel(message).name;
+		return this.container.channelService.hasPermission(channelName, this.permission);
+	}
+
+	public execute(message: IMessage, args?: string[]): void {
+		this.container.httpService
+			.get(`${this._API_URL}`)
+			.then((response) => {
+				const garages: Garage[] = this.process_response(response.data)
+				var str = ""
+
+				garages.map((elem : Garage) => {
+					str += `${elem.name.replace("Garage ", "")}:`.padStart(6, ' ')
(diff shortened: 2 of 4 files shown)


src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export class GaragePlugin extends Plugin {
+	public name: string = 'Garage Plugin';
+	public description: string = 'Get garage status';
+	public usage: string = 'garage [which garage]';
+	public permission: ChannelType = ChannelType.Public;
+
+	private _API_URL: string = "http://secure.parking.ucf.edu/GarageCount/iframe.aspx";
+	private title_message: string = "**Current UCF Garage Saturation**"
```
joey-colon: The pattern with private variables/methods in this project is always start with a leading `_`. We also use semicolons throughout the codebase. Let's keep consistent to that.

src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,86 @@
+import { Plugin } from '../../common/plugin';
+import { IContainer, IMessage, ChannelType, IHttpResponse } from '../../common/types';
+
+const cheerio = require("cheerio")
```
joey-colon: You should install the cheerio types located [here](https://www.npmjs.com/package/@types/cheerio) and utilize `import * as cheerio from 'cheerio';` instead.

src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,86 @@
...
+	public permission: ChannelType = ChannelType.Public;
+
+	private _API_URL: string = "http://secure.parking.ucf.edu/GarageCount/iframe.aspx";
+	private title_message: string = "**Current UCF Garage Saturation**"
+
+	public process_response(text: String): Garage[] {
+		const $ = cheerio.load(text)
+		const garages = $(".dxgv")
+
+		var last = ""
```
joey-colon: We typically never use `var` when declaring a variable. A good rule of thumb is to always use `const` unless you are actually modifying the value of the variable. In that situation, you'll want to be using `let` instead. I recommend you to download ESLint and Prettier and setup the linters to apply the lints on-save of file. It will make everything flow smoother.

src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,86 @@
...
+		var ret: Garage[] = []
```
joey-colon: nit: can we use a more meaningful variable name?

src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,86 @@
...
+
+	public hasPermission(message: IMessage): boolean {
+		const channelName = this.container.messageService.getChannel(message).name;
+		return this.container.channelService.hasPermission(channelName, this.permission);
+	}
+
+	public execute(message: IMessage, args?: string[]): void {
+		this.container.httpService
+			.get(`${this._API_URL}`)
+			.then((response) => {
```
joey-colon: Update this to cast `IHttpResponse`, so `(response: IHttpResponse)`. I will have to update the dog plugin accordingly as well

src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,86 @@
...
+
+	public process_response(text: String): Garage[] {
+		const $ = cheerio.load(text)
+		const garages = $(".dxgv")
+
+		var last = ""
+		var ret: Garage[] = []
+
+		garages.map((idx: number, elem: Object) => {
+			if (idx % 3 == 0)
```
joey-colon: Good practice to always compare using `===` instead of `==`.

src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,86 @@
...
+				last = $(elem).text().replace("\n","")
+			else if (idx % 3 == 1) {
+				const tok = $(elem).text().trim().split("/")
```
joey-colon: nit: rename to `token`

src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,86 @@
...
+				var gar = {
```
joey-colon: nit: rename to `garage`

src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,86 @@
...
+export class GaragePlugin extends Plugin {
+	public name: string = 'Garage Plugin';
+	public description: string = 'Get garage status';
+	public usage: string = 'garage [which garage]';
+	public permission: ChannelType = ChannelType.Public;
+
+	private _API_URL: string = "http://secure.parking.ucf.edu/GarageCount/iframe.aspx";
+	private title_message: string = "**Current UCF Garage Saturation**"
+
+	public process_response(text: String): Garage[] {
```
joey-colon: This method should be private.

src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,86 @@
+import { Plugin } from '../../common/plugin';
+import { IContainer, IMessage, ChannelType, IHttpResponse } from '../../common/types';
+
+const cheerio = require("cheerio")
+
+class Garage {
+	public name : string = "";
```
joey-colon: The spacing pattern is 2 spaces per tab, linter will update this on save.

src/app/plugins/garage.plugin.ts
c650: `import * as cheerio from 'cheerio'` is not recommended. https://github.com/cheeriojs/cheerio#loading

joey-colon: This is assuming that the user is using javascript. We should remain consistent and use the latter import

package.json
```diff
@@ -12,8 +12,10 @@
   "author": "joey",
   "license": "ISC",
   "dependencies": {
+    "@types/cheerio": "^0.22.11",
```
joey-colon: We want types to be a `devDependencies`

src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,116 @@
...
+  public permission: ChannelType = ChannelType.Public;
+
+  private _API_URL: string = 'http://secure.parking.ucf.edu/GarageCount/iframe.aspx';
+  private _TITLE_MSG: string = '**Current UCF Garage Saturation**';
+
+  private _GARAGE_UPD_THRESH: number = 1000 * 60 * 2; // in ms, two minutes.
+  private _LAST_UPD_TIME: number = -this._GARAGE_UPD_THRESH - 1;
+  private _GARAGES: Garage[] = [];
+
+  private process_response(text: String): Garage[] {
```
joey-colon: camelCase with a leading `_`

src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,116 @@
...
+    const $ = cheerio.load(text);
+    const garages = $('.dxgv');
+
+    let last = '';
+    let processed_garages: Garage[] = [];
```
joey-colon: const

src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,116 @@
...
+        garage.percent_full = (100.0 * garage.saturation) / garage.capacity;
+
+        processed_garages.push(garage);
+      }
+    });
+
+    return processed_garages;
+  }
+
+  private async get_garages(): Promise<Garage[]> {
```
joey-colon: camelCase with a leading `_`

src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    let last = '';
+    let processed_garages: Garage[] = [];
+
+    garages.map((idx: number, elem: Object) => {
+      if (idx % 3 === 0)
+        last = $(elem)
+          .text()
+          .replace('\n', '');
+      else if (idx % 3 == 1) {
```
joey-colon: `===`

package.json
c650: This was automatically edited file.

joey-colon: Uninstall the package utilizing `npm uninstall @types/cheerio` and then re-install it using the `--save-dev` flag. `npm install @types/cheerio --save-dev`

src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,116 @@
...
+        const token = $(elem)
+          .text()
+          .trim()
+          .split('/');
+        let garage = {
```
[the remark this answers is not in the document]
c650: I didn't realise you could mutate members when the object is `const`...

src/app/plugins/garage.plugin.ts
```diff
@@ -0,0 +1,117 @@
+import { Plugin } from '../../common/plugin';
+import { IContainer, IMessage, ChannelType, IHttpResponse } from '../../common/types';
+
+// const cheerio = require('cheerio');
```
joey-colon: remove

PR #28, opened 2019-07-25
Class registration system

joey-colon:


src/common/plugin.ts
```diff
@@ -20,5 +20,5 @@ export abstract class Plugin implements IPlugin {
     return this.container.channelService.hasPermission(channelName, this.permission);
   }
 
-  public abstract async execute(message: IMessage): Promise<void>;
+  public abstract async execute(message: IMessage, args?: string[]): Promise<void>;
```
c650: What does this do?

src/services/class.service.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (classType === ClassType.ALL) {
+      return new Map<string, GuildChannel>([
+        ...this.getClasses(ClassType.CS),
+        ...this.getClasses(ClassType.IT),
+      ]);
+    }
+    return this._channels.get(classType) || new Map<string, GuildChannel>();
+  }
+
+  addClasses() {
```
c650: nit: can you specify void return type?

c650:
I don't know how I feel about the way this is getting implemented. I think that classes should be defined in a static file, like JSON or something. This way, each semester only the file has to be changed, and moderators don't have to go through channel deletion and creation N times.

src/common/plugin.ts
joey-colon: Nothing. just wanted to be more clear on the method for when we extend the class (never had all the correct parameters in this method for some reason (?))

joey-colon:
> I don't know how I feel about the way this is getting implemented. I think that classes should be defined in a static file, like JSON or something. This way, each semester only the file has to be changed, and moderators don't have to go through channel deletion and creation N times.

This seems to be a separate issue with how we should be creating class channels. I agree though. Even with a JSON file of class names, I don't believe the registration plugin's logic would change. We still have to grab a reference to the Discord channel object associated with the requested class name (so we can modify the channel's permissions).

If we do the JSON file approach, we can fall into the scenario where we happened to miss a class or someone requested for a class to be added which would cause me to log into the server and update the JSON accordingly which seems cumbersome.

Maybe we can implement some exclusive command where we can maintain class list?
`!addclass 1234_class_prefix`
`!removeclass 1234_class_prefix`

I would like a solution where it does not require me to update a local static file.

joey-colon:
@c650 we can complete a system for managing channels in the future. this current implementation won't require much change when we do add it

Issue #42, opened 2019-08-18
Lexicographically sort classes and roles


c650:
It’s easier to digest the large lists when there is an ordering.

PR #43, opened 2019-08-18
Sort classes, roles

c650:
Adjust code so that !listroles and !listclasses output the results in sorted order.
This makes it easier for users to locate the class they want (in order to verify its presence).

Also removed unneeded import in listroles.plugin.ts and added import of Role from discord.js


c650:
#42

joey-colon:
lgtm

Issue #38, opened 2019-08-17
!listclasses extensibility 


joey-colon:
Support both `!listclasses cs` and `!listclasses it`

PR #44, opened 2019-08-19
Adds capability to filter listclasses with cs/it arg

c650:
Presents full list if argument is invalid

Addresses #38

### DIFF
--- a/src/app/plugins/listclasses.plugin.ts
+++ b/src/app/plugins/listclasses.plugin.ts
@@ -24,8 +24,8 @@ export class ListClassesPlugin extends Plugin {
       badFilterParam = true;
     }
 
-    const listCsClasses = filter == 'CS' || filter == 'ALL';
-    const listItClasses = filter == 'IT' || filter == 'ALL';
+    const listCsClasses = filter === 'CS' || filter === 'ALL';
+    const listItClasses = filter === 'IT' || filter === 'ALL';
 
     if (listCsClasses) {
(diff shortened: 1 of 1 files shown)


src/app/plugins/listclasses.plugin.ts
```diff
@@ -16,28 +16,45 @@ export class ListClassesPlugin extends Plugin {
     const itClasses = this.container.classService.getClasses(ClassType.IT);
     let response = '```\n';
 
-    response += 'CS Classes:\n';
+    let listCsClasses = !args || args.length == 0 || args[0].toLowerCase() == 'cs';
+    let listItClasses = !args || args.length == 0 || args[0].toLowerCase() == 'it';
```
joey-colon: Utilize `ClassType` enum instead of string literals

src/app/plugins/listclasses.plugin.ts
```diff
@@ -16,28 +16,38 @@ export class ListClassesPlugin extends Plugin {
...
-      csClassNames.push(classObj[1].name);
+    if (filter != ClassType.CS && filter != ClassType.IT && filter != ClassType.ALL) {
+      filter = 'ALL';
+      badFilterParam = true;
     }
 
-    csClassNames.sort().forEach((classObjname) => {
-      response += `${classObjname}\n`;
-    });
+    const listCsClasses = filter == 'CS' || filter == 'ALL';
```
joey-colon: compare to enumerations
`filter === ClassType.CS`

Issue #33, opened 2019-08-16
Prettify !help


joey-colon:
refer to lion v1's help styling (use of discord embeds)

PR #49, opened 2019-08-30
Added embed to help command.

gooca:


src/app/plugins/help.plugin.ts
```diff
@@ -1,38 +1,39 @@
...
-  private _commandsResponse: string = '';
+  private _embed: IRichEmbed = new RichEmbed;
   constructor(public container: IContainer) {
     super();
   }
 
   public async execute(message: IMessage, args?: string[]) {
-    if (this._commandsResponse.length === 0) {
-      this._generateCommandsResponse();
+    if (this._embed.length === 0) {
```
gooca: idk if checking the length of an object is supposed to work like correctly or with this syntax, but it works fine when I tested it. Not sure if this is what it should be doe

joey-colon: You would probably want to be checking the length of the `fields` property which should contain an array of the fields within the RichEmbed object

src/app/plugins/help.plugin.ts
```diff
@@ -1,38 +1,39 @@
...
-        this._commandsResponse += '\n';
-      }
-
-      this._commandsResponse += `**${plugin.name}**`;
-      this._commandsResponse += ` - `;
-      this._commandsResponse += `${plugin.description}`;
-      this._commandsResponse += ` - `;
-      this._commandsResponse += `\`${Constants.Prefix}${plugin.usage}\``;
+      this._embed
+      .addField(
```
joey-colon: nit: indent

joey-colon:
Resolves #33

joey-colon:
lgtm 🐶

PR #62, opened 2019-10-17
Whitelist clubs from bot_reports.

gooca:
international kingkiller

### DIFF
--- a/src/app/handlers/blacklist.handler.ts
+++ b/src/app/handlers/blacklist.handler.ts
@@ -1,10 +1,17 @@
 import { IHandler, IMessage, IContainer } from '../../common/types';
+import Constants from '../../common/constants';
+import { TextChannel } from 'discord.js';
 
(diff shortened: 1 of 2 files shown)


gooca:
ok that was a little cancer i didnt realize i had to cast message.channel

src/app/handlers/blacklist.handler.ts
```diff
@@ -1,11 +1,18 @@
...
   private _expressions: RegExp[] = [/discord.gg/, /group\s?me/];
+  private _whitelistedChannels = new Set([Constants.Channels.Public.Clubs]);
   constructor(public container: IContainer) {}
 
   public async execute(message: IMessage): Promise<void> {
     this._expressions.forEach((expression) => {
+      const channel = message.channel as TextChannel;
+      if (this._whitelistedChannels.has(channel.name)) {
+        return;
+      }
```
c650: this should happen right at the beginning, not in a loop.

PR #81, opened 2019-12-11
Cleanup

joey-colon:


HoriKyouko:
To fix the DeepScan issue you need another = on line 90

PR #87, opened 2019-12-29
Add in crypto currency support to !stock

brettfazio:


src/app/plugins/stock.plugin.ts
```diff
@@ -101,6 +172,28 @@ export class StockPlugin extends Plugin {
     return embed;
   }
 
+  private _makeCryptoEmbed(quote: any) {
```
joey-colon: ideally, you would want to expand on `_makeEmbed()` instead of making a whole new method. maybe passing in an enum for what type of embed could be a better pattern at tackling this

joey-colon:
we also wanted to rename this plugin from `stock` to `price`

brettfazio:
Changed it from `stock` to `price` and then removed the extra method and used enums.

PR #96, opened 2020-01-03
Implement support for variable class registration

keenanturley:
Note: This is my first PR and first time using TypeScript, so I'm open to criticism.

I wanted to be able to register for all my classes at once using a format like `/register class1 class2 class3`, so I tried to implement that. 

If invalid classes are provided among the arguments they will be reported to the user along with the classes that were successfully registered.

One thing I'm not sure about is constructing the response, mainly because I haven't worked with JS strings.

### DIFF
--- a/src/app/plugins/register.plugin.ts
+++ b/src/app/plugins/register.plugin.ts
@@ -1,5 +1,5 @@
 import { Plugin } from '../../common/plugin';
-import { IContainer, IMessage, ChannelType, ClassType } from '../../common/types';
+import { IContainer, IMessage, ChannelType, ClassType, ClassResponseType } from '../../common/types';
 
 export class RegisterPlugin extends Plugin {
   public name: string = 'Register Plugin';
@@ -26,12 +26,18 @@ export class RegisterPlugin extends Plugin {
         return;
       }
       try {
-        const response = await this.container.classService.register(request);
-        message.reply(response);
+        const responseType: ClassResponseType = await this.container.classService.register(request);
+        if (responseType === ClassResponseType.Success) {
+          message.reply(`You have successfully been added to the ${request.categoryType} category.`);
+        } else {
+          message.reply(`An error occured while attempting to add you to the ${request.categoryType} category.`);
+        }
       } catch (e) {
         message.reply(e);
       }
     } else {
+      let successfulRequests: string[] = [];
+      let failedRequests: string[] = [];
       for (let arg of args) {
         const request = this.container.classService.buildRequest(message.author, [ arg ]);
         if (!request) {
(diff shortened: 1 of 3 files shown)


keenanturley:
Also, I did only tested running the application, not actually using the command on a server.

keenanturley:
I changed the response code I was worried about to something that seems more logical, so I'm happy with it now. 39c0ab1

src/services/class.service.ts
```diff
@@ -22,20 +22,32 @@ export class ClassService {
...
   async register(request: IClassRequest): Promise<string> {
-    const { author, categoryType, requestType, className } = request;
+    const { author, categoryType, requestType, classNames } = request;
     try {
       if (categoryType === null) {
         // Since category types are only dealt with registration of categories,
-        // we are handling an individual class registration.
-        if (!className) {
-          throw new Error('No class name detected');
+        // we are handling a varialbe class registration.
```
joey-colon: nit: spelling error

joey-colon:
there's nothing glaringly wrong with your code implementation of this - nice job for a first timer :). the "issue" i'm thinking about is whether or not we want to push the responsibility of the service to handle multiple class names in one request or make it the plugin's responsibility to handle multiple classes passed into a message and building multiple requests. if we chose the latter, we would not need to modify the implementation of the class service's add/build request function which will probably lead to fewer lines being added/changed.

i'd like to hear @c650 thoughts on this

c650:
> if we chose the latter, we would not need to modify the implementation of the class service's add/build request function which will probably lead to fewer lines being added/changed.

Let’s modify the plugin, not the service. Simply loop over args to do multiple classes.

keenanturley:
I reverted and re-implemented the feature using only the plugin as requested.

There are a few consequences to this such as multiple bot messages sent for the single multi-registration command, however preventing this would require checking against the response strings of the register method, which seemed error-prone if it was changed in the future. If something more than a string was returned indicating validity of the class it would be better.

joey-colon:
> I reverted and re-implemented the feature using only the plugin as requested.
> 
> There are a few consequences to this such as multiple bot messages sent for the single multi-registration command, however preventing this would require checking against the response strings of the register method, which seemed error-prone if it was changed in the future. If something more than a string was returned indicating validity of the class it would be better.

I don't think you need to be sending out multiple messages in the event the user is requesting multiple classes to be registered.

Maybe instead of returning a string for the service's register request, we can leverage some sort of enum to indicate the request's status? Then for each of the requests we execute (assuming we're executing multiple requests), we push it to a failedRequest/successfulRequest arrays depending on what the status code response was, similar to what you had before.

src/app/plugins/register.plugin.ts
```diff
@@ -16,16 +16,35 @@ export class RegisterPlugin extends Plugin {
   }
 
   public async execute(message: IMessage, args?: string[]) {
```
c650: let's re-work this to use some regex to tell between an argument that is a classchannel and an arg that is a category

src/app/plugins/register.plugin.ts
```diff
@@ -16,16 +16,35 @@ export class RegisterPlugin extends Plugin {
...
+      } catch (e) {
+        message.reply(e);
+      }
+    } else {
+      for (let arg of args) {
+        const request = this.container.classService.buildRequest(message.author, [ arg ]);
+        if (!request) {
+          message.reply('I was unable to build your request.');
+          return;
+        }
```
c650: makes too many respondes

keenanturley: If you mean that that it replies with `I was unable to build your request.` too many times, it should stop after the first because of the return statement immediately after. Let me know if I'm misunderstanding you.

src/app/plugins/register.plugin.ts
keenanturley: I don't have much experience with regex. If you have a specific regex that applies I can implement that.

src/app/plugins/register.plugin.ts
c650: hmm, that also would be bad since user should get registered for any valid listed classes

joey-colon:
i think this should be on a slight pause while we take care of #100 in case some sort of implementation changes where this branch gets effected

c650:
How are we doing on this?

c650:
closed for stale

PR #97, opened 2020-01-07
todos

joey-colon:


src/app/plugins/price.plugin.ts
```diff
@@ -21,12 +23,12 @@ export class PricePlugin extends Plugin {
     up: {
       thumbnail_url:
         'https://www.netclipart.com/pp/m/59-594517_arrow-going-up-png-stock-market-graph-up.png',
-      color: '#00ff00',
+      color: '#a3be8c',
```
c650: y

joey-colon: ur color... WACK

c650: my feelings... HURT

src/app/listener.ts
```diff
@@ -1,19 +1,20 @@
...
-    this.container.clientService.on('channelDelete', () => {
-      this.container.classService.updateClasses();
+    this.container.clientService.on('channelDelete', async () => {
+      this._executeHandlers(this._channelHandlers);
     });
-    this.container.clientService.on('channelUpdate', () => {
-      this.container.classService.updateClasses();
+    this.container.clientService.on('channelUpdate', async () => {
+      this._executeHandlers(this._channelHandlers);
     });
```
c650: why is this abstraction necessary?

src/app/listener.ts
```diff
@@ -25,19 +26,27 @@ export class Listener {
...
+    });
+  }
+
+  private async _executeHandlers(handlers: IHandler[], message?: IMessage) {
+    handlers.forEach(async (handler: IHandler) => {
+      try {
+        await handler.execute(message);
+      } catch (e) {
+        console.error(e);
+      }
```
c650: why not just let Promises do their thing here?

src/app/listener.ts
joey-colon: in case we want to ever have any sort of other events to react upon channel events. `listener.ts` shouldnt ever be responsible for invoking `updateClasses()` directly imo, it should just be responsible for running the `execute()` method for all instances that are within our `channelHandlers` (type `IHandler`) array (which will include the classes handler)

src/app/listener.ts
joey-colon: we shud still await here since im not calling await on `_executeHandlers` i think

c650:
lgtm

c650:
merge?

PR #112, opened 2020-03-03
Updated Docker Image

Helithumper:
The old docker image (node:latest) is not best practice as it

a.) Doesn't pin the NodeJS version. If NodeJS were to update, we would automatically update along with it. If there were some sort of breaking change to a newer version of NodeJS, our application might stop working.

b.) `node:latest` is a large image. Once changed to `node:13.8-alpine`, the image shrunk from 1.13GB to 305MB. about 1/3 of the original size.


Helithumper:
Also added a `docker-compose.yml` file for easier development :)

docker-compose.yml
```diff
@@ -0,0 +1,6 @@
+version: '3'
+services:
+  lion:
+    build: .
+    env_file:
+     - .env
```
joey-colon: new line

Helithumper: fixed

c650:
squash and merge because DeepScan issue is my fault :) (uncaught from before DS was incorporated)

PR #121, opened 2020-05-07
Cat API Implementation

CallumCoy:
This adds the Plugin for cat images , like the dog Plugin. Currently available features are:

Post random cat picture.
Post cat picture based off of breed.
Post up to date list of available breeds.
Send 3 pictures at once (bomb command).

It is also set up to do, it connot do these task right now because the API does not yet support the feature:

Post a random cat fact.
Post cat fact based upon breed.


CallumCoy:
I'll fix the coding style that the linter pointed out tonight.

c650:
> Send 3 pictures at once (bomb command).

Not a fan of this. Too spammy.

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+export class CatPlugin extends Plugin {
+  public name: string = 'Cat Plugin';
+  public description: string = 'Generates pictures of cats, via popular command.';
+  public usage: string = 'cat breed or pic bomb(optional) <breed (optional)>';
+  public permission: ChannelType = ChannelType.Public;
+  public pluginChannelName: string = Constants.Channels.Public.Pets;
+
+  private _API_URL: string = 'https://api.thecatapi.com/v1/';
+  private _breeds: string[] = [];
+  private _breedsID: string[] = [];
```
c650, line 14: These are related, combine them into one object ...

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+    this.container.httpService
+      .get(`${this._API_URL}breeds`)
+      .then((response: IHttpResponse) => {
+        const breeds = response.data;
+
+        for (let index = 0; index < breeds.length; index++) {
+            const breed = breeds[index].name;
+            const breedID = breeds[index].id;
+            this._breeds.push(breed.toLowerCase())
+            this._breedsID.push(breedID.toLowerCase())
```
c650, line 27: Instead, keep each breed as a pair of a name and an ID to strongly associate each name with its ID.

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+        //     .then((response: IHttpResponse) => {
+
+        //         message.reply('', {
+        //             files: [response.data[0].text],
+        //         });
+        //     })
+
+        //     .catch((err) => console.log(err));
+
+        // }
```
c650, line 79: Clutter. Please remove.

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    } else if (command == "breeds" || command == "breed") {
```
c650, line 81: `===`

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+        
+        //Simply return the list of suported breeds
+        message.reply(`Breeds supported: \n\`\`\`\n${this._breeds.join('\n')}\`\`\``);
+        return;
+
+    } else if (command == "breads" || command == "bread") {
+
+        // the Easter Egg, my gawd
+        message.reply('', {
+            files: [`https://www.thetoychronicle.com/wp-content/uploads/2016/05/Box-cat-Bread-cat-Rato-workroom-Kim-.jpg`],
```
c650, line 91: nit: un-magic this, make it a well-named constant at the top.

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+
+        command = command.replace("pic", "");
+        var breed = "";
+        var count = 1;
+
+        if (command.includes("bomb")){
+            //processes that the user wants 3 pictures
+            count = 3;
+            command = command.replace("bomb", "");
+        }
```
c650, line 105: please no

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+        return;
+
+    } else if (command == "breads" || command == "bread") {
+
+        // the Easter Egg, my gawd
+        message.reply('', {
+            files: [`https://www.thetoychronicle.com/wp-content/uploads/2016/05/Box-cat-Bread-cat-Rato-workroom-Kim-.jpg`],
+        });
+
+    } else if (command.includes("pic")){
```
c650, line 94: `.includes()` should really be some prefix match right?

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+        // lets the system know that pic has been determined
+
+        command = command.replace("pic", "");
+        var breed = "";
+        var count = 1;
```
c650, line 99: `var` is bad!

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+            //finds the breed
+            const index = this._breeds.findIndex((curBreed) => {
+                return command.includes(curBreed);
+            })
+
+            console.log(index)
+
+            // checks if the bread was found or not
+            if (index == -1) {
+                message.reply(`I could not find that breed, take this instead.  Please type, "cat breed" to get a list of compatible breeds.`);
```
c650, line 119: Only one space after period(s)...

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+            command = command.replace("bomb", "");
+        }
+
+        // determines that their was aditional text added on
+        if (command.trim() != "") {
+
+            //finds the breed
+            const index = this._breeds.findIndex((curBreed) => {
+                return command.includes(curBreed);
+            })
```
c650, line 113: ;

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+        
+        //recieves the according info and posts, or derps
+        await this.container.httpService
+        .get(`${this._API_URL}images/search?limit=${count}${breed}`)
+        .then((response: IHttpResponse) => {
+            for (let index = 0; index < count; index++) {
+                message.reply('', {
+                    files: [response.data[index].url],
+                });
+            }
```
c650, line 133: Unnecessary loop because only one pic should go out per invocation of the "cat" command.

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+        })
+        .catch((err) => console.log(err));
```
c650, line 135: nit: `.catch(console.log)`

src/app/plugins/cat.plugin.ts
CallumCoy: I personally prefer the includes, as this allows the user to ignore order.  Although if you want it changing to that I can change it to that.

src/app/plugins/cat.plugin.ts
CallumCoy: Not too sure where I am missing a semi-colon here,  although in my current tidying up I have added the semi-colons that I missed. Didn't realise that TypeScript wasn't checking for semi-colons so admittedly I got lazy checking for them.

keenanturley: Line 113, after the )

CallumCoy:
Newest merge request, includes all fixes except include (I personally prefer it over prefix).  Hit me with your next review session.

Quick review of what has changed:

Plugin no longer supports the bomb command.
Clean enough to pass the linter test.
Acted upon all comments witht he exception of one.

c650:
>  include (I personally prefer it over prefix)

I disagree

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+
+            this._cats.push(breedInfo);
+        }
+      })
+      .catch((err) => console.log(err));
+  }
+
+  public async execute(message: IMessage, args?: string[]) {
+    
+    let command = this._parseCommand(args || []);
```
joey-colon, line 43: `command` should be `const`

joey-colon: i would also rename `command` to `breed`, refer to subsequent comments

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+    
+    if (command === "breeds" || command === "breed") {
```
joey-colon, line 45: before transforming `args` into a string, we can do a sanity check to check if `args[0] === 'breeds'`, if it returns false, then we can build the breed string

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+        let reply = "";
+
+        this._cats.forEach(cat => {
+            reply = reply.concat(cat.breed, "\n");
+        });
+
+        message.reply(`Breeds supported: \n\`\`\`\n${reply}\`\`\``);
+        return;
+
+    } else if (command === "breads" || command === "bread") {
```
joey-colon, line 57: remove

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+
+        // the Easter Egg, my gawd
+        message.reply('', {
+            files: [this._breadCat],
+        });
+
+    } else if (command.includes("pic")) {
```
joey-colon, line 64: if we hit this case, we're assuming that its already a picture. the user should only be running either `!cat bombay` or `!cat breeds` and we already handled the latter earlier on

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+        // lets the system know that pic has been determined
+
+        command = command.replace("pic", "");
+        let breed: string = "";
+
+        // determines that their was aditional text added on
+        if (command.trim() != "") {
+
+            //finds the breed
+            this._cats.forEach(cat => {
```
joey-colon, line 74: we should be using `.find()` here.

`const breedEntry = this._cats.find((cat) => cat.breed === command);` 

and if it doesnt exist, `breedEntry` would be `undefined`

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+                if (command.includes(cat.breed)) {
+                    breed = cat.id;
+                    return;
+                }
+            });
+
+            // checks if the bread was found or not
+            if (breed === "") {
+                message.reply(`I could not find that breed. Have this picture instead. Please type, "cat breed" to get a list of compatible breeds.`);
```
joey-colon, line 83: correct me if im wrong, but this will be both sending the message and replying with a picture. either allow the command `!cat` to be valid or send this message and don't send any picture when the user doesn't specify a breed

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+        
+        //recieves the according info and posts, or derps
+        await this.container.httpService
+        .get(`${this._API_URL}images/search?limit=1${breed}`)
+        .then((response: IHttpResponse) => {
+            message.reply('', {
+                files: [response.data[0].url],
+            });
+        })
+        .catch(console.log);
```
joey-colon, line 97: keep your `.catch()` blocks consistent if youre doing the same logic. either use `.catch(console.log(err))` or `.catch(console.log)`, you have both throughout the code

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+            //finds the breed
+            this._cats.forEach(cat => {
+                if (command.includes(cat.breed)) {
+                    breed = cat.id;
+                    return;
+                }
+            });
+
+            // checks if the bread was found or not
+            if (breed === "") {
```
joey-colon, line 82: i suggest asserting that str length === 0

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+        });
+
+    } else if (command.includes("pic")) {
+        // lets the system know that pic has been determined
+
+        command = command.replace("pic", "");
+        let breed: string = "";
+
+        // determines that their was aditional text added on
+        if (command.trim() != "") {
```
joey-colon, line 71: this check won't be needed when you implement `.find()` below

src/app/plugins/cat.plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export class CatPlugin extends Plugin {
+  public name: string = 'Cat Plugin';
+  public description: string = 'Generates pictures of cats, via popular command.';
+  public usage: string = 'cat breed or pic bomb(optional) <breed (optional)>';
+  public permission: ChannelType = ChannelType.Public;
+  public pluginChannelName: string = Constants.Channels.Public.Pets;
+
+  private _API_URL: string = 'https://api.thecatapi.com/v1/';
+  private _breadCat: string = `https://www.thetoychronicle.com/wp-content/uploads/2016/05/Box-cat-Bread-cat-Rato-workroom-Kim-.jpg`;
```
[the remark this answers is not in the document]
CallumCoy: I can get rid of the lines 14 and 15 since they have a default, and 18 as that will just remove a fun little feature, but don't we need the _API_URL?

src/app/plugins/cat.plugin.ts
CallumCoy: Don't see why we would remove the part that that gives back a list of breeds.  If you're worried about the length of the list we can look at doing an emote scroll style list.

src/app/plugins/cat.plugin.ts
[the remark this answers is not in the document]
joey-colon: oh yeah, i was specifically referring to only line 18

src/app/plugins/cat.plugin.ts
CallumCoy: nevermind I got you I think single line problems show 4 lines for some reason.

CallumCoy:
newest commmit has implemented the list of requests/ pointers.

removed:
no more !cat bread
no longer sends random picture on invalid breeds

Added:
Pic is no longer a keyword just type cat <breed>
For a random cat pic !cat ran has been implented.

cat.plugin.ts
```diff
@@ -0,0 +1,85 @@
...
+                        name: breedData.name.toLowerCase(),
+                        id: breedData.id.toLowerCase(),
+                    }
+                })
+            })
+            .catch(console.log);
+    }
+
+    public async execute(message: IMessage, args?: string[]) {
+        if (args != undefined && (args[0] === "breeds" || args[0] === "breed")) {
```
joey-colon, line 39: i would change to check if `args && args[0].includes('breed')`

cat.plugin.ts
```diff
@@ -0,0 +1,85 @@
...
+            //Simply return the list of suported breeds
+            let reply = "";
+
+            this._breeds.forEach(breed => {
```
joey-colon, line 43: this can just be `const reply = this._breeds.join('\n');`

cat.plugin.ts
```diff
@@ -0,0 +1,85 @@
...
+        let searchCom = "";
+
+        // checks if their was a bread was a breed, then if that breed is recognised
+        if (breedIn.length != 0) {
+
+            const breedEntry = this._breeds.find((breed) => breed.name === breedIn);
+
+            if (breedEntry != undefined) {
+                searchCom = breedEntry.id;
+            } else if (breedIn != "ran") {
```
joey-colon, line 63: i lack context on this. why would `breedIn` be equal `ran`?

CallumCoy: Not gonna lie took me a few minutes to remember, it was for a random picture.  I have now renamed it to random.

cat.plugin.ts
CallumCoy: not sure why reply would be "this._breeds"  as that is the list of breeds that are available, I believe you meant "breed.name" which is the breed that the user requested.  However, unless I get rid of the random feature this is the neatest way to do it.  If their is a reason that you do not want  me to use let I can change it.

cat.plugin.ts
CallumCoy: I can't do the args by itself, as the program feaks out if args is undefined, however I have changed the later part to include.

CallumCoy: Changed the plan for this, changed it to that, and I just decline ann undefined items/ empty lists prior to this line.

CallumCoy:
1 fix, 3 updates, 1 removal.

Fixed problem where breeds was giving back the wrong breed.

Updated the ran command to random, since even I forgot what it meant.
Updated the args so if it is undefined or empty it is instantly thrown out.
Updated_parseCommand only takes args no as "args || []" had become redundant due to an previous update.

Removed a foreach loop that was not needed, and changed its variable type to const from let.

c650:
Wait why does this have a cat.plugin.ts at the root

PR #125, opened 2020-05-28
Mod Report Plugin

c650:
You can add a report or list all reports on a User.
The users are tracked by Guild and User ID.


c650:
I agree about time stamps. I’ll add that to this PR. 

Regarding persistence, I think it’s easy enough to use mongoDB on the same server.

Issue #126, opened 2020-05-30
Mod report command args


c650:
I forgot names can have spaces, so assuming that arg[1] is the name is not always correct

PR #129, opened 2020-06-05
Patch #126

c650:
Arg parsing caused command to fail when username had spaces (so not just arg[1])

### DIFF
--- a/src/app/plugins/modreport.plugin.ts
+++ b/src/app/plugins/modreport.plugin.ts
@@ -1,0 +1,52 @@
+import Constants from '../../common/constants';
+import { Plugin } from '../../common/plugin';
+import { ChannelType, IContainer, IHttpResponse, IMessage } from '../../common/types';
+import { Report } from '../../services/report.service';
+
+export class ModReportPlugin extends Plugin {
+  public name: string = 'Mod Report Plugin';
+  public description: string = 'add a report against a user';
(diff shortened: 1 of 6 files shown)


src/app/plugins/modreport.plugin.ts
```diff
@@ -10,11 +10,25 @@ export class ModReportPlugin extends Plugin {
   public permission: ChannelType = ChannelType.Staff;
   public pluginChannelName: string = Constants.Channels.Staff.UserOffenses;
 
+  private _commadRegex: RegExp = /(add|list)\s+([^#]+#\d{4})\s*(.*)/;
```
joey-colon, line 13: nit- typo in name

src/app/plugins/modreport.plugin.ts
```diff
@@ -10,11 +10,25 @@ export class ModReportPlugin extends Plugin {
...
   public async execute(message: IMessage, args: string[]) {
+    const full_arg = args.join(' ');
+    const match_arr = full_arg.match(this._commadRegex);
+
+    if (!match_arr) {
+      message.reply('Invalid syntax.');
+      return;
+    }
+
+    args[0] = match_arr[1];
```
joey-colon, line 28: do we really need to mutate `args` here? i'm leaning towards the side of creating separate variables with more descriptive names

c650: Agree. Will fix

Issue #134, opened 2020-07-07
Implement Warnings and Bans via Lion for better tracking


c650:
Perhaps alter modreport to issue warnings automatically and facilitate bans

c650:
# Stories

## *Add* Mod Report
- mod adds mod report on user
- report is stored in db
- lion confirms

## *Warn* user
- files report like [Add Report](##AddModReport)
- also DMs user that they have a warning
- checks the number of warnings issued in some configurable time period
  - if it's over a threshold, ban is issued
    - optionally, an un-ban is scheduled
- user is told how many warnings have been recently filed
- lion confirms that report was filed and warning sent

## *Ban* user
- same as [Warn](##WarnUser) but ban no matter the warning count.

PR #136, opened 2020-07-09
Add warning and banning to modreport

c650:
#134


c650:
Still need to add scheduled un-ban and the functionality to query how many warnings in a given period

joey-colon:
res conflicts

Issue #123, opened 2020-05-27
logger wrapper


joey-colon:
Create a logger wrapper that uses something like [winston](https://www.npmjs.com/package/winston) and replace all console.log's with logs from this wrapper so we can have more insight on errors being thrown.

joey-colon:
(ideally, we should be persisting these errors somewhere, but this will not be impossible until we have a logger class created)

PR #139, opened 2020-08-04
create winston loggerService

c650:
#123


src/services/logger.service.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (process.env.NODE_ENV !== 'production') {
+      this._loggerInstance.add(
+        new Winston.transports.Console({
+          format: Winston.format.simple(),
+        })
+      );
+    }
+  }
+
+  public get() {
```
joey-colon, line 29: instead of exposing the logger instance to the client, create methods such as `.info()`, `.warn()`, etc that wrap around the logger instance. it isn't good to expose the instance to the consumer incase we decide to move away from winston for some reason

joey-colon: also, eventually i will want to push these logs to some 3rd party place so we can view production logs in realtime. we wouldn't be able to implement that with this current implementation.

c650: there are a ton of things like `info`/`warn` :(

c650: re: realtime logs, they get published to files; but we can add things to the winston logger so that they also publish elsewhere.

c650:
Made suggested changes. Added way to connect to Papertrail

joey-colon:
make sure you resolve conflict

PR #142, opened 2020-08-11
add code snippet filter, attachment check, to bot logger

c650:


src/app/handlers/blacklist.handler.ts
```diff
@@ -23,6 +29,17 @@ export class BlacklistHandler implements IHandler {
         'Hey, we are currently not allowing for UCF Zoom links to be posted within the Discord.'
       );
       message.delete();
+      return;
+    }
+
+    // If it's a class channel, we do some extra checks.
+    if (channel.name.toLowerCase().match(this._classChannelPattern)) {
```
joey-colon, line 36: you don't need to use this regex when you have a collection of class channels already. should only use regex as last resort imo

https://github.com/joey-colon/lion/blob/036c57a27d6bd97b4c68c43e57afdf1886c4d307/src/services/class.service.ts#L14

c650: ok will do

Issue #143, opened 2020-08-12
ee/gen ed class support


joey-colon:
a user should be able to type `!listclasses` and see classes for cs, it, gen ed, ee

c650:
#144

PR #144, opened 2020-08-12
refactor class registration system to decouple from specific categories (#143)

c650:


src/services/class.service.ts
```diff
@@ -48,7 +51,7 @@ export class ClassService {
   async unregister(request: IClassRequest): Promise<string> {
     const { author, categoryType, className } = request;
     try {
-      if (categoryType === null) {
+      if (categoryType === undefined) {
```
joey-colon, line 54: in register func, ur evaluating `!variable` i would use the same expression here

PR #149, opened 2020-08-18
refactor report ToString, fix bugs

c650:
* ban could crash bot if fails
* warning loses attachment during sendModMessage due to shallow copy
* adding report was not awaiting the response, leading to incomplete feedback

### DIFF
--- a/src/services/moderation.service.ts
+++ b/src/services/moderation.service.ts
@@ -128,8 +128,6 @@ export class ModService {
       reportId: await fileReportResult,
     });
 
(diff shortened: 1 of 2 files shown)


src/services/moderation.service.ts
```diff
@@ -129,6 +128,8 @@ export class ModService {
       reportId: await fileReportResult,
     });
 
+    console.log('warning: ' + report);
```
joey-colon: remove

Issue #135, opened 2020-07-09
Job to schedule unbans


c650:
For #134, we need a way to do temporary bans.

We can use a mongo collection to store scheduled unbans and poll daily.

PR #152, opened 2020-08-19
adds scheduled unban (#135)

c650:


src/services/moderation.service.ts
```diff
@@ -215,6 +228,44 @@ export class ModService {
...
+            await guild.unban(ban.user);
+          } catch (e) {
+            this._loggerService.error('Failed to unban user ' + ban.user, e);
+          }
+          bulk?.find({ _id: ban._id }).updateOne({ $set: { active: false } });
+        })
+        .toArray();
+
+      for (const unban of unbans || []) {
+        await unban;
```
joey-colon, line 260: Instead of awaiting here, you could wrap unbans in a Promise.all()

PR #155, opened 2020-08-22
[wip] add online user record functionality

joey-colon:


src/common/types.ts
```diff
@@ -121,3 +121,8 @@ export interface ILoggerWrapper {
   info(message: any, ...args: any[]): any;
   debug(message: any, ...args: any[]): any;
 }
+
+export interface IUserRecord {
+  count: number;
+  date: string;
+}
```
c650, line 128: Mongo can take `Date` objects.

src/app/jobs/user_record.job.ts
```diff
@@ -0,0 +1,57 @@
...
+        .get()
+        .channels.array()
+        .find((c) => c.name === this._announcementChannel.name);
+      if (!needle) {
+        return;
+      }
+      this._announcementChannel.channel = container.clientService.channels.get(
+        needle.id
+      ) as TextChannel;
+    }
```
c650, line 41: make its own function

src/app/jobs/user_record.job.ts
```diff
@@ -0,0 +1,57 @@
...
+    super();
+  }
+
+  async getGlobalMax(container: IContainer) {
+    try {
+      const userRecords = (await container.storageService.getCollections()).userrecords?.find();
+      if (!userRecords?.hasNext()) {
+        return -Infinity;
+      }
+      return ((await userRecords.next()) || {}).count;
```
c650, line 23:
```ts
(await userRecords.next())?.count
```

src/app/jobs/user_record.job.ts
```diff
@@ -0,0 +1,57 @@
...
+      .get()
+      .members.array()
+      .filter((member) => member.presence.status !== 'offline').length;
+    if (currentCount > cachedCount) {
+      this._announcementChannel.channel.send(`
+        **🎉🎉 New online user record 🎉🎉**
+        ${currentCount}
+        `);
+
+      // Update mongo record with `currentCount`
```
c650, line 54: Should be something like

```
updateOne({}, {$set: {count: currentCount, date: ...})
```

src/common/types.ts
c650: Also add `guild: Snowflake` field. Makes it much easier to distinguish between servers (prod and dev servers)

Kurtoid:
Are you sure this isn't taking it a bit too far?

src/app/jobs/user_record.job.ts
```diff
@@ -0,0 +1,57 @@
...
+        .find((c) => c.name === this._announcementChannel.name);
+      if (!needle) {
+        return;
+      }
+      this._announcementChannel.channel = container.clientService.channels.get(
+        needle.id
+      ) as TextChannel;
+    }
+
+    const cachedCount = (await this.getGlobalMax(container)) || -Infinity;
```
CallumCoy, line 43: Why not just make it `|| 0` since that is the lowest you could possibly get.  More curious about why, than saying do it.

c650: 0 is a valid number. -Infinity is much less valid and makes for a better sentinel.

PR #158, opened 2020-08-25
Dog update

CallumCoy:
Added random keyword to the dog plugin.


CallumCoy:
I personally didn't edit it, the only thing I can think changed is from when I audited the install.

PR #161, opened 2020-08-26
create react.handler.ts to let users vote to pin messages

c650:


src/app/handlers/react.handler.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    const onlyPin = (react: MessageReaction) => react.emoji.name === '📌';
+
+    message
+      .awaitReactions(onlyPin, { max: this._PIN_THRESH })
+      .then((collection) => {
+        const count = collection.first()?.count;
+        count >= this._PIN_THRESH && message.pin();
+      })
+      .catch();
```
joey-colon, line 31: might wanna emit logs if we enter the catch block

c650: thought about it, but it'll happen on like every message.

PR #163, opened 2020-08-28
Added alias system

CallumCoy:
where a creator can designate a name for their wake command(all commands are assured the name of the file will work).  Along with updated the help plugin to reflect these aliases.

### DIFF
--- a/src/app/handlers/command.handler.ts
+++ b/src/app/handlers/command.handler.ts
@@ -5,19 +5,19 @@ export class CommandHandler implements IHandler {
   constructor(public container: IContainer) { }
 
   public async execute(message: IMessage): Promise<void> {
-    const command = this.build(message.content);
+    const alias = this.build(message.content);
     const plugins = this.container.pluginService.plugins;
-    const commands = this.container.pluginService.commands;
+    const aliases = this.container.pluginService.aliases;
 
-    if (!command) {
+    if (!alias) {
       return;
     }
 
-    const plugin = plugins[commands[command.name]];
+    const plugin = plugins[aliases[alias.name]];
 
     if (!!plugin) {
 
-      if (!plugin.validate(message, command.args)) {
+      if (!plugin.validate(message, alias.args)) {
         message.reply(`Invalid arguments! Try: \`${Constants.Prefix}${plugin.usage}\``);
         return;
       }
@@ -27,7 +27,7 @@ export class CommandHandler implements IHandler {
       }
 
       try {
-        await plugin.execute(message, command.args);
+        await plugin.execute(message, alias.args);
       } catch (e) {
         this.container.loggerService.error(e);
       }
--- a/src/app/handlers/command.handler.ts
+++ b/src/app/handlers/command.handler.ts
@@ -10,7 +10,7 @@ export class CommandHandler implements IHandler {
     const aliases = this.container.pluginService.aliases;
 
     //checks to see if the user is actually talking to the bot
-    if (command === undefined) {
+    if (!command) {
       return;
     }
 
--- a/src/app/handlers/command.handler.ts
+++ b/src/app/handlers/command.handler.ts
@@ -2,17 +2,20 @@ import { IMessage, IContainer, IHandler } from '../../common/types';
 import Constants from '../../common/constants';
 
 export class CommandHandler implements IHandler {
-  constructor(public container: IContainer) {}
+  constructor(public container: IContainer) { }
 
   public async execute(message: IMessage): Promise<void> {
     const command = this.build(message.content);
     const plugins = this.container.pluginService.plugins;
+    const commands = this.container.pluginService.commands;
+
     if (!command) {
       return;
     }
 
-    if (!!plugins[command.name]) {
-      const plugin = plugins[command.name];
+    const plugin = plugins[commands[command.name]];
+
+    if (!!plugin) {
 
       if (!plugin.validate(message, command.args)) {
         message.reply(`Invalid arguments! Try: \`${Constants.Prefix}${plugin.usage}\``);
--- a/src/app/handlers/command.handler.ts
+++ b/src/app/handlers/command.handler.ts
@@ -2,22 +2,22 @@ import { IMessage, IContainer, IHandler } from '../../common/types';
 import Constants from '../../common/constants';
 
 export class CommandHandler implements IHandler {
-  constructor(public container: IContainer) { }
+  constructor(public container: IContainer) {}
 
   public async execute(message: IMessage): Promise<void> {
-    const alias = this.build(message.content);
+    const command = this.build(message.content);
     const plugins = this.container.pluginService.plugins;
     const aliases = this.container.pluginService.aliases;
 
-    if (!alias) {
+    //checks to see if the user is actually talking to the bot
+    if (command === undefined) {
       return;
     }
 
-    const plugin = plugins[aliases[alias.name]];
+    const plugin = plugins[aliases[command.name]];
 
-    if (!!plugin) {
-
-      if (!plugin.validate(message, alias.args)) {
+    if (plugin) {
+      if (!plugin.validate(message, command.args)) {
         message.reply(`Invalid arguments! Try: \`${Constants.Prefix}${plugin.usage}\``);
         return;
       }
@@ -27,7 +27,7 @@ export class CommandHandler implements IHandler {
       }
 
       try {
-        await plugin.execute(message, alias.args);
+        await plugin.execute(message, command.args);
       } catch (e) {
         this.container.loggerService.error(e);
       }
--- a/src/app/plugins/cat.plugin.ts
+++ b/src/app/plugins/cat.plugin.ts
@@ -11,6 +11,7 @@ export class CatPlugin extends Plugin {
   public name: string = 'Cat Plugin';
   public description: string = 'Generates pictures of cats.';
   public usage: string = 'cat <breed (optional)>';
+  public pluginCommands = ['cats'];
   public permission: ChannelType = ChannelType.Public;
   public pluginChannelName: string = Constants.Channels.Public.Pets;
 
@@ -32,7 +33,7 @@ export class CatPlugin extends Plugin {
           };
         });
       })
-      .catch((err) => this.container.loggerService.warn(err));
+      .catch((err) => this.container.loggerService.warn(err));;
   }
 
   public async execute(message: IMessage, args?: string[]) {
--- a/src/app/plugins/cat.plugin.ts
+++ b/src/app/plugins/cat.plugin.ts
@@ -11,7 +11,7 @@ export class CatPlugin extends Plugin {
   public name: string = 'Cat Plugin';
   public description: string = 'Generates pictures of cats.';
   public usage: string = 'cat <breed (optional)>';
-  public pluginCommands = ['cats'];
+  public pluginAlias = ['cats'];
   public permission: ChannelType = ChannelType.Public;
   public pluginChannelName: string = Constants.Channels.Public.Pets;
 
@@ -33,7 +33,7 @@ export class CatPlugin extends Plugin {
           };
         });
       })
-      .catch((err) => this.container.loggerService.warn(err));;
+      .catch((err) => this.container.loggerService.warn(err));
   }
 
   public async execute(message: IMessage, args?: string[]) {
--- a/src/app/plugins/help.plugin.ts
+++ b/src/app/plugins/help.plugin.ts
@@ -1,35 +1,95 @@
 import { Plugin } from '../../common/plugin';
-import { IContainer, IMessage, ChannelType } from '../../common/types';
+import { IContainer, IMessage, ChannelType, IPluginHelp } from '../../common/types';
 import Constants from '../../common/constants';
 import { RichEmbed } from 'discord.js';
 
 export class HelpPlugin extends Plugin {
   public name: string = 'Help Plugin';
   public description: string = 'Displays supported commands and usage statements.';
-  public usage: string = 'help';
+  public usage: string = 'help [Plugin Command]';
+  public pluginCommands = [];
   public permission: ChannelType = ChannelType.Bot;
-  private _embed: RichEmbed = new RichEmbed();
+  private _embed: RichEmbed[] = [];
+  private _pluginEmbed: IPluginHelp = {};
+
   constructor(public container: IContainer) {
     super();
+
+    this._embed[0] = new RichEmbed();
+    this._embed[1] = new RichEmbed();
   }
+
+
 
   public async execute(message: IMessage, args?: string[]) {
-    if (this._embed.fields!.length === 0) {
-      this._generateEmbed();
+    const commands = this.container.pluginService.commands
+    const input: string = (args === undefined || args.length === 0) ? '' : this._parseCommand(args);
+
+    if (!!commands[input]) {
+
+      const pluginName = commands[input];
+      const plugin = this.container.pluginService.plugins[pluginName];
+
+      if (plugin.permission === ChannelType.Admin || plugin.permission === ChannelType.Staff) {
+
+        if (this._embed[1].fields!.length === 0) {
+          this._generateEmbed(false, 1);
+        }
+
+        message.reply(this._embed[1]);
+
+      } else {
+        if (this._pluginEmbed[pluginName] === undefined) {
+          this._pluginEmbed[pluginName] = new RichEmbed();
+          this._generatePluginEmbed(pluginName);
+        }
+        message.reply(this._pluginEmbed[pluginName]);
+      }
+    } else if (input === "all") {
+      if (this._embed[0].fields!.length === 0) {
+        this._generateEmbed(true, 0);
+      }
+
+      message.reply(this._embed[0]);
+
+    } else {
+
+      if (this._embed[1].fields!.length === 0) {
+        this._generateEmbed(false, 1);
+      }
+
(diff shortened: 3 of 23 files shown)


src/app/plugins/cat.plugin.ts
```diff
@@ -32,7 +33,7 @@ export class CatPlugin extends Plugin {
           };
         });
       })
-      .catch((err) => this.container.loggerService.warn(err));
+      .catch((err) => this.container.loggerService.warn(err));;
```
c650, line 36: remove extra `;`

src/common/types.ts
```diff
@@ -28,6 +28,7 @@ export interface IPlugin {
   name: string;
   description: string;
   usage: string;
+  pluginCommands: string[];
```
c650, line 31: Why not make this optional so that plugins aren't forced to have an empty array?

src/services/plugin.service.ts
```diff
@@ -12,6 +13,24 @@ export class PluginService {
...
+  registerCommands(pluginName: string, args?: any): void {
+
+    const commands = this.plugins[pluginName].pluginCommands;
+
+    if (!commands.includes(pluginName)) {
+      this.commands[pluginName] = pluginName;
+    }
+
+    commands.forEach((command: string, index: number) => {
+      this.commands[command] === undefined ? this.commands[command] = pluginName : this.plugins[pluginName].pluginCommands.splice(index, 1);
```
c650, line 33: sketchy. Instead, pass over an already-used alias and log a warning.

```
if (this.commands[command]) {
  // log warning
} else {
  this.commands[command] = pluginName;
}

```

src/common/types.ts
c650: Also call it `alias` not commands.

CallumCoy: When I made the interface it was the first thing I did, and at the time I didn't know how to go around the if null process for a call without it been redundant.  However since I add the main command also to the list I can make it unrequired, just that I didn't consciously realize that I had solved the initial problem.  I will add it in an update.

src/services/plugin.service.ts
CallumCoy: I'll look at the other warning logs and implement it along with the other noted improvements.

src/services/plugin.service.ts
```diff
@@ -16,21 +16,38 @@ export class PluginService {
...
+  registerAliases(pluginName: string, args?: any): void {
+
+    const aliases = this.plugins[pluginName].pluginAlias;
+
+    if (aliases === undefined) {
+      this.aliases[pluginName] = pluginName;
+    } else {
+      if (!aliases.includes(pluginName)) {
+        if (this.aliases[pluginName] !== undefined) {
+          args.loggerService.warn(
```
CallumCoy, line 33: I know you think this is sketchy, but I think it is a valid fail-safe assuring all plugins that they can at least be called by the name of the plugin.

joey-colon:
oh and also, get prettier and set it up to run on save (thank me later)
https://www.codereadability.com/automated-code-formatting-with-prettier/

src/services/plugin.service.ts
```diff
@@ -16,21 +16,38 @@ export class PluginService {
...
-
-    if (!commands.includes(pluginName)) {
-      this.commands[pluginName] = pluginName;
+  registerAliases(pluginName: string, args?: any): void {
+
+    const aliases = this.plugins[pluginName].pluginAlias;
+
+    if (aliases === undefined) {
+      this.aliases[pluginName] = pluginName;
+    } else {
```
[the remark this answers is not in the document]
CallumCoy: This line isn't meant to catch an error. It is meant to determine if a list even exists for the current plugin.  If you look up 3 more lines you will see it is defined as such.

src/app/handlers/command.handler.ts
```diff
@@ -2,19 +2,22 @@ import { IMessage, IContainer, IHandler } from '../../common/types';
...
   public async execute(message: IMessage): Promise<void> {
-    const command = this.build(message.content);
+    const alias = this.build(message.content);
     const plugins = this.container.pluginService.plugins;
-    if (!command) {
+    const aliases = this.container.pluginService.aliases;
+
+    if (!alias) {
       return;
     }
```
c650, line 14: Log an error

src/app/handlers/command.handler.ts
```diff
@@ -2,19 +2,22 @@ import { IMessage, IContainer, IHandler } from '../../common/types';
...
 
-    if (!!plugins[command.name]) {
-      const plugin = plugins[command.name];
+    const plugin = plugins[aliases[alias.name]];
+
+    if (!!plugin) {
```
c650, line 18: nit: just `plugin`

src/app/handlers/command.handler.ts
```diff
@@ -2,19 +2,22 @@ import { IMessage, IContainer, IHandler } from '../../common/types';
 import Constants from '../../common/constants';
 
 export class CommandHandler implements IHandler {
-  constructor(public container: IContainer) {}
+  constructor(public container: IContainer) { }
 
   public async execute(message: IMessage): Promise<void> {
-    const command = this.build(message.content);
+    const alias = this.build(message.content);
```
c650, line 8: `command` is still the right name here. oops

src/app/plugins/help.plugin.ts
```diff
@@ -1,35 +1,97 @@
...
+
+        message.reply(this._embed[1]);
+      } else {
+        if (this._pluginEmbed[pluginName] === undefined) {
+          this._pluginEmbed[pluginName] = new RichEmbed();
+          this._generatePluginEmbed(pluginName);
+        }
+        message.reply(this._pluginEmbed[pluginName]);
+      }
+    } else if (input === `all`) {
```
c650, line 43: use single quote string literal, not backtick

src/app/plugins/help.plugin.ts
```diff
@@ -1,35 +1,97 @@
...
+    this._embed[targ].setColor('#0099ff').setTitle('**__These are the commands I support__**');
+
+    for (const targName of plugins) {
+      const plugin = this.container.pluginService.get(targName);
+      const aliases = plugin.pluginAlias || [];
 
-    for (let i = 0; i < plugins.length; i++) {
-      const plugin = this.container.pluginService.get(plugins[i]);
+      const altCalls =
+        all && aliases.length != 0 ? `aliases: ${aliases.join(', ')} , \n` : `aliases: None \n`;
```
c650, line 68: common prefix of "aliases: " should be outside the logic. Also, `aliases.length && aliases.join() || "None"`

src/app/plugins/listclasses.plugin.ts
```diff
@@ -41,7 +42,7 @@ export class ListClassesPlugin extends Plugin {
       for (const r of response) {
         try {
           await message.reply(r);
-        } catch (e) {}
+        } catch (e) { }
```
c650, line 45: Are you using a different format config ??

src/app/plugins/price.plugin.ts
```diff
@@ -13,11 +13,11 @@ export class PricePlugin extends Plugin {
...
   public description: string = 'Get financial quotes';
   public usage: string = 'price <ticker>; price AAPL';
+  public pluginAlias = [];
   public permission: ChannelType = ChannelType.Public;
   public pluginChannelName: string = Constants.Channels.Public.Finance;
 
-  private _STOCK_API_URL: string = 'https://cloud.iexapis.com/stable';
-  private _CRYPTO_API_URL: string = 'https://www.alphavantage.co/query?';
+  private _API_URL: string = 'https://www.alphavantage.co/query?';
   private _ADVANCED_QUOTE_LINK: string = 'https://finance.yahoo.com/quote/';
```
c650, line 22: This is all outdated -- please merge in master.

src/common/plugin.ts
```diff
@@ -11,6 +11,8 @@ export abstract class Plugin implements IPlugin {
 
   public abstract get permission(): ChannelType;
 
+  public pluginAlias?: string[];
```
c650, line 14: Since this is optional, did you really need to declare it in each plugin?

src/common/types.ts
```diff
@@ -51,16 +52,16 @@ export interface IContainer extends BottleContainer {
...
+export interface IUser extends User { }
 
 export interface IChannelCategory {
   [name: string]: string;
 }
 
-export interface IChannel extends Collection<Snowflake, GuildChannel> {}
-export interface IHttpResponse extends AxiosResponse {}
+export interface IChannel extends Collection<Snowflake, GuildChannel> { }
+export interface IHttpResponse extends AxiosResponse { }
```
c650, line 64: nit: formatting

src/services/plugin.service.ts
```diff
@@ -12,6 +13,37 @@ export class PluginService {
...
-    return (this.plugins[pluginName] = new PluginLoader(pluginName, args) as IPlugin);
+
+    const reference = (this.plugins[pluginName] = new PluginLoader(pluginName, args) as IPlugin);
+
+    this.registerAliases(pluginName);
+
+    return reference;
+  }
+
+  registerAliases(pluginName: string, _args?: any): void {
```
c650, line 24: Why `_args` not `args`?

src/services/plugin.service.ts
```diff
@@ -12,6 +13,37 @@ export class PluginService {
...
+      this.aliases[pluginName] = pluginName;
+    } else {
+      if (!aliases.includes(pluginName)) {
+        if (this.aliases[pluginName] !== undefined) {
+          throw new Error(
+            `Duplicate alias detected: ${pluginName} is claiming its primary alias ${pluginName}, prevouisly claimed by ${this.aliases[pluginName]}.`
+          );
+        }
+        this.aliases[pluginName] = pluginName;
+      }
```
c650, line 37: So the main plugin name is also considered an alias?

src/services/plugin.service.ts
```diff
@@ -12,6 +13,37 @@ export class PluginService {
...
+
+      aliases.forEach((alias: string) => {
+        if (this.aliases[alias] !== undefined) {
+          throw new Error(
+            `Duplicate alias detected: ${pluginName} is trying to claim ${alias}, but ${this.aliases[alias]} has already claimed it.`
+          );
+        }
```
c650, line 44: duplicate code.

src/services/plugin.service.ts
```diff
@@ -12,6 +13,37 @@ export class PluginService {
...
+        this.aliases[alias] = pluginName;
+      });
+    }
```
c650, line 47: We should check _all_ the plugin's aliases before adding any of them.

```ts
const aliasAlreadyInUse = aliases.reduce((acc, e) => acc || this.aliaes[e], false);
if (aliasAlreadyInUse) {
  // throw
}
aliases.forEach( (e) => this.aliases[e] = pluginName );
```

src/services/plugin.service.ts
c650, line 44: This will fail to add aliases after the first duplicate is found.

src/app/plugins/listclasses.plugin.ts
CallumCoy: I'm only using the default I'll change it for future.

src/services/plugin.service.ts
CallumCoy: yes, unless we want to set plugin to.

`const plugin = plugins[command.name] || plugins[command.name];`

src/services/plugin.service.ts
CallumCoy: Joey wants it to error on finding a duplicate plugin so that is meant to happen.

CallumCoy: While they both check for duplicates, they give different feedback.  I could put it in the function but it would literally only return a bool value.

src/services/plugin.service.ts
c650: oh nvm

src/services/plugin.service.ts
c650: it doesn't make sense. [alias1, duplicatedAlias, alias2]. alias2 would never be added.

c650: it doesn't make sense. [alias1, duplicatedAlias, alias2]. alias2 would never be added.

CallumCoy: Do you mean if a plugin has something like alias = [cats, cats], or if the Cat plugin had alias = [cat] then it will crash?  Cause the idea of the error was meant to stop a second author from trying to overwrite it.  If that is the issue I can do:

`(this.aliases[alias] !== undefined && this.aliases[alias] !== 'pluginName')`

c650: no the issue is that the behaviour should be "if any alias is invalid, dont add any of them, let the developer know they messed up" __not__ "keep adding aliases until one is invalid, then dont add anymore"

src/app/handlers/command.handler.ts
CallumCoy: Why would this be an error, all that means is the message is not to be read as a command?

CallumCoy: changed it to `(command === undefined)`, and left a comment to explain it, but as I said I don't see a reason to log it.  Since every message that doesn't start with ! will trigger a warning if we did that.

src/services/plugin.service.ts
joey-colon: @c650 i think you are misunderstanding. the behavior with how duplicate plugin names work is if the bot is registering one and comes across a duplicate key, we are stopping execution of program so that the developer is able to notice that something went wrong.

i was suggesting to apply the same logic here to remain consistent with what we've done in the past. i would prefer to not have duplicate plugin to stop execution of program and then duplicate alias to throw a log in the console. it makes sense to remain consistent with how we're handling the duplicate case. this decision is still up for debate but it should be dealt with in another pr if we want to change it. for now i am for stopping execution of program if duplicate alias is found

src/app/handlers/command.handler.ts
```diff
@@ -2,22 +2,22 @@ import { IMessage, IContainer, IHandler } from '../../common/types';
...
 
   public async execute(message: IMessage): Promise<void> {
-    const alias = this.build(message.content);
+    const command = this.build(message.content);
     const plugins = this.container.pluginService.plugins;
     const aliases = this.container.pluginService.aliases;
 
-    if (!alias) {
+    //checks to see if the user is actually talking to the bot
+    if (command === undefined) {
```
c650, line 13: Not a fan of explicit `=== undefined` better to just discretely use the truthiness of `command`.

src/app/plugins/help.plugin.ts
```diff
@@ -23,54 +23,55 @@ export class HelpPlugin extends Plugin {
...
       if (plugin.permission === ChannelType.Admin || plugin.permission === ChannelType.Staff) {
-        if (this._embed[1].fields!.length === 0) {
-          this._generateEmbed(false, 1);
+        // as the admin/mod commands aren't meant to be known this just shows the basic help as if nothing happened.
+        if (this._embed[0].fields!.length === 0) {
+          this._generateEmbed(0);
         }
 
-        message.reply(this._embed[1]);
+        message.reply(this._embed[0]);
```
c650, line 36: not a fan of hardcoded indexing into array here. what does it mean?

src/app/plugins/help.plugin.ts
```diff
@@ -23,54 +23,55 @@ export class HelpPlugin extends Plugin {
...
           this._generatePluginEmbed(pluginName);
         }
         message.reply(this._pluginEmbed[pluginName]);
       }
-    } else if (input === `all`) {
-      if (this._embed[0].fields!.length === 0) {
-        this._generateEmbed(true, 0);
+    } else if (input === 'all') {
+      if (this._embed[1].fields!.length === 0) {
+        this._generateEmbed(1);
```
c650, line 46: again with the magic numbers

src/app/plugins/help.plugin.ts
```diff
@@ -23,54 +23,55 @@ export class HelpPlugin extends Plugin {
...
+      message.reply(this._embed[1]);
     } else {
-      if (this._embed[1].fields!.length === 0) {
-        this._generateEmbed(false, 1);
+      if (this._embed[0].fields!.length === 0) {
+        this._generateEmbed(0);
       }
 
-      message.reply(this._embed[1]);
+      message.reply(this._embed[0]);
```
c650: saa

src/app/plugins/help.plugin.ts
```diff
@@ -23,54 +23,55 @@ export class HelpPlugin extends Plugin {
...
     }
   }
 
-  private _generateEmbed(all: boolean, targ: number) {
+  private _generateEmbed(targ: number) {
```
c650, line 59: You just use `targ` as 0 or 1? it should be boolean then, and have a more meaningful name

src/services/plugin.service.ts
```diff
@@ -37,7 +37,7 @@ export class PluginService {
       }
 
       aliases.forEach((alias: string) => {
-        if (this.aliases[alias] !== undefined) {
+        if (this.aliases[alias] !== undefined && this.aliases[alias] !== pluginName) {
```
c650, line 40: No need for explicit `!== undefined`.

c650 (changes requested):

c650 (approved):

CallumCoy:
@joey-colon Just so you know this pull request needs you to approve the requested changes. and you can review it further if you want.

joey-colon (approved):

PR #164, opened 2020-08-29
Cat upgrade

CallumCoy:
Catching up with the dog plugin working with no command.
Widened the breed list and used smaller font, via using embed for the list and using 3 columns (is seems this is the max embed allows).


c650:
bump @CallumCoy

PR #174, opened 2020-09-21
WordBlackList added

serspencer:
1. Code cleaned
2. All offences are labelled
3. Club channel is now only whitelisted to link sharing
4. Only one offence reported at a time (ie. if bad word and negative statement only badword is reported)


src/app/handlers/blacklist.handler.ts
```diff
@@ -2,37 +2,56 @@ import { TextChannel } from 'discord.js';
 import Constants from '../../common/constants';
 import { IContainer, IHandler, IMessage, ClassType } from '../../common/types';
 import { Moderation } from '../../services/moderation.service';
+import { readFileSync } from 'fs';
+
+const vader = require('vader-sentiment');
+const blockedWordsLine = (readFileSync('src/common/wordblacklist.txt', 'utf-8')).split(/\r?\n/);
```
c650, line 8: rename to 'blacklistedWords'

src/app/handlers/blacklist.handler.ts
```diff
@@ -2,37 +2,56 @@ import { TextChannel } from 'discord.js';
...
     { regex: /discord\.gg/, label: 'discord' },
     { regex: /group\s?me/, label: 'GroupMe' },
     { regex: /chegg\.com/, label: 'Chegg' },
     { regex: /coursehero\.com/, label: 'CourseHero' },
     { regex: /quizlet\.com/, label: 'Quizlet' },
+    { regex: /ucf\.zoom\.us/, label: 'Zoom' },
   ];
 
-  private _whitelistedChannels = new Set([Constants.Channels.Public.Clubs]);
+  private _whitelistedChannels = new Set([Constants.Channels.Public.Clubs]);              //Channels that are whitelisted for Link Sharing only.
```
c650, line 25: nit: line too long

src/app/handlers/blacklist.handler.ts
```diff
@@ -2,37 +2,56 @@ import { TextChannel } from 'discord.js';
...
+        message.delete();
+        return;
+      }
 
-    if (this._whitelistedChannels.has(channel.name)) {
+    //VaderSentiment Negativity Check
+    //console.log(intensity);                                                             //View Message Sentiment Score
+    if (intensity['compound'] < 0 ) {
+      this.container.messageService.sendBotReport(message, "Negativity");
+      // message.delete();                                                                //Optional Delete Negative Messages
```
c650, line 44: delete line.

src/app/handlers/blacklist.handler.ts
c650: also make sure that if the file doesn't exist we fail gracefully and have an empty list of words.

c650 (changes requested):

joey-colon:
i don't necessarily agree with managing the state of bad words on local file system.

@c650 maybe it's worth setting up a s3 bucket/equivalent to be able to dynamically load in the words. also, we can use something like https://github.com/LDNOOBW/List-of-Dirty-Naughty-Obscene-and-Otherwise-Bad-Words or https://github.com/zacanger/profane-words/blob/master/words.json - even directly sending a http request to these raw files would be better than using local file system imo

serspencer:
Should be setup so if you choose to implement a word list from anywhere it would be rather simple.  I have removed the native file for the time being.

src/app/handlers/blacklist.handler.ts
```diff
@@ -2,37 +2,61 @@ import { TextChannel } from 'discord.js';
...
+import { readFileSync } from 'fs';
+//Import VaderSentiment
+const vader = require('vader-sentiment');
+//Try to open the black listed words file
+let blacklistedWords: string[];
+try {
+  blacklistedWords = (readFileSync('src/common/wordblacklist.txt', 'utf-8')).split(/\r?\n/);
+} catch (error) {
+  blacklistedWords = [];
+  console.error(error);
```
c650, line 14: use the logger service instead of console.error

src/app/handlers/blacklist.handler.ts
```diff
@@ -2,37 +2,61 @@ import { TextChannel } from 'discord.js';
...
 
   private _whitelistedChannels = new Set([Constants.Channels.Public.Clubs]);
   constructor(public container: IContainer) {}
 
   public async execute(message: IMessage): Promise<void> {
     const channel = message.channel as TextChannel;
+    const intensity = vader.SentimentIntensityAnalyzer.polarity_scores(message.content);
+
+    //Word Black List
+    if (blacklistedWords.some((blockedWords) =>
```
c650, line 40: save the result of the some() call into a variable for more readable `if` conditon

src/app/handlers/blacklist.handler.ts
```diff
@@ -2,37 +2,61 @@ import { TextChannel } from 'discord.js';
...
+    if (blacklistedWords.some((blockedWords) => 
+      message.content.toLocaleLowerCase().includes(blockedWords))) {
+        this.container.messageService.sendBotReport(message, "Blocked Word");
+        message.delete();
+        return;
+      }
 
-    if (this._whitelistedChannels.has(channel.name)) {
+    //VaderSentiment Negativity Check
+    if (intensity['compound'] < 0 ) {
```
c650, line 48: is this too strict?

src/app/plugins/slowmode.plugin.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  public async execute(message: IMessage, args: string[]) {
+    const match_arr = args.join(' ').match(this._commandPattern);
+
+    if (args[0] === '!slowmode') {
+      message.reply('Slowmode!');
+      return;
+    }
+  }
+}
```
c650, line 26: should be in its own pr

src/bootstrap/plugin.loader.ts
```diff
@@ -17,6 +17,7 @@ import { FetchClassChannelsPlugin } from '../app/plugins/fetchclasschans.plugin'
 import { AddClassChannelsPlugin } from '../app/plugins/addclasschans.plugin';
```
c650, line 17: should be in its own pr

c650 (changes requested):

src/bootstrap/plugin.loader.ts
serspencer: Yeah what the heck?  That is a completely different branch.

src/app/handlers/blacklist.handler.ts
serspencer: You do have to work hard to get an overall compound but I can adjust to -0.25.  It can always be changed if we see to many false positives.

src/app/handlers/blacklist.handler.ts
```diff
@@ -11,7 +11,7 @@ try {
   blacklistedWords = (readFileSync('src/common/wordblacklist.txt', 'utf-8')).split(/\r?\n/);
 } catch (error) {
   blacklistedWords = [];
-  console.error(error);
+  console.log(error);
```
c650, line 14: not what i meant. `this.container.loggerService.error(...)`

src/app/handlers/blacklist.handler.ts
```diff
@@ -35,17 +35,18 @@ export class BlacklistHandler implements IHandler {
   public async execute(message: IMessage): Promise<void> {
     const channel = message.channel as TextChannel;
     const intensity = vader.SentimentIntensityAnalyzer.polarity_scores(message.content);
+    const blockedWords = blacklistedWords.some;
 
     //Word Black List
-    if (blacklistedWords.some((blockedWords) => 
+    if (blockedWords((blockedWords) =>
```
c650: also not what i meant. 

```
const hasBlockedWord = ...some(...)

if (hasBlockedWord) {

}
```

src/app/handlers/blacklist.handler.ts
Kurtoid: This could also be a `.env` parameter

c650:
stale

PR #175, opened 2020-09-21
PubSub usage fixed

serspencer:


c650 (changes requested):
there's a ton of stuff in here. not just pubsub.

PR #176, opened 2020-09-21
Updated aliases for price to include stock

serspencer:
This should close out the issue for aliases.


c650 (changes requested):
Again there is a ton of stuff in this PR. rEbASe

Issue #178, opened 2020-09-23
Make sure that message.member is defined


c650:
According to [this issue](https://github.com/discordjs/discord.js/issues/2984), sometimes message.member is not defined. We use message.member in some commands so it's important that it's defined; it's caused issues with failed commands in the past. 

https://discord.js.org/#/docs/main/stable/class/GuildMemberManager?scrollTo=fetch

PR #180, opened 2020-09-23
create _tryEnsureMessageMember in listener.ts

c650:
#178


joey-colon (approved):
the way u implemented this makes sense and i would do the same. just a note if we need to do more adaptations on the message object, we should introduce adapters and avoid populating this file with the adapting logic

c650:
Agree. If we have to do something like this a second time, we will introduce a new abstraction. Perhaps called `MessageTransform` or something.

Issue #173, opened 2020-09-18
Bot command to restrict access to specific channels


c650:
#politics

PR #182, opened 2020-09-25
adds command to ban user from specific channels.

c650:
#173

### DIFF
--- a/src/services/moderation.service.ts
+++ b/src/services/moderation.service.ts
@@ -283,42 +283,46 @@ export class ModService {
     channels: GuildChannel[]
   ): Promise<GuildChannel[]> {
     const uid = Moderation.Helpers.resolveUser(guild, username);
+    const successfulBanChannelList: GuildChannel[] = [];
 
     if (!uid) {
       this._loggerService.error(`Failed to resolve ${username} to a user.`);
-      return [];
+      return successfulBanChannelList;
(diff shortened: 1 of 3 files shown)


src/services/moderation.service.ts
```diff
@@ -275,6 +274,53 @@ export class ModService {
...
+    const uid = Moderation.Helpers.resolveUser(guild, username);
+
+    if (!uid) {
+      this._loggerService.error(`Failed to resolve ${username} to a user.`);
+      return [];
+    }
+
+    const ret: GuildChannel[] = [];
+    const proms = [];
+    for (const chan of channels) {
```
joey-colon, line 294: this can be a .reduce()

src/services/moderation.service.ts
```diff
@@ -275,6 +274,53 @@ export class ModService {
...
+              `Failed to adjust permissions for ${username} in ${chan.name}`,
+              ex
+            );
+          })
+      );
+    }
+
+    await Promise.all(proms);
+
+    this._insertReport(
```
joey-colon, line 311: await

joey-colon (approved):

src/services/moderation.service.ts
c650: my intention was to not block on reporting

PR #183, opened 2020-09-25
introduce plugin codegen

joey-colon:
run via `npm run generate plugin "plugin name"` 

e.g.`npm run generate plugin "uc florida"`

### DIFF
--- a/scripts/generate.js
+++ b/scripts/generate.js
@@ -1,0 +1,84 @@
+#!/usr/bin/env node
+const fs = require('fs/promises');
+
+const buildPluginContent = (pluginName) =>
+  `import { Plugin } from '../../common/plugin';
+import { IContainer, IMessage, ChannelType } from '../../common/types';
+
+export class ${pluginName} extends Plugin {
+  public name: string = '${pluginName}';
+  public description: string = 'Some sort of a description.';
+  public usage: string = '${pluginName}';
+  public permission: ChannelType = ChannelType.Public;
+
+  constructor(public container: IContainer) {
+    super();
+  }
+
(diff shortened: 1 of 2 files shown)


scripts/generate.js
```diff
@@ -0,0 +1,84 @@
...
+  return tokens.join('\n').replace(
+    `};`,
+
+    `  ${pluginName.toLowerCase().substr(0, pluginName.length - 6)}: ${pluginName},
+};`
+  );
+};
+
+const createPlugin = async (name) => {
+  const sanitizedName = name.toLowerCase().replace(/ /g, '_');
```
c650, line 49: use `\s`

scripts/generate.js
```diff
@@ -0,0 +1,84 @@
...
+    0,
+    `import { ${pluginName} } from '../app/plugins/${fileName}.plugin';`
+  );
+  return tokens.join('\n').replace(
+    `};`,
+
+    `  ${pluginName.toLowerCase().substr(0, pluginName.length - 6)}: ${pluginName},
+};`
+  );
+};
```
c650, line 46: bit sketchy

maybe we should put a comment in `plugin.loader.ts` that says 'do not edit; file managed by generate.js'

c650 (approved):

scripts/generate.js
joey-colon: we can create a parser to confirm that `plugin.loader.ts` is in the correct format, have it test on circleci, and break build if it isn't.

c650: agree

PR #188, opened 2020-10-01
mvp slowmode plugin

c650:


src/app/plugins/slowmode.plugin.ts
```diff
@@ -0,0 +1,63 @@
...
+  constructor(public container: IContainer) {
+    super();
+  }
+
+  public validate(message: IMessage, args?: string[]) {
+    return !!args && this._commandPattern.test(args.join(' '));
+  }
+
+  public async execute(message: IMessage, args: string[]) {
+    const match = args.join(' ').match(this._commandPattern);
```
joey-colon, line 26: this is an anti pattern since you are already given `args` which *should* have what you are wanting and overall the approach reduces readability when u introduce regexes into the codebase when there are simpler workarounds. 

all you would need to do is:
```
const [, expiration, slowmodeSetting, ...channels] = args || [];

if (!expiration || !slowModeSetting || !channels) {
  return;
}
```

joey-colon (approved):

src/app/plugins/slowmode.plugin.ts
```diff
@@ -0,0 +1,63 @@
...
+      return;
+    }
+
+    const [_, expiration, slowmodeSetting, channels] = match;
+
+    const channel_objs =
+      channels
+        .match(this._channelIDRegex)
+        ?.map((c) => this.container.guildService.get().channels.get(c.replace(/\D/g, '')))
+        .filter((c) => c !== undefined) || [];
```
joey-colon, line 39: random comment, you can do `.filter(Boolean)` here and becomes more readable (imo)

PR #189, opened 2020-10-02
Pub Sub Plugin

CallumCoy:
Dues to one of the previous plugins owning the most fitting name I renamed the old plugin to tendysub, as that plugin was only for the chicken tender sub.
Commands that come with the plugin
[blank] or random: gets a random sub and return the current price the last sale date, and a picture.
[sub name]: gets the current price, the last sale date, and a picture of the sub
list or types: return a list of available subs. (I implemented an auto growing feature here as Abe is working on the API still)
update: This command updates the list of available subs.  (Once again implemented as Abe is still working on the API, and I wanted to keep the idea that the list only needs to be updated from time to time)

note: not sure why my old commits from the alias are in here, however, all the changed files match up with only the plugin update.

### DIFF
--- a/src/app/plugins/pubsub.plugin.ts
+++ b/src/app/plugins/pubsub.plugin.ts
@@ -1,23 +1,121 @@
+import Constants from '../../common/constants';
 import { Plugin } from '../../common/plugin';
-import { IContainer, IMessage, ChannelType } from '../../common/types';
+import { ChannelType, IContainer, IHttpResponse, IMessage } from '../../common/types';
+import { RichEmbed } from 'discord.js';
+
+class Breed {
+  public name: string = '';
+  public id: string = '';
+}
 
 export class PubSubPlugin extends Plugin {
   public name: string = 'Pub Sub Plugin';
-  public description: string = 'Are publix chicken tender subs on sale?';
-  public usage: string = '';
-  public pluginAlias = [];
-  public permission: ChannelType = ChannelType.Bot;
+  public description: string =
+    'Get them prices for your subs, and a steamy picture of the subs you need in your life.';
+  public usage: string = 'pubSub <sub (optional)>';
+  public pluginAlias = ['sub', 'subs', 'sandwich', 'samwich', 'sarnie'];
+  public permission: ChannelType = ChannelType.Public;
+  public pluginChannelName: string = Constants.Channels.Public.Food;
+
+  private _API_URL: string = 'https://pubsub-api.dev/';
+  private _subs: string[] = [];
+  private _embedSubs = new RichEmbed();
 
   constructor(public container: IContainer) {
     super();
+    this.updateList();
+  }
+
+  private updateList() {
+    this.container.httpService
+      .get(`${this._API_URL}allsubs/`)
+      .then((response: IHttpResponse) => {
+        const breeds = response.data;
+
+        this._subs = breeds.map((subData: { name: string }) => {
+          return subData.name
+            .toLowerCase()
+            .split('-')
+            .join(' ');
+        });
+        this._generateEmbedSubs();
+      })
+      .catch((err) => this.container.loggerService.warn(err));
   }
 
   public async execute(message: IMessage, args?: string[]) {
-    const state = this.container.storeService.get('pubSubStore').state;
-    if (state?.onSale) {
-      message.reply('Chicken tender subs are on sale :eyes:');
-    } else {
-      message.reply('Chicken tender subs are **NOT** on sale.');
+    if (args === undefined || args.length === 0) {
+      args = [''];
     }
+
+    if (args[0].includes('type') || args[0].includes('list')) {
+      //Simply return the list of supported subs
+      message.reply(this._embedSubs);
+      return;
+    }
+
+    const subIn = this._parseCommand(args);
+
+    // checks if their sub was a sub, then if that sub is recognised
+    const subEntry = this._subs.find((sub: string) => sub === subIn) || '';
+
+    if (subIn == 'reboot') {
+      this.updateList();
+      return;
+    } else if (subEntry == '' && subIn != 'random' && subIn != '') {
+      message.reply('Sub not found.  Use "list" or "type" to get a list of possible subs');
+      return;
+    }
+
+    this.container.loggerService.debug(subEntry);
+
+    //recieves the according info and posts, or derps
+    await this.container.httpService
+      .get(`${this._API_URL}subs/?name=${subEntry.split(' ').join('-')}`)
+      .then((response: IHttpResponse) => {
+        const subData = response.data[0];
+        const saleInfo = subData.status
+          ? `a dicounted price of ${subData.price} until ${subData.last_sale.split('-')[1]}`
+          : `${subData.price}, the last sale Date was ${subData.last_sale}`;
+
+        const embed: RichEmbed = new RichEmbed();
+
+        embed.setColor('#0099ff').setTitle(subData.sub_name.split('-').join(' '));
+
+        embed.description = `Get your own *${subData.sub_name
+          .split('-')
+          .join(
+            ' '
+          )}* for ${saleInfo}. So what are you waiting for?  Come on down to Publix now to get yourself a beautiful sub. Just look at this beauty right here! 😍😍😍🤤🤤🤤`;
+
+        embed.setImage(subData.image);
+
+        message.reply(embed);
+      })
+      .catch((err) => this.container.loggerService.warn(err));
+  }
+
+  private _generateEmbedSubs() {
+    const numCols = Math.min(3, Math.ceil(this._subs.length / 10));
+    const numRows = Math.ceil(this._subs.length / numCols);
+
+    this._embedSubs = new RichEmbed();
+
+    this._embedSubs.setColor('#0099ff').setTitle('Types of Subs');
+
+    for (let Cols = 0; Cols < numCols; Cols++) {
+      const columnSubs = this._subs.slice(numRows * Cols, numRows * (Cols + 1));
+
+      this._embedSubs.addField(
+        `${columnSubs[0].charAt(0)} - ${columnSubs[columnSubs.length - 1].charAt(0)}`,
+        columnSubs.join('\n'),
+        true
+      );
+    }
+  }
+
+  // gets the commands and puts spaces between all words
+  private _parseCommand(args: string[]): string {
+    return args.map((str) => str.toLowerCase()).join(' ');
   }
 }
--- a/src/app/plugins/pubsub.plugin.ts
+++ b/src/app/plugins/pubsub.plugin.ts
@@ -30,9 +30,7 @@ export class PubSubPlugin extends Plugin {
       .then((response: IHttpResponse) => {
         const subs = response.data;
 
-        this._subs = subs.map((subData: { name: string }) => {
-          return subData.name.toLowerCase();
-        });
+        this._subs = subs.map((subData: { name: string }) => subData.name.toLowerCase());
       })
       .catch((err) => this.container.loggerService.warn(err));
   }
@@ -47,8 +45,6 @@ export class PubSubPlugin extends Plugin {
       return;
     }
 
-    console.log(`${this._subs[0] === subIn}`);
-
     // checks if their sub was a sub, then if that sub is recognised
     const subType = this._subs.find((sub: string) => sub === subIn) || '';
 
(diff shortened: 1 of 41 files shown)


src/app/plugins/pubsub.plugin.ts
```diff
@@ -1,23 +1,116 @@
...
   }
 
   public async execute(message: IMessage, args?: string[]) {
-    const state = this.container.storeService.get('pubSubStore').state;
-    if (state?.onSale) {
-      message.reply('Chicken tender subs are on sale :eyes:');
-    } else {
-      message.reply('Chicken tender subs are **NOT** on sale.');
+    if (args === undefined || args.length === 0) {
+      args = [''];
```
joey-colon, line 43: don't reassign a parameter variable

src/app/plugins/pubsub.plugin.ts
```diff
@@ -1,23 +1,116 @@
...
+    }
+
+    if (args[0].includes('type') || args[0].includes('list')) {
```
joey-colon, line 46: would be more readable if you did something like
```
const [cmd] = args || [];

if (cmd === 'type' || ...)
```

src/app/plugins/pubsub.plugin.ts
```diff
@@ -1,23 +1,116 @@
...
+      return;
+    }
+
+    this.container.loggerService.debug(subEntry);
+
+    //recieves the according info and posts, or derps
+    await this.container.httpService
+      .get(`${this._API_URL}subs/?name=${subEntry.split(' ').join('-')}`)
+      .then((response: IHttpResponse) => {
+        const subData = response.data[0];
```
joey-colon, line 71: const [subData] = response.data;

src/app/plugins/pubsub.plugin.ts
```diff
@@ -1,23 +1,116 @@
...
+      message.reply(this._embedSubs);
+      return;
+    }
+
+    const subIn = this._parseCommand(args);
+
+    // checks if their sub was a sub, then if that sub is recognised
+    const subEntry = this._subs.find((sub: string) => sub === subIn) || '';
+
+    if (subIn == 'reboot') {
```
joey-colon, line 57: this should not be done on user cmd, but rather using a TTL approach. see the garage plugin for an example implementation

src/app/plugins/pubsub.plugin.ts
```diff
@@ -1,23 +1,116 @@
...
+    if (args[0].includes('type') || args[0].includes('list')) {
+      //Simply return the list of supported subs
+      message.reply(this._embedSubs);
+      return;
+    }
+
+    const subIn = this._parseCommand(args);
+
+    // checks if their sub was a sub, then if that sub is recognised
+    const subEntry = this._subs.find((sub: string) => sub === subIn) || '';
```
joey-colon, line 55: you don't need to cast this to be an empty string. this returns undefined and it is good enough for your check below

src/app/plugins/pubsub.plugin.ts
```diff
@@ -1,23 +1,116 @@
...
+
+    if (subIn == 'reboot') {
+      this.updateList();
+      return;
+    } else if (subEntry == '' && subIn != 'random' && subIn != '') {
```
joey-colon, line 60: `===` `!==`

src/app/plugins/pubsub.plugin.ts
```diff
@@ -1,23 +1,116 @@
...
+  }
+
+  private updateList() {
+    this.container.httpService
+      .get(`${this._API_URL}allsubs/`)
+      .then((response: IHttpResponse) => {
+        const subs = response.data;
+
+        this._subs = subs.map((subData: { name: string }) => {
+          return subData.name
```
joey-colon, line 31: this is unneeded complexity with how you are serializing each entry and reflects later on in the code. its fine to keep it as `-` and replace user input of spaces to `-`

src/app/plugins/pubsub.plugin.ts
```diff
@@ -1,23 +1,116 @@
...
+
+  private _generateEmbedSubs() {
+    const numCols = Math.min(3, Math.ceil(this._subs.length / 10));
+    const numRows = Math.ceil(this._subs.length / numCols);
+
+    this._embedSubs = new RichEmbed();
+
+    this._embedSubs.setColor('#0099ff').setTitle('Types of Subs');
+
+    for (let Cols = 0; Cols < numCols; Cols++) {
```
joey-colon, line 101: dont prefix var names with capital letter

joey-colon (changes requested):

src/app/plugins/pubsub.plugin.ts
CallumCoy: I was young and dumb when I wrote this code.  I'll fix it in the cat api as well in another pull request.

src/app/plugins/pubsub.plugin.ts
CallumCoy: changed due to the previous fix.

src/app/plugins/pubsub.plugin.ts
CallumCoy: :open_mouth: I forgot all about pattern matching.

src/app/plugins/pubsub.plugin.ts
CallumCoy: I actually do since when subEntry === '' that get Abes API to return a random sub.  The other route would be to change the const to a var and reassign it upon reading.  But I personly think that would be counterproductive.

src/app/plugins/pubsub.plugin.ts
```diff
@@ -28,60 +32,51 @@ export class PubSubPlugin extends Plugin {
         const subs = response.data;
 
         this._subs = subs.map((subData: { name: string }) => {
-          return subData.name
-            .toLowerCase()
-            .split('-')
-            .join(' ');
+          return subData.name.toLowerCase();
```
c650, line 34: if you omit the {} in the lambda: `=> subData.name.toLowerCase()` , no need to say `return`

src/app/plugins/pubsub.plugin.ts
```diff
@@ -28,60 +32,51 @@ export class PubSubPlugin extends Plugin {
...
-    if (args[0].includes('type') || args[0].includes('list')) {
+    if (subIn.includes('type') || subIn.includes('list')) {
       //Simply return the list of supported subs
+      this._generateEmbedSubs();
       message.reply(this._embedSubs);
       return;
     }
 
-    const subIn = this._parseCommand(args);
+    console.log(`${this._subs[0] === subIn}`);
```
c650, line 50: remove console.log

c650: or use `container.loggerService.debug`

src/app/plugins/pubsub.plugin.ts
```diff
@@ -28,60 +32,51 @@ export class PubSubPlugin extends Plugin {
...
-      this.updateList();
-      return;
-    } else if (subEntry == '' && subIn != 'random' && subIn != '') {
+    if (subType === '' && subIn !== 'random' && subIn !== '') {
       message.reply('Sub not found.  Use "list" or "type" to get a list of possible subs');
       return;
     }
 
-    this.container.loggerService.debug(subEntry);
+    this.container.loggerService.debug(subType);
```
c650, line 60: saa

src/app/plugins/pubsub.plugin.ts
```diff
@@ -28,60 +32,51 @@ export class PubSubPlugin extends Plugin {
...
       .then((response: IHttpResponse) => {
-        const subData = response.data[0];
-        const saleInfo = subData.status
-          ? `a dicounted price of ${subData.price} until ${subData.last_sale.split('-')[1]}`
-          : `${subData.price}, the last sale Date was ${subData.last_sale}`;
+        const [subData] = response.data;
+        const parsedNamed = subData.sub_name.split('-').join(' ');
+
+        const saleInfo =
+          subData.status.toLowerCase() === 'true' //not sure why I have to do it this way but I do
```
c650, line 70: why not status.toBoolean or something?

src/app/plugins/pubsub.plugin.ts
```diff
@@ -28,60 +32,51 @@ export class PubSubPlugin extends Plugin {
...
     await this.container.httpService
-      .get(`${this._API_URL}subs/?name=${subEntry.split(' ').join('-')}`)
+      .get(`${this._API_URL}subs/?name=${subType}`)
       .then((response: IHttpResponse) => {
-        const subData = response.data[0];
-        const saleInfo = subData.status
-          ? `a dicounted price of ${subData.price} until ${subData.last_sale.split('-')[1]}`
-          : `${subData.price}, the last sale Date was ${subData.last_sale}`;
+        const [subData] = response.data;
+        const parsedNamed = subData.sub_name.split('-').join(' ');
```
c650, line 67: split/join is same as just `replace`

src/app/plugins/pubsub.plugin.ts
```diff
@@ -28,60 +32,51 @@ export class PubSubPlugin extends Plugin {
...
 
-        embed.setColor('#0099ff').setTitle(subData.sub_name.split('-').join(' '));
+        embed.setColor('#0099ff').setTitle(parsedNamed);
 
-        embed.description = `Get your own *${subData.sub_name
-          .split('-')
-          .join(
-            ' '
-          )}* for ${saleInfo}. So what are you waiting for?  Come on down to Publix now to get yourself a beautiful sub. Just look at this beauty right here! 😍😍😍🤤🤤🤤`;
+        embed.description = `Get your own *${parsedNamed}* for ${saleInfo}. So what are you waiting for?  Come on down to Publix now to get yourself a beautiful sub. Just look at this beauty right here! 😍😍😍🤤🤤🤤`;
```
c650, line 78: we have a :pubsub: emote

c650: also prefer setDescription if there is one

src/app/plugins/pubsub.plugin.ts
```diff
@@ -90,16 +85,26 @@ export class PubSubPlugin extends Plugin {
...
+    }
+
     const numCols = Math.min(3, Math.ceil(this._subs.length / 10));
     const numRows = Math.ceil(this._subs.length / numCols);
 
     this._embedSubs = new RichEmbed();
 
     this._embedSubs.setColor('#0099ff').setTitle('Types of Subs');
+    const subs = this._subs.map((sub) => {
+      return sub.split('-').join(' ');
```
c650, line 102: saa, split+join = replace

c650 (changes requested):

src/app/plugins/pubsub.plugin.ts
CallumCoy: would have to make function for it, which I don't think is worth it.  As that just return it's truthy value.

src/app/plugins/pubsub.plugin.ts
CallumCoy: / /g for future me. wasn't sure how to get it to replace all. and the above is what I got on my first search.

src/app/plugins/pubsub.plugin.ts
CallumCoy: I don't actually know where this came from XD.  someone put it in my cat plugin so I thought it needed to be their XD

src/app/plugins/pubsub.plugin.ts
```diff
@@ -1,23 +1,114 @@
...
+  }
+
+  private async _generateEmbedSubs(): Promise<void> {
+    const lastListUpdate: number = Date.now() - this._LAST_UPD_TIME;
+    if (lastListUpdate < this._SUB_UPD_THRESH) {
+      return;
     } else {
-      message.reply('Chicken tender subs are **NOT** on sale.');
+      this._LAST_UPD_TIME = Date.now(); //since it only updates once a day, don't need to worry about accuracy
     }
```
c650, line 89: no need for if/else because the if branch just returns. move the else body out  and just have standalone if

c650 (approved):

c650 (changes requested):
> note: not sure why my old commits from the alias are in here, however, all the changed files match up with only the plugin update.

Please rebase your branch off current master and roll on only the latest commits dealing with the actual changes. Thanks.

CallumCoy:
> > note: not sure why my old commits from the alias are in here, however, all the changed files match up with only the plugin update.
> 
> Please rebase your branch off current master and roll on only the latest commits dealing with the actual changes. Thanks.

Rebasing leads to it flat out deleting the alias system despite the fact that I pulled before making the plug-in.  If you want I can close this pull request, then delete, and recreate my fork with the changed files.

c650:
> Rebasing leads to it flat out deleting the alias system despite the fact that I pulled before making the plug-in. If you want I can close this pull request, then delete, and recreate my fork with the changed files.

Make sure that your fork of master is up to date with ours, then you can rebase off that.

c650:
stale

PR #193, opened 2020-10-07
Scores Plugin: Support for upcoming games

tanndlin:


src/app/plugins/scores.plugin.ts
```diff
@@ -49,7 +46,11 @@ export class ScoresPlugin extends Plugin {
       teamArg = argArray.join(' ');
     }
 
-    const endpoint = this._ENDPOINTS.get(sportArg.toLowerCase()) || '';
+    const endpoint = this._ENDPOINTS.get(sportArg.toLowerCase());
+    if (endpoint == null) {
```
c650, line 50: just !endpoint

c650 (approved):

c650 (approved):
lgtm

c650 (approved):

src/app/plugins/scores.plugin.ts
```diff
@@ -29,6 +32,11 @@ export class ScoresPlugin extends Plugin {
     let teamArg;
     let sportArg;
 
+    if (parsedArgs.toLowerCase() === 'help') {
+      this._getHelp(message);
```
c650, line 36: sendHelp

src/app/plugins/scores.plugin.ts
```diff
@@ -148,10 +153,25 @@ export class ScoresPlugin extends Plugin {
...
+          { name: 'Examples', value: '!scores NCAA UCF\n!scores NCAA Florida State\n!scores NFL Seattle\n!scores MLB Atlanta' }
+        ]
+      }
+    });
+  }
+
+  private _tryUpcomingGame(embedBuffer: RichEmbed[], game: any, teamArg: string): boolean {
+
+    let teamsData = game.match(this._UPCOMING_REGEX);
+    const time = String(game.match(this._UPCOMING_TIME_REGEX)); //.split does not work unless you cast string
```
c650, line 174: why not access .match()[0] ?

joey-colon: if u do what corgo suggests, don't directly use indexes!! use array destructuring :)

tanndlin: I don't access [0] because game.match(reg) can be null. I check for that first then access when processing the data

joey-colon: If you do `String(null)`, it will return a string literal of `'null'` is this desired? if not, you can do `const time = game.match(this._UPCOMING_TIME_REGEX) || '').split(...)`

tanndlin: I don't want to continue if it is null, so that's why I don't access until `if(!teamsData || !time){ return false; }`
If the matches return null that means either the team requested A) Isn't playing this week, or B) Doesn't exist.
So it is helpful to know when it is null rather than set it to empty string and go on

src/app/plugins/scores.plugin.ts
```diff
@@ -62,81 +72,155 @@ export class ScoresPlugin extends Plugin {
...
 
         visitorData[0] = visitorData[0].slice(1); //Remove the prefix that starts the visitors name
 
         const visitorName = this._getTeamNameFromTeamData(visitorData);
         const homeName = this._getTeamNameFromTeamData(homeData);
 
+        //Check if game does not contain all search terms
+        if (
+          !(
+            this._strictlyContainsAllTokens(visitorName, teamArg) ||
```
c650, line 108: should be && 

'if they don't _both_ contain all tokens' => if ( !(first && second ) )

src/app/plugins/scores.plugin.ts
```diff
@@ -62,81 +72,155 @@ export class ScoresPlugin extends Plugin {
...
-
-          if (homeScore > visitorScore) { winning = 1; }  //Winning
-          else if (homeScore < visitorScore) { winning = 0; } //Losing
-          else if (homeScore == visitorScore) { winning = 2; } //Tied
-
-          opponentName = visitorName;
-          teamName = homeName;
-        }
+        const winning = this._strictlyContainsAllTokens(visitorName, teamArg) ?
+          this._evaluateScores(visitorScore, homeScore) : this._evaluateScores(homeScore, visitorScore);
```
c650, line 122: nit: ugly

src/app/plugins/scores.plugin.ts
```diff
@@ -62,81 +72,155 @@ export class ScoresPlugin extends Plugin {
...
+        const winningString = game.toLowerCase().includes('final') ? this._FINAL_LABELS[winning] : this._WINNING_LABELS[winning];
 
-        const winningString = this._WINNING_LABELS[winning];
+        const opponentName = this._strictlyContainsAllTokens(visitorName, teamArg) ? homeName : visitorName;
```
c650, line 125: same condition evaluated twice -- actually several times! `_strictlyContainsAllTokens()` isn't fast, either!!

src/app/plugins/scores.plugin.ts
```diff
@@ -62,81 +72,155 @@ export class ScoresPlugin extends Plugin {
...
+        }
         embed.setFooter(visitorScore + ' - ' + homeScore);
         embedBuffer.push(embed);
-
       }
 
       //Send embeds in buffer
-      embedBuffer.forEach(e => {
-        message.channel.send(e)
+      embedBuffer.forEach((e) => {
```
c650, line 145: why not `forEach(message.channel.send)`? take advantage of the fact you can pass the actual function, to be called by the implementation of `forEach`

src/app/plugins/scores.plugin.ts
```diff
@@ -62,81 +72,155 @@ export class ScoresPlugin extends Plugin {
...
 
+  private _evaluateScores(a: number, b: number): number {
+    if (a > b) {
+      return 1; //Winning
+    } else if (a < b) {
+      return 0; //Losing
+    } else {
+      return 2; //Tied
+    }
+  }
```
c650, line 165: due to returning in each branch, no need to chain else's:

```
if (a > b) return 1

if (a < b) return 0

return 2
```

c650: better yet:

if (a == b) return 2

return a > b

src/app/plugins/scores.plugin.ts
```diff
@@ -62,81 +72,155 @@ export class ScoresPlugin extends Plugin {
...
+
+    //Make sure data is valid
+    if (teamsData.join(' ') === '' || time.join(' ') === '') {
+      return false;
+    }
+
+    teamsData = teamsData[0].replace('=', '').split('%20'); //Trim and split
+    const matchup = teamsData.join(' ');
+
+    teamsData = teamsData.join(' ').split(' at '); //Eliminate 'at' and combine multi-word team names
```
c650, line 201: why not `matchup.split` instead of joining `teamsData` again?

src/app/plugins/scores.plugin.ts
```diff
@@ -62,81 +72,155 @@ export class ScoresPlugin extends Plugin {
...
+    const [visitorName, homeName] = teamsData;
+
+    //Check if game does not contain all search terms
+    if (
+      !(
+        this._strictlyContainsAllTokens(visitorName, teamArg) ||
+        this._strictlyContainsAllTokens(homeName, teamArg)
+      )
```
c650, line 209: see above comment on similar code

c650 (changes requested):

src/app/plugins/scores.plugin.ts
tanndlin: It checks to see if one of the teams, contains the search term, not both. If either are true i do not want to skip

c650: gotcha, nvm

tanndlin: I changed the comment to represent that a little better

src/app/plugins/scores.plugin.ts
tanndlin: If i replace the condition with a variable storing it, would that be better? or should I try another method than swapping the parameters in _evalScores?

src/app/plugins/scores.plugin.ts
tanndlin: It throws an error when I do `forEach(message.channel.send)`

src/app/plugins/scores.plugin.ts
c650: if (cond) {swap}
evaluateScores...

Better even, if you returned -1,0,1 you could just negate it

c650 (approved):

c650 (approved):

PR #196, opened 2020-10-10
Add Roles Type and optional role requirement to plugin.ts

gooca:


src/common/plugin.ts
```diff
@@ -11,6 +11,8 @@ export abstract class Plugin implements IPlugin {
 
   public abstract get permission(): ChannelType;
 
+  public minRoleToRun?: number;
```
c650, line 14: prefer Role enum

src/common/types.ts
```diff
@@ -139,3 +139,11 @@ export interface ICommandLookup {
...
   [pluginName: string]: IPlugin;
 }
+
+export const Roles: {[role: string]: number} = {
+  "Suspended": -10,
+  "RegularUser": 0,
+  "Teaching Assistant": 20,
+  "Moderator": 30,
+  "Admin": 40,
+}
```
c650, line 149: saa

c650 (changes requested):

src/common/types.ts
gooca: roles is an array like ["dank", "Admin"] and when I have to index over the enum I can't index with random strings with TS

gooca: nm joey said it might lemme check

joey-colon (approved):

PR #199, opened 2020-10-11
Add role.service and role check 

gooca:


src/services/role.service.ts
```diff
@@ -0,0 +1,23 @@
...
+  hasPermission(member: GuildMember, minRoleToRun: number) {
+    let roles = member.roles;
+    let highestRole = this._getHighestRole(roles);
+    return highestRole >= minRoleToRun;
+  }
+
+  private _getHighestRole(roles: Collection<string, Role>) {
+    let highestRole = 0;
+
+    roles.forEach((role: Role) => {
```
joey-colon, line 14: use .reduce()

src/services/role.service.ts
```diff
@@ -0,0 +1,23 @@
+import { RoleType, RoleTypeKey } from './../common/types';
+import { GuildMember, Role, Collection } from 'discord.js';
+
+export class RoleService {
+  hasPermission(member: GuildMember, minRoleToRun: number) {
+    let roles = member.roles;
+    let highestRole = this._getHighestRole(roles);
```
joey-colon, line 7: `const`

joey-colon (approved):

src/services/role.service.ts
gooca: no

PR #202, opened 2020-10-12
Created MarketPlacePlugin

LemensRUs:


src/app/plugins/marketplace.plugin.ts
```diff
@@ -0,0 +1,70 @@
...
+  public pluginAlias = [];
+  public permission: ChannelType = ChannelType.Public;
+  public pluginChannelName: string = Constants.Channels.Public.BuySellTrade;
+
+  constructor(public container: IContainer) {
+    super();
+  }
+
+  public async execute(message: IMessage, args: string[]) {
+    if (!args) {
```
joey-colon, line 19: if args = empty array, it will be truthy, we'll want to check `!args || args.length === 0`

src/app/plugins/marketplace.plugin.ts
```diff
@@ -0,0 +1,70 @@
...
+      message.reply('Invalid syntax.');
+      return;
+    }
+
+    const [sub_command] = args; //set sub command = args[0].
```
joey-colon, line 24: nice

src/app/plugins/marketplace.plugin.ts
```diff
@@ -0,0 +1,70 @@
...
+      this._handleAddMarket(message, description);
+    } else if (sub_command === 'list') {
+      this._handleListMarket(message);
+    } else {
+      message.reply('Invalid command. See !help');
+    }
+  }
+
+  private _handleAddMarket(message: IMessage, description: string) {
+    if (description !== '') {
```
joey-colon, line 37: `if (description)`

joey-colon:
at a high level this looks good, but we'll need to persist this to the mongo db @c650

src/app/plugins/marketplace.plugin.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  private _handleListMarket(message: IMessage) {
+    message.channel.fetchMessages({ limit: 100 }).then((msgs) => {
+      const itemsForSale: String[] = [];
+
+      //iterate over messages.
+      msgs.forEach(function(msg) {
+        const content = msg.content;
+
+        if (content.includes('!marketplace add')) {
```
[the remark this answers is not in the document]
c650: use class constants! also remember that the command prefix `!` is configurable, so you can't always rely on it

```
private _NEW_LISTING_PREFIX = 'marketplace add';
private _LISTING_PREFIX_LEN  = _NEW_LISTING_PREFIX.length + 1;  // +1 for !
```

but also this could break if you introduce an alias for `marketplace`. Maybe you should have a `parseMessage` that handles these nuances for you.

src/app/plugins/marketplace.plugin.ts
```diff
@@ -0,0 +1,78 @@
...
+  private _handleAddMarket(message: IMessage, description: string) {
+    const stringForUser = description
+      ? 'Your item has been added! Please delete your message once it is sold.'
+      : 'Invalid Listing.';
+    const embed = new RichEmbed();
+    embed.setDescription(stringForUser);
+    this._replyToUser(message, embed);
+  }
+
+  private _handleListMarket(message: IMessage) {
```
c650, line 45: add exception handling

src/app/plugins/marketplace.plugin.ts
```diff
@@ -43,29 +51,38 @@ export class MarketPlacePlugin extends Plugin {
   }
 
   private _handleListMarket(message: IMessage) {
-    message.channel.fetchMessages({ limit: 100 }).then((msgs) => {
-      const itemsForSale: String[] = [];
+    try {
```
c650, line 54: you have to await the (fetchMessages.then()...) or have a .catch() at the end

src/app/plugins/marketplace.plugin.ts
```diff
@@ -75,4 +92,7 @@ export class MarketPlacePlugin extends Plugin {
       await message.reply(embed);
     }
   }
+  private _parseMessage(args: string[]) {
+    console.log(args);
+  }
```
c650, line 97: ?

src/app/plugins/marketplace.plugin.ts
```diff
@@ -27,10 +27,8 @@ export class MarketPlacePlugin extends Plugin {
       message.reply('Invalid syntax.');
       return;
     }
-    this._parseMessage(args);
 
-    const [sub_command] = args; //set sub command = args[0].
-    const description = message.content.slice(this._NEW_LISTING_PREFIX.length + 2);
+    const [sub_command, description] = args; //set sub command = args[0].
```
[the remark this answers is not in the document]
tanndlin: We can, but it technically doesn't matter. All we do with `descrption` is make sure it is not an empty string. Then we dump it, bc list market reads from the original message that added it

c650 (approved):
lgtm

PR #206, opened 2020-10-18
Send roles info through files instead msg.content

### LINKED ISSUE #204: ManageRoles patch too-long message... again

c650:
Requires /tmp, so docker should be run with --tmpfs /tmp

Fixes #204


joey-colon (approved):
await the fs operation

PR #223, opened 2020-11-08
DM Anon Reports

### LINKED ISSUE #115: Report user offense via DM
Allows whistleblowers to be anon and lets staff not have to have open DM's.

c650:
Fixes #115


src/app/plugins/anonreport.plugin.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  public usableInDM = true;
+  public usableInGuild = false;
+
+  constructor(public container: IContainer) {
+    super();
+  }
+
+  public async execute(message: IMessage, args?: string[]) {
+    this.container.modService.fileAnonReport(message).then(() => {
```
joey-colon, line 19: handle the case when the promise fails

joey-colon (approved):

src/app/plugins/anonreport.plugin.ts
c650: Thanks I forgot will fix

Issue #213, opened 2020-10-27
AddRoles Timeout, not seen before


c650:
We've become aware of some new behaviour. The last deployment of the bot was earlier today, https://github.com/joey-colon/lion/commit/77b56ae5d030be2734e7ee202cc321601ed5070e 

Adding a role times out. But the role shows up as added (sometimes?).
```
Oct 27 09:46:59 f30c2d0b80a6 default error User xxx attempted to add the role IVoted but failed: Error: Adding the role timed out. 
Oct 27 09:46:59 f30c2d0b80a6 default error     { 
Oct 27 09:46:59 f30c2d0b80a6 default error       message: 'User xxx attempted to add the role IVoted but failed: Error: Adding the role timed out.', 
Oct 27 09:46:59 f30c2d0b80a6 default error       level: 'error', 
Oct 27 09:46:59 f30c2d0b80a6 default error       timestamp: '2020-10-27T16:46:59.589Z', 
Oct 27 09:46:59 f30c2d0b80a6 default error       [Symbol(level)]: 'error', 
Oct 27 09:46:59 f30c2d0b80a6 default error       [Symbol(message)]: '{"message":"User xxx attempted to add the role IVoted but failed: Error: Adding the role timed out.","level":"error","timestamp":"2020-10-27T16:46:59.589Z"}' 
Oct 27 09:46:59 f30c2d0b80a6 default error     } 
Oct 27 12:30:16 f30c2d0b80a6 default error User xxx attempted to add the role IVoted but failed: Error: Adding the role timed out. 
Oct 27 12:30:16 f30c2d0b80a6 default error     { 
Oct 27 12:30:16 f30c2d0b80a6 default error       message: 'User xxx attempted to add the role IVoted but failed: Error: Adding the role timed out.', 
Oct 27 12:30:16 f30c2d0b80a6 default error       level: 'error', 
Oct 27 12:30:16 f30c2d0b80a6 default error       timestamp: '2020-10-27T19:30:16.022Z', 
Oct 27 12:30:16 f30c2d0b80a6 default error       [Symbol(level)]: 'error', 
Oct 27 12:30:16 f30c2d0b80a6 default error       [Symbol(message)]: '{"message":"User xxx attempted to add the role IVoted but failed: Error: Adding the role timed out.","level":"error","timestamp":"2020-10-27T19:30:16.022Z"}' 
Oct 27 12:30:16 f30c2d0b80a6 default error     } 
[...]

PR #224, opened 2020-11-08
Add unverified role if default pfp

### LINKED ISSUE #194: Auto-detect default-avatar accounts and mark Un-verified
Subscribe to user join guild event

Get the avatar image, detect if its the default avatar, flag if so

tanndlin:
fixes #194

### DIFF
--- a/src/app/handlers/auto_unverify.handler.ts
+++ b/src/app/handlers/auto_unverify.handler.ts
@@ -1,0 +1,18 @@
+import { IContainer, IHandler, IMessage } from '../../common/types';
+
+export class AutoUnverifyHandler implements IHandler {
+  constructor(public container: IContainer) {}
+
+  public async execute(message: IMessage): Promise<void> {
+    const avatar = message.author.avatar; //This is null if they have default avatar
+
+    if (!avatar) {
+      const unverifiedRole = message.guild.roles.filter((role) => role.name === 'Un verified');
+      const targetGuildMember = message.guild.members
+        .filter((user) => user.id === message.author.id)
+        .first();
+
+      targetGuildMember.addRoles(unverifiedRole);
+    }
+  }
+}
--- a/src/app/handlers/auto_unverify.handler.ts
+++ b/src/app/handlers/auto_unverify.handler.ts
@@ -1,16 +1,16 @@
-import { IContainer, IHandler, IMessage } from '../../common/types';
+import { GuildMember } from 'discord.js';
+import { IContainer, IHandler } from '../../common/types';
 
 export class AutoUnverifyHandler implements IHandler {
   constructor(public container: IContainer) {}
 
-  public async execute(message: IMessage): Promise<void> {
-    const hasAvatar = Boolean(message.author.avatar);
+  public async execute(member: GuildMember): Promise<void> {
+    const hasAvatar = Boolean(member.user.avatar);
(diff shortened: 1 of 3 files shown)


tanndlin:
Works in test server, but make sure I did the whole "handler" thing right

src/app/handlers/auto_unverify.handler.ts
```diff
@@ -0,0 +1,18 @@
+import { IContainer, IHandler, IMessage } from '../../common/types';
+
+export class AutoUnverifyHandler implements IHandler {
+  constructor(public container: IContainer) {}
+
+  public async execute(message: IMessage): Promise<void> {
+    const avatar = message.author.avatar; //This is null if they have default avatar
```
joey-colon, line 7: this can have a more descriptive name so we can remove the comment

`const hasAvatar = Boolean(message.author.avatar);`

src/app/handlers/auto_unverify.handler.ts
```diff
@@ -0,0 +1,18 @@
...
+
+    if (!avatar) {
```
joey-colon, line 9: nothing is necessarily wrong here, but a good pattern u should adopt is the use of guard clauses.

more spceifically, a guard clause basically refers to having guards against your code with early returns at the top, and at the bottom of the code is your "happy" case.

for this, it would be something like

```
if hasAvatar
  return;

// execute the code since the happy case here is not having an avatar
```

it's not that obvious the benefits of using a guard clause here, but its something that will come up again and it will make your code way more maintainable.

joey-colon (approved):

src/app/handlers/auto_unverify.handler.ts
tanndlin: Yeah, I've been learning that from Corgo, didn't think ab it here cuz the "Happy Case" wasn't large. Will change

src/app/handlers/auto_unverify.handler.ts
```diff
@@ -0,0 +1,19 @@
...
+  constructor(public container: IContainer) {}
+
+  public async execute(message: IMessage): Promise<void> {
+    const hasAvatar = Boolean(message.author.avatar);
+    if (hasAvatar) {
+      return;
+    }
+
+    const unverifiedRole = message.guild.roles.filter((role) => role.name === 'Un verified');
+    const targetGuildMember = message.guild.members
```
c650, line 13: Could be slow

tanndlin: Are you referring to #213 ?
I literally have like 40 roles, all called 'new role' in my test server, and the addRole is almost instant to the user joining

c650: https://github.com/joey-colon/lion/blob/fb013501794232b15131aae7a5c683aa42c892e1/src/app/listener.ts#L54 Try this flow to look up member

tanndlin:
Dont merge, just found a bug

c650 (approved):
Lgtm

c650:
Did you check that it doesn’t mark Un verified when the user has custom avatar? And also did you check that “Un verified” is correct spelling

tanndlin:
Any custom avatar will return a non-null value with `member.author.avatar`
And yes, double checked spelling of the role

PR #228, opened 2020-11-09
Force Acknowledge of CoC

### LINKED ISSUE #151: use bot to enforce acknowledgement of CoC
Have all users blocked from viewing channels on join -- only CoC channel is visible. Users must acknowledge via message reaction. Thereafter, user is permitted to see other channels.

tanndlin:
fixes #151
This works except when I add https://github.com/tanndlin/lion/blob/8fdfebccee33323ede4d417346e82579d242947d/src/services/handler.service.ts#L27
The auto unverify stops working. Did I not do that right?


src/app/handlers/acknowledge.handler.ts
```diff
@@ -0,0 +1,26 @@
+import { MessageReaction, User } from 'discord.js';
+import { IContainer, IHandler } from '../../common/types';
+
+export class AcknowledgeHandler implements IHandler {
+  constructor(public container: IContainer) {}
+
+  public async execute(reaction: MessageReaction, user: User): Promise<void> {
+    //Get Member from user ID
+    const member = this.container.guildService.get().members.find((m) => m.id === user.id);
```
c650, line 9: Try https://discord.js.org/#/docs/main/stable/class/GuildMemberManager?scrollTo=fetch

```js
members.fetch(user.id)
```
should be faster than O(n) search

src/app/handlers/acknowledge.handler.ts
```diff
@@ -0,0 +1,26 @@
...
+    const CoC_Channel = member.guild.channels.find((c) => c.name === 'code_of_conduct');
```
c650, line 10: https://github.com/joey-colon/lion/blob/master/src/common/constants.ts#L8

src/app/handlers/acknowledge.handler.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    if (reaction.message.channel != CoC_Channel) {
```
c650, line 12: !==

src/app/handlers/acknowledge.handler.ts
```diff
@@ -0,0 +1,26 @@
...
+      return;
+    }
+
+    const role = member.guild.roles
+      .filter((r) => r.name.toLowerCase() === 'unacknowledged')
+      .first();
+    const hasRole = Boolean(member.roles.filter((r) => r === role).array().length);
```
c650, line 19: Use `Array.some()` not filter.length

src/app/handlers/acknowledge.handler.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    if (!hasRole) {
+      return;
+    }
+    member.removeRole(role);
```
c650, line 24: might as well 

```js
hasRole && member.removeRole
```

tanndlin:
The auto unverify runs, all the way until the point where it adds the role, and the role never gets added

src/app/handlers/auto_unacknowledge.handler.ts
```diff
@@ -0,0 +1,13 @@
+import { GuildMember } from 'discord.js';
+import { IContainer, IHandler } from '../../common/types';
+
+export class AutoAcknowledgeHandler implements IHandler {
+  constructor(public container: IContainer) {}
+
+  public async execute(member: GuildMember): Promise<void> {
+    const unacknowledgedRole = member.guild.roles.filter(
+      (role) => role.name.toLowerCase() === 'unacknowledged'
+    );
```
c650, line 10: can we do it in constructor to save it?

src/app/handlers/acknowledge.handler.ts
```diff
@@ -0,0 +1,26 @@
...
+    const member = this.container.guildService.get().members.find((m) => m.id === user.id);
+    const CoC_Channel = member.guild.channels.find((c) => c.name === 'code_of_conduct');
+
+    if (reaction.message.channel != CoC_Channel) {
+      return;
+    }
+
+    const role = member.guild.roles
+      .filter((r) => r.name.toLowerCase() === 'unacknowledged')
+      .first();
```
c650, line 18: calc and save in constructor

src/app/listener.ts
```diff
@@ -24,6 +25,12 @@ export class Listener {
     this.container.clientService.on('ready', () => {
       this.container.loggerService.info(`Loaded ${this.container.jobService.size()} jobs...`);
       this.container.loggerService.info('Lion is now running!');
+
+      //Cache messages in #code_of_conduct to enable the onMessageReact Handler for them
+      const cocChannel = this.container.guildService
+        .get()
+        .channels.find((c) => c.name == 'code_of_conduct') as TextChannel;
+      cocChannel.fetchMessages({ limit: 10 });
```
c650, line 33: this is necessary?

c650:
https://github.com/joey-colon/lion/issues/213 @tanndlin

src/app/listener.ts
tanndlin: Yes, `on('messageReactionAdd'...)` only works on cached messages. Message are only cached if fetched, or sent after the bot starts tunning

c650: gotcha.

src/app/handlers/auto_unacknowledge.handler.ts
tanndlin: Bot Crashes when i try that

src/app/handlers/acknowledge.handler.ts
tanndlin: saa

src/app/handlers/acknowledge.handler.ts
tanndlin: guild.members is a `Collection` not a `GuildMemberManager`, so it doesn't have a fetch method

src/app/handlers/auto_unacknowledge.handler.ts
c650: okay then consider calcing it here but then saving the result

src/app/handlers/acknowledge.handler.ts
c650: if it's a `Collection` then the `.get()` should take the user id (type `Snowflake`)

src/app/handlers/auto_unacknowledge.handler.ts
tanndlin: Bot crashes
`Property '_unacknowledgedRole' has no initializer and is not definitely assigned in the constructor`

tanndlin: could change to
`const unacknowledgedRole = member.guild.roles.get('unacknowledged');`

tanndlin: nvm, doesn't work

tanndlin:
If I add both roles inside of auto_unverify.handler, it works, both roles are added. Should I do that and delete auto_unacknowledge.handler?
Not ideal but the only thing that works so far

src/app/handlers/auto_unacknowledge.handler.ts
c650: > Property '_unacknowledgedRole' has no initializer and is not definitely assigned in the constructor

means you have to initialise it to `undefined` at the very least. Then in the `execute` you check if `undefined`, and if so load the role; otherwise continue

c650:
> If I add both roles inside of auto_unverify.handler, it works, both roles are added. Should I do that and delete auto_unacknowledge.handler?
Not ideal but the only thing that works so far

I'd say this is a good idea if you refactor and name it `new_guildmember.handler`

src/app/handlers/new_member.handler.ts
```diff
@@ -0,0 +1,33 @@
+import { GuildMember } from 'discord.js';
+import { IContainer, IHandler } from '../../common/types';
+
+export class NewMemberHandler implements IHandler {
+  private _unverifiedRole: any = undefined;
+  private _unacknowledgedRole: any = undefined;
```
joey-colon, line 6: we don't *have* to use `any` here. it's better to use something like

`RoleType | undefined` (not sure what discord role type is).

a common abstraction that people do, and might be worth to transition into now, is creating a type `Maybe`

```
type Maybe<T> = T | undefined | null;
``` 

and then ur type for this variable would just be `Maybe<RoleType>`

tanndlin: Put this in the same file on top of the class? or?

c650: types.ts makes sense

src/app/handlers/acknowledge.handler.ts
```diff
@@ -0,0 +1,32 @@
+import { MessageReaction, User } from 'discord.js';
+import { IContainer, IHandler } from '../../common/types';
+import Constants from '../../common/constants';
+
+export class AcknowledgeHandler implements IHandler {
+  constructor(public container: IContainer) {}
+
+  public async execute(reaction: MessageReaction, user: User): Promise<void> {
+    //Get Member from user ID
```
joey-colon, line 9: redundant comment. let the code speaks for itself

src/app/handlers/acknowledge.handler.ts
```diff
@@ -0,0 +1,32 @@
...
+    const member = this.container.guildService.get().members.get(user.id);
+    if (!member) {
+      return;
+    }
+
+    const CoC_Channel = member.guild.channels.find(
+      (c) => c.name === Constants.Channels.Public.CodeOfConduct
+    );
```
c650, line 17: Cache this

src/app/handlers/acknowledge.handler.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    if (reaction.message.channel != CoC_Channel) {
+      return;
+    }
+
+    const role = member.guild.roles
+      .filter((r) => r.name.toLowerCase() === 'unacknowledged')
+      .first();
```
c650, line 25: cache this

src/app/handlers/new_member.handler.ts
```diff
@@ -0,0 +1,34 @@
...
+  private _unacknowledgedRole: Maybe<Role> = undefined;
+
+  constructor(public container: IContainer) {}
+
+  public async execute(member: GuildMember): Promise<void> {
+    if (!this._unacknowledgedRole) {
+      this._unacknowledgedRole = member.guild.roles
+        .filter((role) => role.name === 'Unacknowledged')
+        .first();
+    }
```
c650, line 16: might be useful to refactor into a mapping

```js
rolesToAdd = {
  "unack...": ...
};

// later in `execute`
if (cacheMiss) {
  member.guild.roles.filter( r => Object.keys(rolesToadd).some( r.name ));
}
```

src/app/handlers/acknowledge.handler.ts
```diff
@@ -1,32 +1,39 @@
...
-      (c) => c.name === Constants.Channels.Public.CodeOfConduct
-    );
+    if (!this._CoC_Channel) {
+      this._CoC_Channel = member.guild.channels.find(
+        (c) => c.name === Constants.Channels.Public.CodeOfConduct
+      );
+    }
 
-    if (reaction.message.channel != CoC_Channel) {
+    if (reaction.message.channel != this._CoC_Channel) {
```
c650, line 24: !==

src/app/handlers/acknowledge.handler.ts
```diff
@@ -1,32 +1,39 @@
...
-    const role = member.guild.roles
-      .filter((r) => r.name.toLowerCase() === 'unacknowledged')
-      .first();
-    const hasRole = Boolean(member.roles.array().some((r) => r === role));
+    if (!this._targetRole) {
+      this._targetRole = member.guild.roles
+        .filter((r) => r.name.toLowerCase() === 'unacknowledged')
+        .first();
+    }
+    const hasRole = Boolean(member.roles.array().some((r) => r === this._targetRole));
```
c650, line 33: wrapping in `Boolean` not necessary because `some()` returns boolean

src/app/handlers/new_member.handler.ts
```diff
@@ -1,32 +1,33 @@
...
 
   public async execute(member: GuildMember): Promise<void> {
-    if (!this._unacknowledgedRole) {
-      this._unacknowledgedRole = member.guild.roles
-        .filter((role) => role.name === 'Unacknowledged')
-        .first();
+    if (!this._roleCache.unacknowledged || !this._roleCache['Un verified']) {
+      Object.keys(this._roleCache).map((key: string) => {
+        this._roleCache[key] = member.guild.roles.filter((r) => r.name === key).first();
+      });
```
c650: no need to `map()` since you are essentially doing a `forEach`

src/app/handlers/new_member.handler.ts
```diff
@@ -1,32 +1,33 @@
...
     }
-    member.addRole(this._unacknowledgedRole);
+
+    //Required to remove optional | undefined
+    if (!this._roleCache.Unacknowledged || !this._roleCache['Un verified']) {
```
c650, line 20: shoudlnt check unverified here

src/app/handlers/new_member.handler.ts
```diff
@@ -1,32 +1,33 @@
...
+      return;
+    }
+    member.addRole(this._roleCache.Unacknowledged);
```
c650, line 24: should check for unverified role in cache here.

src/app/handlers/new_member.handler.ts
c650: the lack of Unverified role in teh guild shouldnt prevent us from adding unack role if it exists

c650: bump?

tanndlin: My bad, resolved wrong convo and forgot ab it

c650 (dismissed):
lgtm

src/app/handlers/new_member.handler.ts
```diff
@@ -0,0 +1,38 @@
+import { GuildMember, Role } from 'discord.js';
+import { IContainer, IHandler } from '../../common/types';
+import { Maybe } from '../../common/types';
+
+export class NewMemberHandler implements IHandler {
+  private _roleCache: Record<string, Maybe<Role>> = {
+    'Un Acknowledged': undefined,
```
joey-colon, line 7: personal preference, i would create something like `const UNACKNOWLEDGE_ROLE = 'Un Acknowledged';` since you reference this string 4 times. if you ever have to change what this string is, it would only be in one place you need to change instead of 4 places.

tanndlin: When i do this, the type of the role does not turn into a `Role` rather than `Maybe<Role>` after the check for undefined.
With this code
```
if (!this._roleCache[this._UNACKNOWLEDGED_ROLE]) {
      return;
}
const role = this._roleCache[this._UNACKNOWLEDGED_ROLE];
```
role is still type `Maybe<Role>`
image

tanndlin: `this._UNACKNOWLEDGED_ROLE` is a string

tanndlin: Actually, I could just assert it as a role, since I know it's not undefined after the if

joey-colon: yeah i would just tell typescript it's a role

`const role = this._roleCache[this._UNACKNOWLEDGED_ROLE] as Role`

tanndlin: The `const role = ...` was redundant,I just did that to show it's type. 
I asserted in inline `member.addRole(<Role>this._roleCache[this._UNACKNOWLEDGED_ROLE]);`

src/app/handlers/acknowledge.handler.ts
```diff
@@ -0,0 +1,37 @@
...
+        (c) => c.name === Constants.Channels.Public.CodeOfConduct
+      );
+    }
+
+    if (reaction.message.channel !== this._CoC_Channel) {
+      return;
+    }
+
+    if (!this._targetRole) {
+      this._targetRole = member.guild.roles.filter((r) => r.name === 'Un Acknowledged').first();
```
c650, line 29: i thought you had changed 'Un Acknowledged' to a class constant?

tanndlin: The `Maybe<Role>` is a class property, you want me to move the `name:string` up there too?

tanndlin: That was a class constant in new_member.handler

src/app/handlers/new_member.handler.ts
```diff
@@ -0,0 +1,41 @@
...
+    'Un verified': undefined,
+  };
+  private _UNACKNOWLEDGED_ROLE: string = 'Un Acknowledged';
+  private _UNVERIFIED_ROLE: string = 'Un verified';
```
c650, line 11: use these to make the Record above, 

```
private _roleCache: Record<string, Maybe<Role>> = {
    _UNACKNOWLEDGED_ROLE: undefined,
```

src/app/listener.ts
```diff
@@ -24,6 +25,12 @@ export class Listener {
     this.container.clientService.on('ready', () => {
       this.container.loggerService.info(`Loaded ${this.container.jobService.size()} jobs...`);
       this.container.loggerService.info('Lion is now running!');
+
+      //Cache messages in #code_of_conduct to enable the onMessageReact Handler for them
+      const cocChannel = this.container.guildService
+        .get()
+        .channels.find((c) => c.name == 'code_of_conduct') as TextChannel;
```
c650, line 32: ===

c650: Constants.Channels.Public.CodeOfConduct

src/app/handlers/acknowledge.handler.ts
```diff
@@ -5,11 +5,21 @@ import { Maybe } from '../../common/types';
 
 export class AcknowledgeHandler implements IHandler {
   private _CoC_Channel: Maybe<GuildChannel> = undefined;
-  private _targetRole: Maybe<Role> = undefined;
+  private _roleCache: Record<string, Maybe<Role>> = {
+    'Un Acknowledged': undefined,
```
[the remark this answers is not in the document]
tanndlin: No you cannot, object member names cannot be dynamically created like that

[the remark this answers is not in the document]
joey-colon: `[var_name]: value`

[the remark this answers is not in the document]
tanndlin: Ok, sweet! Didn't know you could do that. I'll do that in new_member.handler as well

c650 (approved):

src/app/handlers/new_member.handler.ts
```diff
@@ -0,0 +1,41 @@
+import { GuildMember, Role } from 'discord.js';
+import { IContainer, IHandler } from '../../common/types';
+import { Maybe } from '../../common/types';
```
joey-colon, line 3: combine these imports

src/app/handlers/acknowledge.handler.ts
```diff
@@ -0,0 +1,46 @@
+import { GuildChannel, MessageReaction, User, Role } from 'discord.js';
+import { IContainer, IHandler } from '../../common/types';
+import Constants from '../../common/constants';
+import { Maybe } from '../../common/types';
```
joey-colon, line 4: ditto

joey-colon (approved):

PR #233, opened 2020-11-21
Register multiple classes at once

tanndlin:


c650 (approved):

src/app/plugins/register.plugin.ts
```diff
@@ -17,16 +17,34 @@ export class RegisterPlugin extends Plugin {
...
-    } catch (e) {
-      message.reply(e);
+
+    const successfulClasses: string[] = [];
+    const invalidClasses: string[] = [];
+    for (const arg of args) {
+      const request = this.container.classService.buildRequest(message.author, [arg]);
+      if (!request) {
+        message.reply('I was unable to build your request.');
+        return;
```
joey-colon, line 30: `continue;`

joey-colon (approved):

src/app/plugins/register.plugin.ts
```diff
@@ -17,16 +17,34 @@ export class RegisterPlugin extends Plugin {
...
-      message.reply(response);
-    } catch (e) {
-      message.reply(e);
+
+    const successfulClasses: string[] = [];
+    const invalidClasses: string[] = [];
+    for (const arg of args) {
+      const request = this.container.classService.buildRequest(message.author, [arg]);
+      if (!request) {
+        message.reply('I was unable to build your request.');
```
joey-colon, line 29: also might not need this.

src/app/plugins/register.plugin.ts
```diff
@@ -17,16 +17,34 @@ export class RegisterPlugin extends Plugin {
...
+          successfulClasses.push(arg);
+        } else {
+          invalidClasses.push(arg);
+        }
+      } catch (e) {
+        message.reply(e);
+      }
+    }
+
+    let messageForUser = `Successfully added to ${successfulClasses.length} classes`;
```
joey-colon, line 44: then u will need to handle the cases of when no classes were added successfully (just have specific message if len == 0)

src/app/plugins/register.plugin.ts
c650: wouldnt want this to be sent n times

tanndlin: So check if it failed for each arg, then reply to the user, outside of the for loop, that one or more requests failed?

c650: include it in the array of classes that could not be added (no 'success')

joey-colon (approved):

src/app/plugins/register.plugin.ts
```diff
@@ -17,16 +17,40 @@ export class RegisterPlugin extends Plugin {
...
+      try {
+        const response = await this.container.classService.register(request);
+        if (response.includes('success')) {
+          successfulClasses.push(arg);
+        } else {
+          invalidClasses.push(arg);
+        }
+      } catch (e) {
+        message.reply(e);
+      }
```
c650, line 41: we dont want this to happen n times.

tanndlin: Should I just push the arg to `invalidClasses` and `this.container.loggerService.error(e)`?

src/app/plugins/register.plugin.ts
```diff
@@ -17,16 +17,46 @@ export class RegisterPlugin extends Plugin {
...
+      try {
+        const response = await this.container.classService.register(request);
+        if (response.includes('success')) {
+          successfulClasses.push(arg);
+        } else {
+          invalidClasses.push(arg);
+        }
+      } catch (e) {
+        invalidClasses.push(arg);
+        error = e;
```
c650, line 42: just log the error with logger service. dont expose the actual error to the user

tanndlin: I can, but before, it just replied to the user with the error

c650: you shouldnt expose the actual exception to the user

tanndlin: Ok, should I log each error just in case, or just do one log with an error per call

c650: just log the error when it occurs that's all

c650 (approved):
:confetti_ball:

PR #237, opened 2020-11-25
Fixed Status Plugin

tanndlin:
Changed it so that it parses the GitHub website for data, that way it works in Docker

### DIFF
--- a/src/app/plugins/status.plugin.ts
+++ b/src/app/plugins/status.plugin.ts
@@ -14,6 +14,16 @@ export class StatusPlugin extends Plugin {
     'https://cdn.discordapp.com/avatars/574623716638720000/7d404c72a6fccb4a3bc610490f8d7b72.png';
   private REPO_URL = 'https://github.com/joey-colon/lion/commit/';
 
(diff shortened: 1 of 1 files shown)


c650 (commented):
but the latest commit on github is not necessarily what the bot is running ..?

tanndlin:
Not necessarily, but usually. I'm not sure what I can do, code-wise, to make it work inside of docker. I think that's something you would have to fiddle around with to get the old code working.

tanndlin:
Oh! What we could do, is get all this info on startup. And just spit it out on `!status` call. It still, wont be *necessarily* what the bot is running, but how often does the bot really reset without taking on the latest commit?

c650:
There’s a way to do it when you build the docker image. That’s probably the best route

tanndlin:
Is that something I need to do? I'm not sure how to go about that.

tanndlin:
I'm lost on how to do it through docker, should I close this PR for now?

c650:
Sure yeah no worries

tanndlin:
Closing until I find a solution

PR #239, opened 2020-11-29
Fixed Sports Plugin for cancelled games

tanndlin:


tanndlin:
When a game was cancelled, for some reason, the game wouldn't end with the old delimiter, causing it to be parsed correctly Lion. I found  a new string that no matter what, each entry has

c650 (approved):

PR #240, opened 2020-11-30
Shadowban plugin

### LINKED ISSUE #222: server wide shadow ban
ability to ban a user from all non class channels

tanndlin:
fixes #222
Removes user ability to see all general text channels, and voice channels
They can still see class channels, and the `UCF CS, ECE, AND IT` category

### DIFF
--- a/src/app/plugins/shadowban.plugin.ts
+++ b/src/app/plugins/shadowban.plugin.ts
@@ -1,7 +1,14 @@
 import { Plugin } from '../../common/plugin';
 import { IContainer, IMessage, ChannelType } from '../../common/types';
 import Constants from '../../common/constants';
-import { CategoryChannel, Collection, GuildChannel, User } from 'discord.js';
+import {
+  CategoryChannel,
+  Collection,
+  Guild,
+  GuildChannel,
+  PermissionOverwrites,
+  User,
+} from 'discord.js';
 
 export class ShadowBanPlugin extends Plugin {
   public name: string = 'Shadowban Plugin';
@@ -42,11 +49,11 @@ export class ShadowBanPlugin extends Plugin {
     }
 
     if (subCommand === 'ban') {
-      this._applyToChannels(user, this._banUser);
+      this._applyToChannels(this._banUser(user));
       message.reply(`${user.tag} has been shadowbanned`);
       return;
     } else if (subCommand === 'unban') {
-      this._applyToChannels(user, this._unbanUser);
+      this._applyToChannels(this._unbanUser(user));
       message.reply(`${user.tag} has been unshadowbanned`);
       return;
     } else {
@@ -54,7 +61,7 @@ export class ShadowBanPlugin extends Plugin {
     }
   }
 
-  private _applyToChannels(user: User, callback: (chan: GuildChannel, user: User) => void): void {
+  private async _applyToChannels(callback: (chan: GuildChannel) => void) {
     const categories = this.container.guildService
       .get()
       .channels.filter((chan) => chan.type === 'category') as Collection<string, CategoryChannel>;
@@ -64,20 +71,25 @@ export class ShadowBanPlugin extends Plugin {
       return this.BANNED_CATEGORIES.some((n) => chanName === n);
     });
 
-    catsToBan
-      .reduce(function(acc: GuildChannel[], cat: CategoryChannel) {
-        return [...acc, ...cat.children.array()];
-      }, [])
-      .forEach((chan) => callback(chan, user));
+    const promises = catsToBan.reduce((acc, cat: CategoryChannel) => {
+      acc.push(...cat.children.array().map((chan: GuildChannel) => callback(chan)));
+      return acc;
+    }, []);
+
+    await Promise.all(promises);
   }
 
-  private _banUser(chan: GuildChannel, user: User) {
-    chan.overwritePermissions(user, {
-      VIEW_CHANNEL: false,
-    });
+  private _banUser(user: User) {
+    return (chan: GuildChannel) => {
+      chan.overwritePermissions(user, {
+        VIEW_CHANNEL: false,
+      });
+    };
   }
 
-  private _unbanUser(chan: GuildChannel, user: User) {
-    chan.permissionOverwrites.get(user.id)?.delete();
+  private _unbanUser(user: User) {
+    return (chan: GuildChannel) => {
+      chan.permissionOverwrites.get(user.id)?.delete();
+    };
   }
 }
--- a/src/app/plugins/shadowban.plugin.ts
+++ b/src/app/plugins/shadowban.plugin.ts
@@ -1,7 +1,7 @@
 import { Plugin } from '../../common/plugin';
 import { IContainer, IMessage, ChannelType } from '../../common/types';
 import Constants from '../../common/constants';
-import { CategoryChannel, Collection, User } from 'discord.js';
+import { CategoryChannel, Collection, GuildChannel, User } from 'discord.js';
 
 export class ShadowBanPlugin extends Plugin {
   public name: string = 'Shadowban Plugin';
@@ -73,7 +73,7 @@ export class ShadowBanPlugin extends Plugin {
     message.reply(`${user.tag} has been unshadowbanned`);
   }
 
-  private _getChannelsToBan(): CategoryChannel[] {
+  private _getChannelsToBan(): GuildChannel[] {
     const catsToBan: CategoryChannel[] = [];
     const categories: Collection<
       string,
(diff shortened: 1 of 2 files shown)


src/app/plugins/shadowban.plugin.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  public validate(message: IMessage, args: string[]) {
+    return args && args.length > 1;
+  }
+
+  public async execute(message: IMessage, args?: string[]) {
+    if (!args) {
+      return;
+    }
```
c650, line 34: should be able to skip this by making the param not `?` and `validate` is always called before execute

src/app/plugins/shadowban.plugin.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    const [subCommand, ...targetUser] = args;
+    const member = this.container.guildService
+      .get()
+      .members.filter((m) => m.user.tag === targetUser.join(' '))
```
c650, line 39: dont call the join() N times

src/app/plugins/shadowban.plugin.ts
```diff
@@ -0,0 +1,94 @@
...
+    return args && args.length > 1;
+  }
+
+  public async execute(message: IMessage, args?: string[]) {
+    if (!args) {
+      return;
+    }
+
+    const [subCommand, ...targetUser] = args;
+    const member = this.container.guildService
```
c650, line 37: feel free to use the `resolveUser` from Moderation Service.

src/app/plugins/shadowban.plugin.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    if (subCommand === 'ban') {
+      this._banUser(message, member.user);
+      return;
+    } else if (subCommand === 'unban') {
+      this._unbanUser(message, member.user);
+      return;
+    } else {
+      message.reply('Invalid subcommand\nTry: `!shadowban <ban|unban> <user>`');
```
c650, line 54:
```
${this.description}
```
instead of hardcode

src/app/plugins/shadowban.plugin.ts
```diff
@@ -0,0 +1,94 @@
...
+    const catsToBan: CategoryChannel[] = [];
+    const categories: Collection<
+      string,
+      CategoryChannel
+    > = this.container.guildService
+      .get()
+      .channels.filter((chan) => chan.type === 'category') as Collection<string, CategoryChannel>;
+
+    categories.forEach((cat) => {
+      if (!this.BANNED_CATEGORIES.some((n) => cat.name.toUpperCase() === n)) {
```
c650, line 86: only UpperCase once

src/app/plugins/shadowban.plugin.ts
```diff
@@ -0,0 +1,94 @@
...
+  private _getChannelsToBan(): CategoryChannel[] {
+    const catsToBan: CategoryChannel[] = [];
+    const categories: Collection<
+      string,
+      CategoryChannel
+    > = this.container.guildService
+      .get()
+      .channels.filter((chan) => chan.type === 'category') as Collection<string, CategoryChannel>;
+
+    categories.forEach((cat) => {
```
c650, line 85: reduce()

src/app/plugins/shadowban.plugin.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  private _unbanUser(message: IMessage, user: User) {
+    this._getChannelsToBan().forEach((chan) => {
+      chan.permissionOverwrites.get(user.id)?.delete();
+    });
+
+    message.reply(`${user.tag} has been unshadowbanned`);
+  }
+
+  private _getChannelsToBan(): CategoryChannel[] {
```
c650, line 76: you could make this method take a parameter that is a function F(Channel) -> void and then at the end use reduce to apply that function... then in ban and unban you dont need the forEach explicitly.

c650 (commented):
does this work if a channel within a category doesnt have perms synced?

tanndlin:
No, I was going to do a channel by channel approach but figured all the channels were synced and it would be neater in code this way.

tanndlin:
When I get back home I'll look into it

src/app/plugins/shadowban.plugin.ts
```diff
@@ -0,0 +1,94 @@
...
+      this._unbanUser(message, member.user);
+      return;
+    } else {
+      message.reply('Invalid subcommand\nTry: `!shadowban <ban|unban> <user>`');
+    }
+  }
+
+  private _banUser(message: IMessage, user: User) {
+    this._getChannelsToBan().forEach((chan) => {
+      chan.overwritePermissions(user, {
```
c650, line 60: if this is async we should do a map and then await Promise.all

tanndlin:
Doing a callback was the weirdest thing I've ever done, but I kinda like it 🤔

src/app/plugins/shadowban.plugin.ts
```diff
@@ -0,0 +1,83 @@
...
+      .get()
+      .channels.filter((chan) => chan.type === 'category') as Collection<string, CategoryChannel>;
+
+    const catsToBan = categories.filter((cat: CategoryChannel) => {
+      const chanName = cat.name.toUpperCase();
+      return this.BANNED_CATEGORIES.some((n) => chanName === n);
+    });
+
+    catsToBan
+      .reduce(function(acc: GuildChannel[], cat: CategoryChannel) {
```
c650, line 68: prefer `() =>` over `function ()`

src/app/plugins/shadowban.plugin.ts
```diff
@@ -0,0 +1,83 @@
...
+        return [...acc, ...cat.children.array()];
```
c650, line 69: why not 

```
acc.push(...cat.children.array())
return acc
```

src/app/plugins/shadowban.plugin.ts
```diff
@@ -0,0 +1,83 @@
...
+      }, [])
+      .forEach((chan) => callback(chan, user));
```
c650, line 71: might as well do this part in the reduce

src/app/plugins/shadowban.plugin.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  private _banUser(chan: GuildChannel, user: User) {
```
c650, line 74: for extra coolness:

```
private _banUser(user: User) { 
  return (chan: GuildChannel) => {
    chan.overwritePermissions(user, ...);
  };
}
```

then you can pass `_banUser(user)` and then the callback is `callback(chan)`...

or better yet! 
```
chans.forEach(callback)
```

src/app/plugins/shadowban.plugin.ts
```diff
@@ -0,0 +1,83 @@
...
+    chan.overwritePermissions(user, {
+      VIEW_CHANNEL: false,
+    });
+  }
+
+  private _unbanUser(chan: GuildChannel, user: User) {
```
c650, line 80: saa

src/app/plugins/shadowban.plugin.ts
tanndlin: I'm struggling to see how to do it in the reduce,
`acc,push(...cat.children.array().forEach(...)`?

c650: something like that but change forEach to `map` and then you can reduce the categories to the results of the channel overrides. I thought that channel overrides were async so they'd have to be awaited?

then you can do

```js
promises = ...reduce();
await Promise.all(promises);

```

src/app/plugins/shadowban.plugin.ts
```diff
@@ -64,20 +71,25 @@ export class ShadowBanPlugin extends Plugin {
...
+    await Promise.all(promises);
   }
 
-  private _banUser(chan: GuildChannel, user: User) {
-    chan.overwritePermissions(user, {
-      VIEW_CHANNEL: false,
-    });
+  private _banUser(user: User) {
+    return (chan: GuildChannel) => {
+      chan.overwritePermissions(user, {
```
c650, line 77: `overwritePermissions` returns a promise. You should await it.

```
return async (chan) => {
  await chan.overwritePermissions(...);
}
```

https://discord.js.org/#/docs/main/stable/class/GuildChannel?scrollTo=overwritePermissions

src/app/plugins/shadowban.plugin.ts
```diff
@@ -64,20 +71,25 @@ export class ShadowBanPlugin extends Plugin {
...
+        VIEW_CHANNEL: false,
+      });
+    };
   }
 
-  private _unbanUser(chan: GuildChannel, user: User) {
-    chan.permissionOverwrites.get(user.id)?.delete();
+  private _unbanUser(user: User) {
+    return (chan: GuildChannel) => {
```
c650, line 84: saa

src/app/plugins/shadowban.plugin.ts
tanndlin: Still slightly figuring out promises lol, thought that `await Promise.all` was enough

c650 (commented):
now await the `applyToChannels` in `execute` and you're good to go.

c650 (approved):

PR #242, opened 2020-12-03
Send welcome message for new users

### LINKED ISSUE #238: DM user on join
https://github.com/joey-colon/lion/blob/b4b28e8790bf9c62b1c0f6d8b133450634a75dd3/src/app/handlers/new_member.handler.ts#L12-L14
Should DM new users to let them know that they have to acknowledge the CoC. Maybe that will solve the problem of people leaving an joining over and over

tanndlin:
fixes #238
Posting what I looks like in `lion_project`

### DIFF
--- a/src/app/handlers/new_member_role.handler.ts
+++ b/src/app/handlers/new_member_role.handler.ts
@@ -9,7 +9,7 @@ export class NewMemberRoleHandler implements IHandler {
     [this._UNACKNOWLEDGED_ROLE]: undefined,
     [this._UNVERIFIED_ROLE]: undefined,
   };
-  private _VERIFY_CHANNEL: Maybe<TextChannel> = null;
+  private _VERIFY_CHANNEL: Maybe<TextChannel> = undefined;
 
   constructor(public container: IContainer) {}
 
@@ -35,7 +35,7 @@ export class NewMemberRoleHandler implements IHandler {
       return;
     }
 
-    member.addRole(<Role>this._roleCache[this._UNVERIFIED_ROLE]).then(() => {
+    await member.addRole(<Role>this._roleCache[this._UNVERIFIED_ROLE]).then(() => {
       this._pingUserInVerify(member);
       this.container.messageService.sendBotReport(
         `User \`${member.user.tag}\` has been automatically unverified`
@@ -52,7 +52,7 @@ export class NewMemberRoleHandler implements IHandler {
     }
 
     this._VERIFY_CHANNEL.send(member.user.toString()).then((sentMsg) => {
-      (sentMsg as Message).delete(1000 * 10); //Delete after 10 seconds
+      (sentMsg as Message).delete(1000 * 60 * 60 * 12); //Delete after 12 hours
     });
   }
 }
--- a/src/app/handlers/new_member_role.handler.ts
+++ b/src/app/handlers/new_member_role.handler.ts
@@ -1,5 +1,6 @@
-import { GuildMember, Role } from 'discord.js';
+import { GuildMember, Role, TextChannel, GuildChannel, Message } from 'discord.js';
 import { IContainer, IHandler, Maybe } from '../../common/types';
+import Constants from '../../common/constants';
 
 export class NewMemberRoleHandler implements IHandler {
   private _UNACKNOWLEDGED_ROLE: string = 'Un Acknowledged';
@@ -8,6 +9,7 @@ export class NewMemberRoleHandler implements IHandler {
     [this._UNACKNOWLEDGED_ROLE]: undefined,
     [this._UNVERIFIED_ROLE]: undefined,
   };
+  private _VERIFY_CHANNEL: Maybe<TextChannel> = null;
 
   constructor(public container: IContainer) {}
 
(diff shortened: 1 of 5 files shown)


c650 (approved):
Lgtm. If we add any more stuff to this we may want to separate the different parts (adding roles, sending messages, ...) into different handlers (?) triggered by the event

tanndlin:
> Lgtm. If we add any more stuff to this we may want to separate the different parts (adding roles, sending messages, ...) into different handlers (?) triggered by the event

I'd rather do that now than later. I had issues with adding both roles from separate handlers, but if they are in the same one, and the welcome message is in a different handler, then it may work. I'll work on that in a bit

src/app/handlers/welcome.handler.ts
```diff
@@ -0,0 +1,52 @@
...
+export class WelcomeHandler implements IHandler {
+  private _LION_URL: string = 'https://github.com/joey-colon/lion';
+
+  constructor(public container: IContainer) {}
+
+  public async execute(member: GuildMember): Promise<void> {
+    const hasAvatar = Boolean(member.user.avatar);
+    const embed = this._createEmbed(hasAvatar);
+    try {
+      member.send(embed);
```
c650, line 14: if this is async you have to .catch and await

c650 (approved):

src/app/handlers/welcome.handler.ts
```diff
@@ -0,0 +1,50 @@
...
+    embed.addField(
+      `Read our Code of Conduct`,
+      `Please read our \`#${Constants.Channels.Public.CodeOfConduct}\` and react to the message to show you\
+      acknowledge our guidelines, and to gain access to the whole server.`,
+      true
+    );
+
+    if (!isVerified) {
+      embed.addField(
+        'You have been unverified in our server',
```
joey-colon, line 43: `You have been marked as unverified in our server`

joey-colon (approved):

src/app/handlers/welcome.handler.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export class WelcomeHandler implements IHandler {
+  private _LION_URL: string = 'https://github.com/joey-colon/lion';
+
+  constructor(public container: IContainer) {}
+
+  public async execute(member: GuildMember): Promise<void> {
+    const hasAvatar = Boolean(member.user.avatar);
+    const embed = this._createEmbed(hasAvatar);
+    await member.send(embed).catch();
```
joey-colon, line 13: if we hit the catch it means they have dms disabled. we should probably send it to a channel and maybe ping them (?)

tanndlin: We would prob need a new channel with the message in it. Overwrite their perms for it if we hit the catch, ping. and take away overwrite after they acknowledge

c650: Might as well just have another channel only Unacknowledged can see

tanndlin: What if they have/haven't been marked as unverified? You probably don't want that in there since not *everyone* is unverified

c650: maybe if they're unverified we should @ them in the verify channel.

c650: this would re-couple the welcome message with the roles, or at least require synchronisation.

tanndlin: not necessarily, all new users are unacknowledged.
And we can check to see if they have a pfp to see if they will be unverified

tanndlin: Also, ping in verify only if DMs are off?

src/app/handlers/new_member_role.handler.ts
```diff
@@ -33,9 +35,24 @@ export class NewMemberRoleHandler implements IHandler {
...
+  private _pingUserInVerify(member: GuildMember) {
+    if (!this._VERIFY_CHANNEL) {
+      this._VERIFY_CHANNEL = this.container.guildService
+        .get()
+        .channels.filter((chan: GuildChannel) => chan.name === Constants.Channels.Bot.Verify)
+        .first() as TextChannel;
+    }
+
+    this._VERIFY_CHANNEL.send(member.user.toString()).then((sentMsg) => {
+      (sentMsg as Message).delete(1000 * 10); //Delete after 10 seconds
```
c650, line 55: might not be long enough; make it 12h

src/app/handlers/welcome.handler.ts
```diff
@@ -1,16 +1,36 @@
...
+        this._UNACKNOWLEDGED_CHANNEL = this.container.guildService
+          .get()
+          .channels.filter((chan) => chan.name === Constants.Channels.Bot.Unacknowledged)
+          .first() as TextChannel;
+      }
+
+      //Check if the default embed is in the channel
+      this._UNACKNOWLEDGED_CHANNEL.fetchMessages({ limit: 100 }).then((messages) => {
+        if (!messages.size) {
+          (this._UNACKNOWLEDGED_CHANNEL as TextChannel).send(this._createEmbed(true));
```
c650, line 25: return ?

src/app/handlers/welcome.handler.ts
```diff
@@ -1,16 +1,36 @@
...
+        }
+
+        //Do this inside the then, to make sure it fetchMessages doesn't fetch the mention
+        (this._UNACKNOWLEDGED_CHANNEL as TextChannel)
```
c650, line 29: return

c650 (commented):
careful not to leave promises dangling

src/app/handlers/new_member_role.handler.ts
tanndlin: It doesn't block, but depending on how many people join at once, there could be a good few pings at once in the channel. Is that ok?

c650 (approved):

joey-colon (approved):

PR #244, opened 2020-12-04
Send Market listings in #buy_sell_trade

### LINKED ISSUE #243: Keep one !market list call inside of #buy_sell_trade
Currently Lion sends the user a DM of all the listings, and sends it in #buy_sell_trade if their DMs are closed.
Some people don't want to have to leave the channel to see the listings

A few solutions:

1. Send an updated list to #buy_sell_trade on each `!market add` call, and delete the last list posted.
2. Pin a single list, and just edit that.
3. Send the list on each `!market list` call and delete the old one if something has changed. Don't do anything if there were no changes since last `!market list` call. (Possibly send a link to the last list if nothing changed, which we could queue for deletion after a few seconds as to not spam the channel).

On top of all this, maybe add an optional argument for sending the list as DM regardless, if the user prefers.

tanndlin:
fixes #243

### DIFF
--- a/src/app/plugins/marketplace.plugin.ts
+++ b/src/app/plugins/marketplace.plugin.ts
@@ -50,7 +50,7 @@ export class MarketPlacePlugin extends Plugin {
   }
 
   private async _handleListMarket(message: IMessage, shouldDmUser: boolean) {
-    const oldMessages = await this._fetchMesages(message, 300);
+    const oldMessages = await this._fetchMessages(message, 300);
     const itemsForSale = await this._fetchListings(oldMessages);
     this._deleteOldListingPosts(oldMessages);
 
@@ -82,7 +82,7 @@ export class MarketPlacePlugin extends Plugin {
     }
   }
 
-  private async _fetchMesages(message: IMessage, limitParam: number) {
+  private async _fetchMessages(message: IMessage, limitParam: number) {
     let i: number;
     let last_id = message.id;
     const buffer: Message[] = [];
--- a/src/app/plugins/marketplace.plugin.ts
+++ b/src/app/plugins/marketplace.plugin.ts
@@ -1,6 +1,6 @@
 import Constants from '../../common/constants';
 import { Plugin } from '../../common/plugin';
-import { IContainer, IMessage, ChannelType } from '../../common/types';
+import { IContainer, IMessage, ChannelType, Maybe } from '../../common/types';
 import { RichEmbed, Message } from 'discord.js';
 
 export class MarketPlacePlugin extends Plugin {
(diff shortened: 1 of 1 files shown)


src/app/plugins/marketplace.plugin.ts
```diff
@@ -34,7 +34,7 @@ export class MarketPlacePlugin extends Plugin {
     if (sub_command === 'add') {
       this._handleAddMarket(message, description);
     } else if (sub_command === 'list') {
-      this._handleListMarket(message);
+      this._handleListMarket(message, description?.toLowerCase() === 'dm');
```
c650, line 37: nit: `description` not the best name here

src/app/plugins/marketplace.plugin.ts
```diff
@@ -49,23 +49,31 @@ export class MarketPlacePlugin extends Plugin {
...
-    this._replyToUser(message, embed);
+
+    message.channel.send(embed);
+    if (shouldDmUser) {
+      this._replyToUser(message, embed);
+    }
     message.delete();
   }
 
+  private _deleteOldListingPosts(messages: Message[]) {
```
c650, line 69: consider that only one listing post "should" exist above (any others can just be buried, it's fine)

why not keep `Message` object for the last non-DM Listings Message, then when you make a new one, delete the old one, then save the new one as the new "last" 

this way we dont fetch and loop, which is slow.

tanndlin: True, I'm conditioned to try to commit as little to memory as possible so I didn't even think of that 😄

src/app/plugins/marketplace.plugin.ts
```diff
@@ -11,9 +11,9 @@ export class MarketPlacePlugin extends Plugin {
   public permission: ChannelType = ChannelType.Public;
   public pluginChannelName: string = Constants.Channels.Public.BuySellTrade;
   private _NEW_LISTING_PREFIX = 'marketplace add';
-  private _GET_LISTING_PREFIX = 'marketplace list';
   private _NEW_ALIAS_PREFIX = 'market add';
   private _TARGET_REACTION = '💰';
+  private _LAST_LISTING_POST: Maybe<IMessage> = undefined;
```
c650, line 16: since this isn't a class constant it should be lowercase

src/app/plugins/marketplace.plugin.ts
```diff
@@ -49,21 +49,45 @@ export class MarketPlacePlugin extends Plugin {
...
+      [this._LAST_LISTING_POST] = messages.filter(
+        (msg) => msg.author.bot && msg.id != newPosting.id
+      );
+    });
+
+    //It's possible to have not posted a list in the last 100 messages
+    if (!this._LAST_LISTING_POST) {
+      this._LAST_LISTING_POST = newPosting;
+      return;
+    }
```
c650, line 88: nit: blank after }

c650 (approved):

c650 (approved):

c650:
approved; can we merge?

c650 (approved):

PR #255, opened 2020-12-21
Only allow mods to register all

### LINKED ISSUE #235: Only allow the Moderator role to !register all
https://github.com/joey-colon/lion/blob/b4b28e8790bf9c62b1c0f6d8b133450634a75dd3/src/app/plugins/register.plugin.ts#L26-L28
Should make sure the user has the Moderator role, if the arg is `all`.

tanndlin:
fixes #235

### DIFF
--- a/src/app/plugins/register.plugin.ts
+++ b/src/app/plugins/register.plugin.ts
@@ -28,7 +28,7 @@ export class RegisterPlugin extends Plugin {
         const isModerator = this.container.guildService.userHasRole(message.author, 'Moderator');
         if (!isModerator) {
           message.reply('You must be a `Moderator` to register for `all`.');
-          continue;
+          return;
         }
       }
 
--- a/src/app/plugins/register.plugin.ts
+++ b/src/app/plugins/register.plugin.ts
@@ -24,6 +24,19 @@ export class RegisterPlugin extends Plugin {
     const successfulClasses: string[] = [];
     const invalidClasses: string[] = [];
     for (const arg of args) {
+      if (arg === 'all') {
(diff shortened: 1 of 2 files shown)


c650 (approved):

src/app/plugins/register.plugin.ts
```diff
@@ -24,6 +25,19 @@ export class RegisterPlugin extends Plugin {
     const successfulClasses: string[] = [];
     const invalidClasses: string[] = [];
     for (const arg of args) {
+      if (arg === 'all') {
+        const member = this.container.guildService.get().members.get(message.author.id);
+        if (!member) {
+          continue;
+        }
+
+        const isModerator = MemberUtils.hasRole(member, 'Moderator');
```
c650, line 34: maybe only expose hasRole through guildService

```js
const isModerator = this.container.guildService.memberHasRole(message.author, '...');
```

... or some other way such that we don't have a repeated "member lookup" phase.

c650 (approved):

src/app/plugins/register.plugin.ts
```diff
@@ -24,6 +24,14 @@ export class RegisterPlugin extends Plugin {
     const successfulClasses: string[] = [];
     const invalidClasses: string[] = [];
     for (const arg of args) {
+      if (arg === 'all') {
+        const isModerator = this.container.guildService.userHasRole(message.author, 'Moderator');
+        if (!isModerator) {
+          message.reply('You must be a `Moderator` to register for `all`.');
+          continue;
```
c650, line 31: maybe a return

c650 (approved):

PR #264, opened 2021-01-06
Market List Exceeding Char Limit Fix

### LINKED ISSUE #263: Market List exceeding 2048 char limit in embed
image

Should be as simple as checking the length of the listings string and splitting into multiple embeds

tanndlin:
fixes #263

### DIFF
--- a/src/app/plugins/marketplace.plugin.ts
+++ b/src/app/plugins/marketplace.plugin.ts
@@ -43,7 +43,7 @@ export class MarketPlacePlugin extends Plugin {
 
   private _handleAddMarket(message: IMessage, description: string) {
     const stringForUser = description
-      ? 'Your item has been added! Please delete your message once it is sold.'
+      ? `Your item has been added! Please react to your message with ${this._TARGET_REACTION} once it is sold.`
       : 'Invalid Listing.';
     const embed = new RichEmbed();
     embed.setDescription(stringForUser);
--- a/src/app/plugins/marketplace.plugin.ts
+++ b/src/app/plugins/marketplace.plugin.ts
@@ -54,24 +54,21 @@ export class MarketPlacePlugin extends Plugin {
     const oldMessages = await this._fetchMessages(message, 300);
     const itemsForSale = await this._fetchListings(oldMessages);
 
-    let curLength = 0;
-    let curItems: string[] = [];
-    const embeds = [];
-    itemsForSale.forEach((item) => {
-      if (curLength + item.length > 2000) {
-        embeds.push(this._createListingEmbed(curItems));
-        curItems = [item];
-        curLength = item.length;
-        return;
+    const chunks = [];
+    while (itemsForSale.length > 0) {
+      let curLength = 0;
+      const temp = [];
+
+      while (itemsForSale.length && curLength + itemsForSale.slice(-1)[0].length < 2000) {
+        temp.push(itemsForSale.pop() as string);
+        curLength = temp.join('').length;
       }
 
-      curLength += item.length;
-      curItems.push(item);
-    });
-    //Push rest of items
-    embeds.push(this._createListingEmbed(curItems));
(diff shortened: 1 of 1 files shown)


tanndlin:
What should I do for the case where someone creates a listing that is greater than 2000 characters? That would stick the code in an infinite loop.
Should we just not allow it? or just filter out messages that are too big

c650:
> What should I do for the case where someone creates a listing that is greater than 2000 characters? That would stick the code in an infinite loop. Should we just not allow it and delete that listing? or just filter out messages that are too big

If you decide to go with a length limit, enforce it when someone tries to add a listing. However, I would think that if the listing was too large you would not even be able to send the message to post the listing.

tanndlin:
> > What should I do for the case where someone creates a listing that is greater than 2000 characters? That would stick the code in an infinite loop. Should we just not allow it and delete that listing? or just filter out messages that are too big
> 
> If you decide to go with a length limit, enforce it when someone tries to add a listing. However, I think that if the listing was too large you would not even be able to send the message to post the listing.

Ohhh, true. There is a 2k character limit cap when sending a message. In that case, we're all good

src/app/plugins/marketplace.plugin.ts
```diff
@@ -14,7 +14,7 @@ export class MarketPlacePlugin extends Plugin {
   private _LISTING_PREFIX = '!marketplace add';
   private _ALIAS_PREFIX = '!market add';
   private _TARGET_REACTION = '💰';
-  private _lastListingPost: Maybe<IMessage> = undefined;
+  private _lastListingPost: Maybe<IMessage[]> = undefined;
```
c650, line 17: nit: since you use an array now you could just use empty array as your 'undefined' value

src/app/plugins/marketplace.plugin.ts
```diff
@@ -35,49 +35,67 @@ export class MarketPlacePlugin extends Plugin {
...
+    while (itemsForSale.length > 0) {
+      let curLength = 0;
+      const temp = [];
 
-    await message.channel
-      .send(embed)
-      .then(async (sentMsg) => await this._deleteOldListingPost(message, sentMsg as Message));
-    if (shouldDmUser) {
-      this._replyToUser(message, embed);
+      while (itemsForSale.length && curLength + itemsForSale.slice(-1)[0].length < 2000) {
```
c650, line 62: 2000 is magic number

c650: make sure `slice` isn't too slow for a loop

src/app/plugins/marketplace.plugin.ts
```diff
@@ -35,49 +35,67 @@ export class MarketPlacePlugin extends Plugin {
...
+    embed.setDescription(items.reverse().join('\n\n'));
+    return embed;
+  }
+
+  private async _deleteOldListingPost(listCall: IMessage, newPosting: IMessage[]) {
     //.get To make sure the message wasnt deleted already
-    if (this._lastListingPost && listCall.channel.messages.get(this._lastListingPost.id)) {
-      await this._lastListingPost.delete().catch();
+    if (this._lastListingPost && listCall.channel.messages.get(this._lastListingPost[0].id)) {
+      await this._lastListingPost.forEach((msg) => msg.delete().catch());
```
c650, line 89: this might have to be the `await Promise.all(... map() )` flow

src/app/plugins/marketplace.plugin.ts
```diff
@@ -86,15 +104,15 @@ export class MarketPlacePlugin extends Plugin {
       return;
     }
 
-    this._lastListingPost.delete();
+    await this._lastListingPost.forEach((msg) => msg.delete());
```
c650, line 107: saa

c650 (approved):

c650:
send it?

PR #267, opened 2021-01-07
prevent !register ALL from working

Kurtoid:


joey-colon:
thanks for reporting!

i believe we should transform args to lowercase before running the plugin which is located [here](https://github.com/joey-colon/lion/blob/master/src/app/handlers/command.handler.ts#L50)

Kurtoid:
Would converting to lowercase there break anything else?

c650:
yes

c650:
There are a handful of plugins that take a username as one of the arguments, which is case sensitive

c650 (approved):
lgtm

Issue #268, opened 2021-01-08
Class Check Plugin


tanndlin:
Make a plugin to see what classes a user is registered for
Probably make command work in user_offenses

c650:
probably bot_logs

c650:
Be mindful of the class list being too large for one message :flushed:

tanndlin:
https://github.com/joey-colon/lion/blob/c9cb4206739b57c7f522aba9d31aba6b4b40e49d/src/app/plugins/checkclasses.plugin.ts#L44-L46
I'll add a check for that aside from this

tanndlin:
fixed by #270

PR #270, opened 2021-01-08
Check Classes For User Plugin

tanndlin:
image

### DIFF
--- a/src/app/plugins/checkclasses.plugin.ts
+++ b/src/app/plugins/checkclasses.plugin.ts
@@ -1,4 +1,4 @@
-import { GuildChannel } from 'discord.js';
+import { Guild, GuildChannel } from 'discord.js';
 import { Plugin } from '../../common/plugin';
 import { IContainer, IMessage, ChannelType } from '../../common/types';
 
@@ -9,7 +9,8 @@ export class CheckClassesPlugin extends Plugin {
   public pluginAlias = [];
   public permission: ChannelType = ChannelType.Staff;
 
-  //   private _allClassChannels: GuildChannel[] = [];
+  private _MAX_CHAR_LIMIT: number = 2000;
+  private _ALLOW_NUM: number = 3072; //ID for allowed permission overwrite
 
   constructor(public container: IContainer) {
     super();
@@ -32,8 +33,8 @@ export class CheckClassesPlugin extends Plugin {
     }
 
     const allClassChannels = this._getAllClassChannels();
-    const chansContainingUser = allClassChannels.filter((chan) =>
-      chan.permissionOverwrites.get(member.id)
+    const chansContainingUser = allClassChannels.filter(
+      (chan) => chan.permissionOverwrites.get(member.id)?.allow === this._ALLOW_NUM
     );
 
     if (chansContainingUser.length === 0) {
@@ -46,10 +47,10 @@ export class CheckClassesPlugin extends Plugin {
       return;
     }
 
-    message.reply(
-      `User is registered for:\n\`${chansContainingUser.map((c) => c.name).join(' | ')}\``
-    );
+    const chansForMessage: string[] = this._convertChansToString(chansContainingUser);
+    chansForMessage.forEach((m) => message.reply(`User is registered for:\n\`${m}\``));
   }
+
   private _getAllClassChannels(): GuildChannel[] {
     const guildChans = this.container.guildService.get().channels;
 
@@ -58,5 +59,16 @@ export class CheckClassesPlugin extends Plugin {
       .array();
 
     return classChans;
+  }
+
+  private _convertChansToString(userChans: GuildChannel[]): string[] {
+    const toString = userChans.map((c) => c.name).join(' | ');
+    if (toString.length < this._MAX_CHAR_LIMIT) {
+      return [toString];
+    }
+
+    const middle = userChans.length / 2;
+    const splitChans = [userChans.slice(0, middle), userChans.slice(middle)];
+    return splitChans.map((chans) => chans.map((chan) => chan.name).join(' | '));
   }
 }
--- a/src/app/plugins/checkclasses.plugin.ts
+++ b/src/app/plugins/checkclasses.plugin.ts
@@ -50,7 +50,9 @@ export class CheckClassesPlugin extends Plugin {
     }
 
     const chansForMessage: string[] = this._convertChansToString(chansContainingUser);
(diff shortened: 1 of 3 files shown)


tanndlin:
fixes #268

c650:
Perhaps include the number of classes in the message so we dont have to count

src/app/plugins/checkclasses.plugin.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (chansContainingUser.length === allClassChannels.length) {
+      message.reply('User is registered for all classes.');
+      return;
+    }
+
+    const chansForMessage: string[] = this._convertChansToString(chansContainingUser);
+    chansForMessage.forEach((m) => message.reply(`User is registered for:\n\`${m}\``));
+  }
+
+  private _getAllClassChannels(): GuildChannel[] {
```
c650, line 58: is there not a way to get all class channels from the classService?

src/app/plugins/checkclasses.plugin.ts
```diff
@@ -0,0 +1,74 @@
...
+    const toString = userChans.map((c) => c.name).join(' | ');
+    if (toString.length < this._MAX_CHAR_LIMIT) {
+      return [toString];
+    }
+
+    const middle = userChans.length / 2;
+    const splitChans = [userChans.slice(0, middle), userChans.slice(middle)];
+    return splitChans.map((chans) => chans.map((chan) => chan.name).join(' | '));
+  }
+}
```
c650, line 78: so if it doesnt fit in one message you just put it in two ?

c650: this method is confusinggg

src/app/plugins/checkclasses.plugin.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  private _MAX_CHAR_LIMIT: number = 2000;
+  private _ALLOW_NUM: number = 3072; //ID for allowed permission overwrite
+
+  constructor(public container: IContainer) {
+    super();
+  }
+
+  public validate(message: IMessage, args: string[]) {
+    return args && args.length >= 1;
```
c650, line 22: `args &&` not necessary

c650: deepscan

src/app/plugins/checkclasses.plugin.ts
```diff
@@ -0,0 +1,74 @@
+import { Guild, GuildChannel } from 'discord.js';
+import { Plugin } from '../../common/plugin';
+import { IContainer, IMessage, ChannelType } from '../../common/types';
+
+export class CheckClassesPlugin extends Plugin {
+  public name: string = 'Check Class';
+  public description: string = 'lists the classes someone is in';
+  public usage: string = 'checkclass <user>';
+  public pluginAlias = [];
+  public permission: ChannelType = ChannelType.Staff;
```
c650, line 11: only allow in user_offenses

tanndlin:
image

src/app/plugins/checkclasses.plugin.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  private _MAX_CHAR_LIMIT: number = 2000;
+  private _ALLOW_NUM: number = 3072; //ID for allowed permission overwrite
```
c650, line 15: is there a programmatic way to get this number from discord.js so it isnt so magical?

c650 (approved):

src/app/plugins/checkclasses.plugin.ts
tanndlin: The Deny ID is just 0, should I change it to deny ID and check the opposite?

c650: I am not sure that would work because:

a class channel is by default `DENY @. everyone` so there is no specific permission override to DENY that user (unless they *unregistered* but that's not important).

it's fine the way it is.. just a nitpick really :)

src/app/plugins/checkclasses.plugin.ts
```diff
@@ -0,0 +1,75 @@
...
+      return;
+    }
+
+    const chansForMessage: string[] = this._convertChansToString(chansContainingUser);
+    chansForMessage.forEach((m) =>
+      message.reply(`User is registered for \`${chansContainingUser.length}\` classes:\n\`${m}\``)
+    );
+  }
+
+  private _convertChansToString(userChans: GuildChannel[]): string[] {
```
c650, line 62: nit: map channel to channel.name only once

c650 (approved):

src/app/plugins/checkclasses.plugin.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    const classes = this.container.classService.getClasses(ClassType.ALL);
+    const chansContainingUser = [];
+
+    for (const classObj of classes) {
+      const [, chan] = classObj;
+      if (chan.permissionOverwrites.get(member.id)?.allow) {
+        chansContainingUser.push(chan);
+      }
+    }
```
joey-colon, line 44: nit this could be a Array.filter()

src/app/plugins/checkclasses.plugin.ts
```diff
@@ -0,0 +1,77 @@
...
+      return;
+    }
+
+    const chansForMessage: string[] = this._convertChansToString(chansContainingUser);
+    chansForMessage.forEach((m) =>
+      message.reply(`User is registered for \`${chansContainingUser.length}\` classes:\n\`${m}\``)
+    );
+  }
+
+  private _convertChansToString(userChans: GuildChannel[]): string[] {
```
joey-colon, line 62: this function name doesnt completely tell me what the function does. `_adaptToChannelMessage` would make more sense as you're building a chan message that's able to be sent to discord.

joey-colon (approved):

src/app/plugins/checkclasses.plugin.ts
tanndlin: I don't believe you can filter a Map<string, GuildChannel>.

joey-colon: ah right i think its fine then. if we really wanted to, we could take advantage of Map.values() and do something like

`const chansContainingUser = Array.from(classes.values()).filter((class) => Boolean(class.permissionOverwrites.get(member.id)?.allow));`

but this requires 2 passes and we dont wanna upset @c650 😞

PR #278, opened 2021-01-11
DMed ListRoles Response

### LINKED ISSUE #277: Send !listroles response in DM

tanndlin:
fixes #277

### DIFF
--- a/src/app/plugins/listroles.plugin.ts
+++ b/src/app/plugins/listroles.plugin.ts
@@ -32,6 +32,7 @@ export class ListRolesPlugin extends Plugin {
         res += `${role.name.toLowerCase()}\n`;
       });
     res += '```';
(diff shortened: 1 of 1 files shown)


src/app/plugins/listroles.plugin.ts
```diff
@@ -32,6 +32,7 @@ export class ListRolesPlugin extends Plugin {
         res += `${role.name.toLowerCase()}\n`;
       });
     res += '```';
-    message.reply(res);
+    // message.reply(res);
+    message.author.send(res).catch(() => message.reply(res));
```
c650, line 35: probably have to await so teh `catch` actually happens

c650 (approved):
afraid to deploy rn but lgtm

PR #291, opened 2021-01-17
Universal way of DMing Users

### LINKED ISSUE #285: Universal way of DMing users result of command
We need to have a universal way of attempting to DM users, and once its completed, react to the message to show that it worked for others in the channel.

tanndlin:
Made a method in `messageService` to attempt to DM users, and send the message in the channel if it fails.
Also reacts with '👍' if the DM was successful so others can see that it worked

fixes #285

### DIFF
--- a/src/services/message.service.ts
+++ b/src/services/message.service.ts
@@ -1,14 +1,15 @@
-import { IMessage, IContainer } from '../common/types';
-import { GuildChannel, Guild, TextChannel, RichEmbed, User } from 'discord.js';
+import { IMessage } from '../common/types';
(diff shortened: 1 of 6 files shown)


src/services/message.service.ts
```diff
@@ -30,6 +30,15 @@ export class MessageService {
     this._sendConstructedReport(report, { files: message.attachments.map((e) => e.url) });
   }
 
+  async attempDMUser(message: IMessage, content: string | RichEmbed) {
+    try {
+      await message.author.send(content);
+      await message.react('👍');
```
c650, line 36: Nit use .then combinator

c650 (approved):

c650 (approved):

PR #306, opened 2021-02-02
Send Market listings as a paged embed

tanndlin:


tanndlin:
image

src/services/message.service.ts
```diff
@@ -39,6 +39,34 @@ export class MessageService {
...
+      (reaction: MessageReaction, user: User) =>
+        ['⬅️', '➡️'].includes(reaction.emoji.name) && user.id !== msg.author.id, //Only run if its not the bot putting reacts
+      { time: 1000 * 60 * 10 } //Listen for 10 Minutes
+    );
+
+    let pageIndex = 0;
+    collecter.on('collect', (reaction: MessageReaction) => {
+      msg.reactions.removeAll().then(async () => {
+        reaction.emoji.name === '➡️' ? pageIndex++ : pageIndex--;
+        msg.edit(pages[pageIndex]);
```
c650, line 56: can be out of bounds

tanndlin: It only sends the arrows if its [available.](https://github.com/joey-colon/lion/blob/65a5836da2702a98008c3c1c051ba068bbd1dd82/src/services/message.service.ts#L58-L63)
So the user cannot go past the bounds

c650: nice sneaky invariant. what happens if two people react at the same time though?

c650: in a way such that you could increment past the pages' bounds

c650 (approved):

PR #308, opened 2021-02-06
CreateClassVoice Plugin

### LINKED ISSUE #307: Dynamic Audio Channel Creation
I would like to have users have the ability to make temporary voice chats for their classrooms by calling a command. Something like !classvoicechat in their class channel and it would make #classcode_professor_voice in a Class Voice Chat category.

Requirements:
1. Create a voice channel in a newly created Class Voice Chat category from the class channel.
2. Default to private channel and mirror permissions of the class channel into the audio channel. Only those that can see the audio channel are the ones that are currently registered for the class through bot_commands. 
3. Put the calling user into the channel when the call is made. 
4. Remove the channel after a set time of inactivity. I think something like when the last person leaves the channel, 5 minutes later the channel is deleted. It can be shorter, 1 minute even. I just want to make sure they don't clog up the server with people making random channels and walking away from them.
5. I don't know if it will ever come up but if more than one student wants to make a voice channel it can be a progressive numbering scheme. #classcode_professor_voice1, #classcode_professor_voice2, #classcode_professor_voice3 etc. 

Feel free to name the command something cooler.

tanndlin:
fixes #307

### DIFF
--- a/src/app/plugins/createclassvoice.plugin.ts
+++ b/src/app/plugins/createclassvoice.plugin.ts
@@ -6,7 +6,7 @@ export class CreateClassVoice extends Plugin {
   public name: string = 'Create Class Voice';
   public description: string = 'Creates a temporary voice channel for a class';
   public usage: string = 'createclassvoice';
-  public pluginAlias = [];
+  public pluginAlias = ['createvc'];
   public permission: ChannelType = ChannelType.Private;
 
   constructor(public container: IContainer) {
--- a/src/app/plugins/createclassvoice.plugin.ts
+++ b/src/app/plugins/createclassvoice.plugin.ts
@@ -1,0 +1,60 @@
+import { Plugin } from '../../common/plugin';
+import { IContainer, IMessage, ChannelType, Maybe } from '../../common/types';
+import { CategoryChannel, TextChannel } from 'discord.js';
+
+export class CreateClassVoice extends Plugin {
+  public name: string = 'Create Class Voice';
+  public description: string = 'Creates a temporary voice channel for a class';
+  public usage: string = 'createclassvoice';
+  public pluginAlias = [];
+  public permission: ChannelType = ChannelType.Private;
+
+  private _INACTIVE_THRESHOLD: number = 10;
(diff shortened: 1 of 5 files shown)


joey-colon:
it makes more sense to have this as a job imo

src/app/plugins/createclassvoice.plugin.ts
```diff
@@ -0,0 +1,60 @@
...
+import { CategoryChannel, TextChannel } from 'discord.js';
+
+export class CreateClassVoice extends Plugin {
+  public name: string = 'Create Class Voice';
+  public description: string = 'Creates a temporary voice channel for a class';
+  public usage: string = 'createclassvoice';
+  public pluginAlias = [];
+  public permission: ChannelType = ChannelType.Private;
+
+  private _INACTIVE_THRESHOLD: number = 10;
```
c650, line 12: 10 what? minutes?

src/app/plugins/createclassvoice.plugin.ts
```diff
@@ -0,0 +1,60 @@
...
+  private _AUDIO_CAT: Maybe<CategoryChannel> = null;
+
+  constructor(public container: IContainer) {
+    super();
+  }
+
+  public validate(message: IMessage, args: string[]) {
+    return args.length == 0;
```
c650, line 20: ===

src/app/plugins/createclassvoice.plugin.ts
```diff
@@ -0,0 +1,60 @@
...
+      parent: this._AUDIO_CAT,
+      permissionOverwrites: chan.permissionOverwrites,
+    });
+
+    await voiceChan.createInvite().then(async (invite) => {
+      await message.channel.send(`Channel Created: ${invite}`);
+    });
+
+    let numInChan = voiceChan.members.size;
+    const deleteInterval = setInterval(async () => {
```
c650, line 43: use job for this?

c650 (commented):
perhaps use a Service to manage teh creation/deletion of these channels and a Job to check for inactive and deallocate.

c650 (approved):

PR #312, opened 2021-02-08
Filter out sold listings in marketplace

tanndlin:


src/app/plugins/marketplace.plugin.ts
```diff
@@ -144,8 +144,10 @@ export class MarketPlacePlugin extends Plugin {
   private async _fetchListings(messages: Message[]): Promise<string[]> {
     const calls = messages.filter(
       (msg) =>
-        msg.content.startsWith(this._LISTING_PREFIX) || msg.content.startsWith(this._ALIAS_PREFIX)
-    ); //Filter out non !market adds
+        (msg.content.startsWith(this._LISTING_PREFIX) || //Filter out non !market adds
+          msg.content.startsWith(this._ALIAS_PREFIX)) &&
+        !Boolean(msg.reactions.cache.find((r) => r.emoji.name === this._TARGET_REACTION)) //Filter out sold listings
+    );
```
c650, line 150: order of operations. && has higher precedence. you are thinking its (a || b) && c but as written it is a || (b && c)

c650: ah drats i missed the parens there
all good

c650 (approved):

PR #329, opened 2021-03-24
Better channel redirect

CallumCoy:
Gives channel redirect links rather than just some text when a user uses a command that is invalid for that room.  Also give examples in the cases of Permission types commands. With the exception of private channels since they are special in the case that we have none defined in constants, and that all channels default to that if undeclared.

### DIFF
--- a/src/common/plugin.ts
+++ b/src/common/plugin.ts
@@ -1,4 +1,6 @@
 import { ChannelType, IContainer, IMessage, IPlugin, RoleType } from './types';
+import Constants from '../common/constants';
+import { Container } from 'winston';
 
 export abstract class Plugin implements IPlugin {
   public abstract container: IContainer;
@@ -28,7 +30,15 @@ export abstract class Plugin implements IPlugin {
   public hasPermission(message: IMessage): boolean {
     const channelName = this.container.messageService.getChannel(message).name;
     if (typeof this.pluginChannelName === 'string' && this.pluginChannelName !== channelName) {
-      message.reply(`Please use this command in the \`#${this.pluginChannelName}\` channel.`);
+
+      //if this fails it simply means that a channel the bot expected to exist didn't, nothing bot breaking
+      try{
+        const id = this.container.guildService.getChannel(this.pluginChannelName).id
+        message.reply(`Please use this command in the <#${id}> channel.`);
+      }catch(err){
+        this.container.loggerService.warn(err)
+        message.reply(`Please use this command in the \`#${this.pluginChannelName}\` channel.`);
+      }
       return false;
     }
 
@@ -47,7 +57,29 @@ export abstract class Plugin implements IPlugin {
 
     const response = this.container.channelService.hasPermission(channelName, this.permission);
     if (!response) {
-      message.reply(`Please use this command in a \`${this.permission}\` channel`);
+      const rooms = Object.values(Constants.Channels[this.permission]);
+      let addonText = '';
+
+      //if this fails it simply means that a channel the bot expected to exist didn't
+      try{
+        if (rooms.length === 0) {
+          addonText = "their are no channels oh this type declared";
+        } else if (rooms.length === 1) {
+          let id = this.container.guildService.getChannel(rooms[0]).id
+          addonText = `, <#${id}> is the only channel whith this type`
+        } else{
+          let id = [this.container.guildService.getChannel(rooms[0]).id, this.container.guildService.getChannel(rooms[1]).id]
+          if (rooms.length == 2){
+            addonText = `, <#${id[0]}> and <#${id[1]}> are the only two with this channel type`
+          } else {
+            addonText = `, <#${id[0]}> and <#${id[1]}> are two of the channels this channel type`
+          }
+        }
+      } catch(err){
+        this.container.loggerService.warn(err)
+      } finally {
+        message.reply(`Please use this command in a \`${this.permission}\` channel${addonText}.`);
+      }
     }
     return response;
   }
--- a/src/common/plugin.ts
+++ b/src/common/plugin.ts
@@ -29,14 +29,8 @@ export abstract class Plugin implements IPlugin {
   public hasPermission(message: IMessage): boolean {
     const channelName = this.container.messageService.getChannel(message).name;
     if (typeof this.pluginChannelName === 'string' && this.pluginChannelName !== channelName) {
-      let channel = '';
-      try {
-        channel = `<#${this.container.guildService.getChannel(this.pluginChannelName).id}>`;
-      } catch (err) {
-        this._errorGen([this.pluginChannelName], err);
-        channel = `\`#${this.pluginChannelName}\``;
-      }
-      message.reply(`Please use this command in the ${channel} channel.`);
+      const id = this.container.guildService.getChannel(this.pluginChannelName).id;
+      message.reply(`Please use this command in the ${id} channel.`);
       return false;
     }
 
@@ -55,18 +49,18 @@ export abstract class Plugin implements IPlugin {
 
     const response = this.container.channelService.hasPermission(channelName, this.permission);
     if (!response) {
-      const rooms = Object.values(Constants.Channels[this.permission]);
-      const totalChannels = rooms.length;
-      rooms.splice(3);
+      const channels = Object.values(Constants.Channels[this.permission]);
+      const totalChannels = channels.length;
+      channels.splice(3);
 
       let addonText = '';
 
       try {
         // not sure if this will cause issues as I do pop id, please let me know
-        const id = rooms
-          .filter((room) => {
+        const id = channels
+          .filter((channel) => {
             return this.container.guildService
-              .getChannel(room)
+              .getChannel(channel)
               .permissionsFor(message.member || '')
               ?.has('VIEW_CHANNEL');
           })
@@ -74,17 +68,23 @@ export abstract class Plugin implements IPlugin {
 
         if (this.permission.toString() === 'Private') {
           addonText = ` This is primarily the class channels, and any channels we haven't defined.`;
-        } else if (id.length === 0) {
+        }
+
+        if (id.length === 0) {
           addonText = ' There are no permanent channels of this type.';
-        } else if (id.length === 1) {
+        }
+
+        if (id.length === 1) {
           addonText = ` <#${id[0]}> is the only channel whith this type.`;
-        } else {
+        }
+
+        if (id.length > 1) {
           addonText =
             `\nHere are ${id.length} of the ${totalChannels} supported channel(s): \n` +
             `${id.map((chan) => `<#${chan}>`).join(',\n')}.`;
         }
       } catch (err) {
-        this._errorGen(rooms, err);
+        this._errorGen(channels, err);
       } finally {
         message.reply(`Please use this command in a \`${this.permission}\` channel.${addonText}`);
       }
--- a/src/common/plugin.ts
+++ b/src/common/plugin.ts
@@ -18,6 +18,8 @@ export abstract class Plugin implements IPlugin {
 
   public pluginChannelName?: string;
 
+  private numChannelsShown: number = 6;
+
   // Typical defaults for existing commands.
   public usableInDM = false;
   public usableInGuild = true;
@@ -60,7 +62,7 @@ export abstract class Plugin implements IPlugin {
 
       const channels = Object.values(Constants.Channels[this.permission]);
       const totalChannels = channels.length;
-      channels.splice(10);
+      channels.splice(this.numChannelsShown);
 
       try {
         const id = channels
--- a/src/common/plugin.ts
+++ b/src/common/plugin.ts
@@ -30,7 +30,7 @@ export abstract class Plugin implements IPlugin {
     const channelName = this.container.messageService.getChannel(message).name;
     if (typeof this.pluginChannelName === 'string' && this.pluginChannelName !== channelName) {
       const id = this.container.guildService.getChannel(this.pluginChannelName).id;
-      message.reply(`Please use this command in the ${id} channel.`);
+      message.reply(`Please use this command in the <#${id}> channel.`);
       return false;
     }
 
@@ -49,6 +49,15 @@ export abstract class Plugin implements IPlugin {
 
     const response = this.container.channelService.hasPermission(channelName, this.permission);
     if (!response) {
+      const baseReply = `Please use this command in a \`${this.permission}\` channel.`;
+
+      if (this.permission.toString() === 'Private') {
+        message.reply(
+          `${baseReply} This is primarily the class channels, and any channels we haven't defined.`
+        );
+        return response;
+      }
+
       const channels = Object.values(Constants.Channels[this.permission]);
       const totalChannels = channels.length;
       channels.splice(3);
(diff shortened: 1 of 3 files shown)


c650:
your commits seem unsigned, just fyi

src/common/plugin.ts
```diff
@@ -28,7 +29,14 @@ export abstract class Plugin implements IPlugin {
   public hasPermission(message: IMessage): boolean {
     const channelName = this.container.messageService.getChannel(message).name;
     if (typeof this.pluginChannelName === 'string' && this.pluginChannelName !== channelName) {
-      message.reply(`Please use this command in the \`#${this.pluginChannelName}\` channel.`);
+      //if this fails it simply means that a channel the bot expected to exist didn't, nothing bot breaking
+      try {
+        const id = this.container.guildService.getChannel(this.pluginChannelName).id;
+        message.reply(`Please use this command in <#${id}>.`);
+      } catch (err) {
+        this.container.loggerService.warn(err);
```
c650, line 37: give a little context here maybe

c650, line 37: give a little context here maybe

src/common/plugin.ts
```diff
@@ -28,7 +29,14 @@ export abstract class Plugin implements IPlugin {
...
+        message.reply(`Please use this command in the \`#${this.pluginChannelName}\` channel.`);
+      }
```
c650, line 39: nit: this block is a tad un-DRY

src/common/plugin.ts
```diff
@@ -47,7 +55,32 @@ export abstract class Plugin implements IPlugin {
...
     const response = this.container.channelService.hasPermission(channelName, this.permission);
     if (!response) {
-      message.reply(`Please use this command in a \`${this.permission}\` channel`);
+      const rooms = Object.values(Constants.Channels[this.permission]);
+      let addonText = '';
+
+      //if this fails it simply means that a channel the bot expected to exist didn't
+      try {
+        if (rooms.length === 0 && this.permission.toString() != 'Private') {
+          addonText = 'their are no permanant channels oh this type declared.';
```
c650, line 64: *there
*permanent

c650: *on

src/common/plugin.ts
```diff
@@ -47,7 +55,32 @@ export abstract class Plugin implements IPlugin {
...
+        } else if (rooms.length === 1) {
+          const id = this.container.guildService.getChannel(rooms[0]).id;
+          addonText = `, <#${id}> is the only channel whith this type`;
+        } else {
+          const id = [
+            this.container.guildService.getChannel(rooms[0]).id,
+            this.container.guildService.getChannel(rooms[1]).id,
+          ];
+          if (rooms.length == 2) {
```
c650, line 73: ===

src/common/plugin.ts
```diff
@@ -47,7 +55,32 @@ export abstract class Plugin implements IPlugin {
...
+            addonText = `, <#${id[0]}> and <#${id[1]}> are the only two with this channel type`;
+          } else {
+            addonText = `, <#${id[0]}> and <#${id[1]}> are two of the channels this channel type`;
```
c650, line 76: would rather do a slice on `rooms` that caps the length at <=3, and then do a map&join

tanndlin:
I like the idea
image
Only issue is that there are some channels that are hidden to the user, but it'll still create a link for it. Only `#verify` afaik, but it should check to make sure that the user has the perms for that channel

This is what it would look like to a user without the perms
image

src/common/plugin.ts
CallumCoy: don't think the join is worth it, did the rest though.

CallumCoy: > I like the idea
> image
> Only issue is that there are some channels that are hidden to the user, but it'll still create a link for it. Only `#verify` afaik, but it should check to make sure that the user has the perms for that channel
> 
> This is what it would look like to a user without the perms
> image

Now checks to see if the user is aware of that channel.
image

src/common/plugin.ts
```diff
@@ -28,7 +29,16 @@ export abstract class Plugin implements IPlugin {
   public hasPermission(message: IMessage): boolean {
     const channelName = this.container.messageService.getChannel(message).name;
     if (typeof this.pluginChannelName === 'string' && this.pluginChannelName !== channelName) {
-      message.reply(`Please use this command in the \`#${this.pluginChannelName}\` channel.`);
+      let channel = '';
+      try {
+        channel = `<#${this.container.guildService.getChannel(this.pluginChannelName).id}>`;
+      } catch (err) {
+        this.container.loggerService.warn(
+          `Expected ${this.pluginChannelName} in Constants.ts, but it was not found.  Error info:\n ${err}`
```
c650, line 37: 'but it was not found' is redundant

src/common/plugin.ts
```diff
@@ -47,7 +57,42 @@ export abstract class Plugin implements IPlugin {
...
+        } else if (id.length === 0) {
+          addonText = ' There are no permanent channels of this type.';
+        } else if (id.length === 1) {
+          addonText = ` <#${id[0]}> is the only channel whith this type.`;
+        } else if (id.length > 1) {
+          addonText = ` <#${id[0]}> and <#${id[1]}> are `;
+          if (id.length === 2) {
+            addonText += `the only two with this channel type.`;
+          } else {
+            id.pop();
```
c650, line 85: pop shouldnt cause issues because its on a new array (thanks map) but i dont understand why you pop in the first place?

CallumCoy: Was for when it catches errors, but I just changed the code to do it another way.

src/common/plugin.ts
```diff
@@ -47,10 +55,54 @@ export abstract class Plugin implements IPlugin {
...
+          addonText = ' There are no permanent channels of this type.';
+        } else if (rooms.length === 1) {
+          addonText = ` <#${chanInfo[0].getID()}> is the only channel whith this type.`;
+        } else if (rooms.length > 1) {
+          addonText = ` <#${chanInfo[0].getID()}> and <#${chanInfo[1].getID()}> are `;
+          if (rooms.length === 2) {
+            addonText += `the only two with this channel type.`;
+          } else {
+            addonText += `two of the channels of this type.`;
+          }
```
tanndlin, line 85: You can condenses all this code into a map
``addonText = `Here are the supported channel(s): ${chanInfo.map((c) => `<#${c.getID()}>`).join(', ')}.`;``

src/common/types.ts
```diff
@@ -144,6 +144,23 @@ export interface IPluginLookup {
   [pluginName: string]: IPlugin;
 }
 
+export class CChannelInfo {
+  id: string;
```
tanndlin, line 148: What's the point in this class when you can get all this info straight from the GuildChannel class?

src/common/plugin.ts
```diff
@@ -47,10 +55,54 @@ export abstract class Plugin implements IPlugin {
 
     const response = this.container.channelService.hasPermission(channelName, this.permission);
     if (!response) {
-      message.reply(`Please use this command in a \`${this.permission}\` channel`);
+      const rooms = Object.values(Constants.Channels[this.permission]).splice(0, 3);
```
tanndlin, line 58: If the permission is `Public` this will return the first 3 public channels in Constants.ts, which are `Announcements`, `Welcome`, and `ClassInvite`. None of which the user can type in

src/common/plugin.ts
```diff
@@ -47,10 +55,54 @@ export abstract class Plugin implements IPlugin {
...
+      // not sure if this will cause issues as I do pop id, please let me know
+      const chanInfo: CChannelInfo[] = rooms
+        .filter((room) => {
+          return this.container.guildService
+            .getChannel(room)
+            .permissionsFor(message.member || '')
+            ?.has('VIEW_CHANNEL');
+        })
+        .map((room) => new CChannelInfo(this.container.guildService.getChannel(room).id, room))
+        .splice(0, 2);
```
tanndlin, line 70: You spliced twice, remove the first splice on line 58 and only do it here, after the filter

src/common/plugin.ts
```diff
@@ -47,10 +55,54 @@ export abstract class Plugin implements IPlugin {
...
+
+      try {
+        if (this.permission.toString() === 'Private') {
+          addonText = ` This is primarily the class channels, and any channels we haven't defined.`;
+        } else if (rooms.length === 0) {
```
tanndlin, line 75: `rooms.length` is not the same as `chanInfo.length`

src/common/types.ts
CallumCoy: So I can package up the info and send it to a function and not have any majog numbers.  Otherwise not sure on how I would do it as you can't send random Jsons through function in typeScript as far as I know.

src/common/plugin.ts
CallumCoy: Yeah I noticed that was debating about reordering the list so that it reflects what info we need better, or making a custom public like what I did with private.  Another idea I just came up with while responding is making a new channel type for channels like this, although I think that should be a superate pull request. @c650 @tanndlin Thoughts on adding a seperate channel type for these announcement type channels?

src/common/plugin.ts
CallumCoy: I splice twice for a reason, the first on is so we know if we have 2 or more channels or not.  The other one is because only the first 2 channels are needed beyond that point.

src/common/plugin.ts
CallumCoy: This is so the ending to the message changes accordingly.  After all we can't display all the channels else public would be a wall of text,  so I decided 2 displayed channels will be the cut-off point.

src/common/plugin.ts
CallumCoy: To build up on this I did not want to just on that last one as the map would be doing pointless work at that point.

CallumCoy: Actually changed how I write the message so this is actually fixed now.

src/common/plugin.ts
tanndlin: We could have an `info` channel type for those

src/common/types.ts
tanndlin: You could just send the whole `GuildChannel`

CallumCoy: I could have used 'this' after further inspection, however, I removed the class as I realized the channel id part was causing the catch to trigger so it wouldn't even show the error with that info.  So I just send the name of the channel now to the function now.

src/common/plugin.ts
```diff
@@ -55,50 +55,46 @@ export abstract class Plugin implements IPlugin {
...
-            .getChannel(room)
-            .permissionsFor(message.member || '')
-            ?.has('VIEW_CHANNEL');
-        })
-        .map((room) => new CChannelInfo(this.container.guildService.getChannel(room).id, room))
-        .splice(0, 2);
+      let addonText = '';
 
       try {
+        // not sure if this will cause issues as I do pop id, please let me know
```
tanndlin, line 65: left over comment

src/common/plugin.ts
```diff
@@ -55,50 +55,46 @@ export abstract class Plugin implements IPlugin {
...
-          addonText = ` <#${chanInfo[0].getID()}> is the only channel whith this type.`;
-        } else if (rooms.length > 1) {
-          addonText = ` <#${chanInfo[0].getID()}> and <#${chanInfo[1].getID()}> are `;
-          if (rooms.length === 2) {
-            addonText += `the only two with this channel type.`;
-          } else {
-            addonText += `two of the channels of this type.`;
-          }
+        } else if (id.length === 1) {
+          addonText = ` <#${id[0]}> is the only channel whith this type.`;
```
tanndlin, line 80: with*

tanndlin (changes requested):

src/common/plugin.ts
```diff
@@ -28,7 +29,14 @@ export abstract class Plugin implements IPlugin {
...
     if (typeof this.pluginChannelName === 'string' && this.pluginChannelName !== channelName) {
-      message.reply(`Please use this command in the \`#${this.pluginChannelName}\` channel.`);
+      let channel = '';
+      try {
+        channel = `<#${this.container.guildService.getChannel(this.pluginChannelName).id}>`;
+      } catch (err) {
+        this._errorGen([this.pluginChannelName], err);
+        channel = `\`#${this.pluginChannelName}\``;
+      }
+      message.reply(`Please use this command in the ${channel} channel.`);
```
tanndlin, line 39: There should never be an error here, since the required channel is hard coded in, and the channel names dont change.
This can just be
```
const id = this.container.guildService.getChannel(this.pluginChannelName);
message.reply(`Please use this command in the <#${id}> channel.`);
```

src/common/plugin.ts
```diff
@@ -47,10 +55,49 @@ export abstract class Plugin implements IPlugin {
...
+          })
+          .map((room) => this.container.guildService.getChannel(room).id);
+
+        if (this.permission.toString() === 'Private') {
+          addonText = ` This is primarily the class channels, and any channels we haven't defined.`;
+        } else if (id.length === 0) {
+          addonText = ' There are no permanent channels of this type.';
+        } else if (id.length === 1) {
+          addonText = ` <#${id[0]}> is the only channel with this type.`;
+        } else {
```
tanndlin, line 80: Instead of a bunch of else ifs
just return after each if
```
if(chan is private){
    tell user can only be in class
}

if(id.length === 0){
    blah;
    return;
}
```

src/common/plugin.ts
```diff
@@ -47,10 +55,49 @@ export abstract class Plugin implements IPlugin {
...
     if (!response) {
-      message.reply(`Please use this command in a \`${this.permission}\` channel`);
+      const rooms = Object.values(Constants.Channels[this.permission]);
+      const totalChannels = rooms.length;
+      rooms.splice(3);
+
+      let addonText = '';
+
+      try {
+        const id = rooms
```
tanndlin, line 65: nit: dont like this name

tanndlin (changes requested):

src/common/plugin.ts
CallumCoy: are else if's bad?

tanndlin: Not necessarily, but a lot of the time, they are very pretty, this code is very dense, and could use a little spacing. It makes it easier to follow too

src/common/plugin.ts
```diff
@@ -55,35 +49,41 @@ export abstract class Plugin implements IPlugin {
 
     const response = this.container.channelService.hasPermission(channelName, this.permission);
     if (!response) {
-      const rooms = Object.values(Constants.Channels[this.permission]);
-      const totalChannels = rooms.length;
-      rooms.splice(3);
+      const channels = Object.values(Constants.Channels[this.permission]);
+      const totalChannels = channels.length;
+      channels.splice(3);
```
c650, line 54: what happens if length<3?

src/common/plugin.ts
```diff
@@ -55,35 +49,41 @@ export abstract class Plugin implements IPlugin {
...
             return this.container.guildService
-              .getChannel(room)
+              .getChannel(channel)
               .permissionsFor(message.member || '')
               ?.has('VIEW_CHANNEL');
           })
           .map((room) => this.container.guildService.getChannel(room).id);
 
         if (this.permission.toString() === 'Private') {
           addonText = ` This is primarily the class channels, and any channels we haven't defined.`;
```
c650, line 69: return out of the if's and then make the last one not in an if

src/common/plugin.ts
CallumCoy: it just returns what it can. Assuming you are talking about the splice

CallumCoy:
I decided on 10 since there is still a variety of Channels in the Public type for constants

src/common/plugin.ts
```diff
@@ -47,10 +49,58 @@ export abstract class Plugin implements IPlugin {
...
+      if (this.permission.toString() === 'Private') {
+        message.reply(
+          `${baseReply} This is primarily the class channels, and any channels we haven't defined.`
+        );
+        return response;
+      }
+
+      const channels = Object.values(Constants.Channels[this.permission]);
+      const totalChannels = channels.length;
+      channels.splice(10);
```
tanndlin, line 63: magic number, also is 10 really necessary? That's quite a long error message returned to the user
image

tanndlin (changes requested):

tanndlin (changes requested):
While you are at it, can you make it so that it checks for permissions before checking validate pls?
https://github.com/cs-discord-at-ucf/lion/blob/a136558e23058cc1c3a7ef31d08db52444a27de6/src/app/handlers/command.handler.ts#L26-L33

tanndlin (approved):
lgtm

PR #330, opened 2021-03-25
Implemented a simple plugin for tellling people how to code + extra.

CallumCoy:
Quick plugin for telling people how to code.  Also implement a quick command to print a messages text into code format.

### DIFF
--- a/src/app/plugins/code.plugin.ts
+++ b/src/app/plugins/code.plugin.ts
@@ -10,7 +10,7 @@ export class CodePlugin extends Plugin {
   public permission: ChannelType = ChannelType.Public;
 
   private discordFormatingInfo: string =
-    'https://cdn.discordapp.com/avatars/574623716638720000/7d404c72a6fccb4a3bc610490f8d7b72.png';
+    'https://gist.github.com/matthewzring/9f7bbfd102003963f9be7dbcf7d40e51';
   private formattingMessage: string =
     `You can post in Discord like a real boss by refering to this guide ${this.discordFormatingInfo}.  The coding in discord information is located under the third header.\n` +
     `>>> How to code in discord, speed course edition: \n` +
@@ -24,31 +24,48 @@ export class CodePlugin extends Plugin {
   }
 
   public async execute(message: IMessage, args?: string[]) {
-    const input = args || ['how'];
+    const input = this._parseInput(args);
 
-    if (input[0] === '' || input[0] === 'how') {
+    if (input[0] === 'how') {
       message.channel.send(this.formattingMessage);
-    } else {
-      const messageID = this.isNumeric(input[0]) ? input[0] : input[0].split('/').pop();
+      return;
+    }
 
-      if (this.isNumeric(messageID)) {
-        message.channel.messages
-          .fetch({ around: messageID, limit: 1 })
-          .then((targMessage) => {
-            message.channel.send(
-              `\`\`\`${input[1] || ''}\n ${targMessage.first()?.content}\n\`\`\``
-            );
-          })
-          .catch((err) =>
-            this.container.loggerService.warn(
-              `Failed to find message id ${messageID} in channel <#${messageID}> extra infor if any:\n${err}`
-            )
-          );
-      }
+    const messageID = this._inputToMessageID(input[0]);
+
+    if (messageID) {
+      message.channel.messages
+        .fetch(messageID)
+        .then((targMessage) => {
+          message.channel.send(`\`\`\`${input[1] || ''}\n ${targMessage.content}\n\`\`\``);
+        })
+        .catch((err) =>
+          this.container.loggerService.warn(
+            `Failed to find message id ${messageID} in channel <#${messageID}> extra infor if any:\n${err}`
+          )
+        );
     }
   }
 
-  private isNumeric(possibleNumber: any): Boolean {
+  private _inputToMessageID(input: string): string {
+    const id = input.split('/').pop();
+
+    if (this._isNumeric(id)) {
+      return id || '';
+    }
+
+    return '';
+  }
+
+  private _parseInput(args?: string[]): string[] {
+    if (!args || args.length < 1) {
+      return ['how'];
+    }
+
+    return args;
+  }
+
+  private _isNumeric(possibleNumber: any): Boolean {
     return !isNaN(Number(possibleNumber));
   }
 }
--- a/src/app/plugins/code.plugin.ts
+++ b/src/app/plugins/code.plugin.ts
@@ -1,5 +1,6 @@
 import { Plugin } from '../../common/plugin';
 import { ChannelType, IContainer, IMessage } from '../../common/types';
+import Constants from '../../common/constants';
 
 export class CodePlugin extends Plugin {
   public name: string = 'Code Plugin';
@@ -11,6 +12,7 @@ export class CodePlugin extends Plugin {
 
   private discordFormatingInfo: string =
     'https://gist.github.com/matthewzring/9f7bbfd102003963f9be7dbcf7d40e51';
+
   private formattingMessage: string =
     `You can post in Discord like a real boss by refering to this guide ${this.discordFormatingInfo}.  The coding in discord information is located under the third header.\n` +
     `>>> How to code in discord, speed course edition: \n` +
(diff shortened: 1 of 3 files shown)


src/app/plugins/code.plugin.ts
```diff
@@ -0,0 +1,57 @@
+import Constants from '../../common/constants';
+import { Plugin } from '../../common/plugin';
+import { ChannelType, IContainer, IMessage } from '../../common/types';
+import { MessageEmbed } from 'discord.js';
+
+class Breed {
```
CallumCoy, line 6: I may have copied my Cat plugin as a base diagram shhh XD

src/app/plugins/code.plugin.ts
```diff
@@ -0,0 +1,49 @@
...
+  public pluginAlias = [''];
+  public permission: ChannelType = ChannelType.Public;
+
+  constructor(public container: IContainer) {
+    super();
+  }
+
+  public async execute(message: IMessage, args?: string[]) {
+    if (args === undefined) {
+      args = [''];
```
c650, line 19: re-assigning to parameter is bad practice

src/app/plugins/code.plugin.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    const input = args || ['', ''];
+
+    while (input.length < 2) {
+      input.push('');
+    }
```
c650, line 26: all this and the above code can just be the `const input =` line

src/app/plugins/code.plugin.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    if (input[0] === '' || input[0] === 'how') {
+      message.channel.send(
+        `You can post in Discord like a real boss by refering to this guide ${Constants.DiscordFormatingInfo}.  The coding in discord information is located under the third header.\n` +
+          `>>> How to code in discord, speed course edition: \n` +
+          `\'''<language exstension(optional)>` +
+          `\n<Code>\n` +
+          `\'''\n` +
+          `Replace the ' with \``
+      );
```
CallumCoy, line 28: Not sure if this breaks any good practices, but I think it makes this block a lot more readable.  If you could let me know if this is a nono.

src/app/plugins/code.plugin.ts
```diff
@@ -0,0 +1,48 @@
...
+    } else {
+      const messageID = !isNaN(Number(input[0])) ? input[0] : input[0].split('/').pop();
+
+      if (!isNaN(Number(messageID))) {
```
tanndlin, line 32: check the negative and return if true to reduce indentation
```
if(isNan(...){
   return;
}
```

tanndlin (changes requested):

src/app/plugins/code.plugin.ts
```diff
@@ -0,0 +1,48 @@
...
+      message.channel.send(
+        `You can post in Discord like a real boss by refering to this guide ${Constants.DiscordFormatingInfo}.  The coding in discord information is located under the third header.\n` +
+          `>>> How to code in discord, speed course edition: \n` +
+          `\'''<language exstension(optional)>` +
+          `\n<Code>\n` +
+          `\'''\n` +
+          `Replace the ' with \``
+      );
+    } else {
+      const messageID = !isNaN(Number(input[0])) ? input[0] : input[0].split('/').pop();
```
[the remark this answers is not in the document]
CallumCoy: It takes either so how is it less effort for the user?

src/app/plugins/code.plugin.ts
```diff
@@ -0,0 +1,48 @@
...
+    'Shows someones text in code form (it only works in the channel the command was called in). "Code how" posts a link with formatting help and a quick discord message coding explanation.';
+  public usage: string = 'code <message ID/link (optional)> <lang (optional)>/ code how';
+  public pluginAlias = [''];
+  public permission: ChannelType = ChannelType.Public;
+
+  constructor(public container: IContainer) {
+    super();
+  }
+
+  public async execute(message: IMessage, args?: string[]) {
```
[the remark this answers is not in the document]
CallumCoy: Why though?  This way you don't have to bother writing the how if you're lazy like me.  So when a freshy posts their code without formatting I can just !code, then go back to sleep.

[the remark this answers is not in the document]
tanndlin: Because it ensures that the plugin only runs with correct args, and it lets the user know that the args were not correct

src/app/plugins/code.plugin.ts
```diff
@@ -0,0 +1,54 @@
...
+      message.channel.send(this.formattingMessage);
+    } else {
+      const messageID = this.isNumeric(input[0]) ? input[0] : input[0].split('/').pop();
+
+      if (this.isNumeric(messageID)) {
+        message.channel.messages
+          .fetch({ around: messageID, limit: 1 })
+          .then((targMessage) => {
+            message.channel.send(
+              `\`\`\`${input[1] || ''}\n ${targMessage.first()?.content}\n\`\`\``
```
tanndlin, line 39: input[1] is possibly undefined

tanndlin (changes requested):

src/app/plugins/code.plugin.ts
CallumCoy: the || catch it

src/app/plugins/code.plugin.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    if (messageID) {
+      message.channel.messages
+        .fetch(messageID)
+        .then((targMessage) => {
+          message.channel.send(`\`\`\`${input[1] || ''}\n ${targMessage.content}\n\`\`\``);
+        })
+        .catch((err) =>
+          this.container.loggerService.warn(
+            `Failed to find message id ${messageID} in channel <#${messageID}> extra infor if any:\n${err}`
```
c650, line 44: *info

<#...> isnt helpful in logs

src/app/plugins/code.plugin.ts
```diff
@@ -0,0 +1,71 @@
...
+        );
+    }
+  }
+
+  private _inputToMessageID(input: string): string {
+    const id = input.split('/').pop();
+
+    if (this._isNumeric(id)) {
+      return id || '';
+    }
```
c650, line 55: seems like a contradiction?

src/app/plugins/code.plugin.ts
CallumCoy: Aight I didn't know if the bot reported errors to its logs channel or not.

CallumCoy:
I decided on 10 since even with a new channel type there is still a big variety.

src/app/plugins/code.plugin.ts
```diff
@@ -0,0 +1,72 @@
...
+        .catch((err) => {
+          const channelName = this.container.messageService.getChannel(message).name;
+          this.container.loggerService.warn(
+            `Failed to find message id: ${messageID} in the following channel ${channelName}, extra information if any:\n${err}`
+          );
+        });
+    }
+  }
+
+  private _inputToMessageID(input: string): string {
```
tanndlin, line 51: You should never leave the user with no output, that will be confusing to the user. This is why you should use `validate` to ensure the user gives good input

You could make sure, in validate, that the input, is either 'how' || isLink(args[0])

src/app/plugins/code.plugin.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    if (this._isNumeric(id)) {
+      return id;
+    }
+
+    return '';
+  }
+
+  private _parseInput(args?: string[]): string[] {
+    if (!args || args.length < 1) {
```
tanndlin, line 62: `args` is an array, which is always truthy, so `!args` is always false, therefore redundant.

src/app/plugins/code.plugin.ts
```diff
@@ -0,0 +1,72 @@
...
+      return ['how'];
```
tanndlin, line 63: return `['how', '']` here so you dont have to do a undefined check on line 40

tanndlin (changes requested):

src/app/plugins/code.plugin.ts
CallumCoy: I believe I would still have to do a undefined check if the user just gave a message Id/ link

src/app/plugins/code.plugin.ts
CallumCoy: Arg? Can be undefined which is a false truthy value.

tanndlin: args is an array, so when there are no args given, it's an empty array, which is true in JS

src/app/plugins/code.plugin.ts
```diff
@@ -45,7 +45,13 @@ export class CodePlugin extends Plugin {
             `Failed to find message id: ${messageID} in the following channel ${channelName}, extra information if any:\n${err}`
           );
         });
+      return;
     }
+
+    message.reply(
+      `\`${args?.join(' ')}\` is an invlaid arguement for ${this.name}.` +
+        `Please refer to the following on accepted inputs: \n \`${this.usage}\``
+    );
```
tanndlin, line 54: This is what the `validate` function is for

tanndlin (changes requested):

src/common/constants.ts
```diff
@@ -1,4 +1,5 @@
 const Constants = {
+  MaxCharCount: 2000,
```
tanndlin, line 2: nit: i feel `maxMessageLength` is better

tanndlin (approved):
lgtm

src/common/constants.ts
CallumCoy: Done

PR #331, opened 2021-03-26
added new constants section, and alphabetised them.

CallumCoy:
I think we have grown to the point that we need to split up the channels.  I also think we could benefit if we categorized the channels.


tanndlin:
Let's finish #329 and #330, then work on this one

src/common/constants.ts
```diff
@@ -4,54 +4,56 @@ const Constants = {
   LionPFP:
     'https://cdn.discordapp.com/avatars/574623716638720000/7d404c72a6fccb4a3bc610490f8d7b72.png',
   Channels: {
-    Public: {
+    Info: {
```
tanndlin, line 7: Put these in order as the are in the server, this is kind hard to look at lol

src/common/constants.ts
```diff
@@ -4,54 +4,56 @@ const Constants = {
...
-      Memes: 'memes',
+      CurcuitAnalysis: 'circuit_analysis',
+      Finance: 'finance',
+      Fitness: 'fitness',
+      FlameWars: 'flame_wars',
+      Food: 'food',
+      FoundationExam: 'foundation_exam',
+      GameDevelopment: 'game_development',
+      GameknightChat: 'gameknight_chat',
       General: 'general',
```
tanndlin, line 31: Put `#general` at the top of `public` so that #329 shows it first

tanndlin (changes requested):
Info should be for read-only channels at the top of the server, so `helpful_baubles` and `club_announcements` should be public, but I see where you are coming from

tanndlin:
I meant to put the channels inside of public, as they are in the server, so pretty much how it was before the PR. That was my mistake, I think I put the comment in the wrong spot.

tanndlin (approved):
lgtm

c650:
use of `Info` is fine with me but organising this file doesnt add much value because it's really a ctrl+f file

CallumCoy:
Are we waiting for me to do something on this pull?

PR #333, opened 2021-03-26
Paged Embed for Help Plugin

tanndlin:


src/app/plugins/help.plugin.ts
```diff
@@ -39,53 +41,65 @@ export class HelpPlugin extends Plugin {
...
-        `${Constants.Prefix}${plugin.usage}`,
-        `${type == 'adv' ? altCalls : ''}${plugin.description}`
-      );
+    // Create the pages
+    const numPages = Math.ceil(plugins.length / this.NUM_DISPLAY);
+    for (let i = 0; i < numPages; i++) {
+      pages.push(new MessageEmbed());
     }
+
+    pages.forEach((p) => {
```
c650, line 74: you could construct array of length `numPages` and then map it to embed

```
pages = new Array(numPages); // or something like this

pages = pages.map(...)
```

c650 (approved):

tanndlin:
I'll merge after #329 and #330 to make sure it works with the new plugins.

PR #335, opened 2021-03-28
Refactor Command Patterns

tanndlin:
Currently, I didn't override `validate` if there were no required args, for the sake of less lines. Lmk if I should include it anyways

### DIFF
--- a/src/app/plugins/modreport.plugin.ts
+++ b/src/app/plugins/modreport.plugin.ts
@@ -16,15 +16,19 @@ export class ModReportPlugin extends Plugin {
     super();
   }
 
-  public async execute(message: IMessage, args: string[]) {
-    const match_arr = args.join(' ').match(this._commandPattern);
+  public validate(message: IMessage, args: string[]) {
+    if (!args.length) return false;
 
+    const match_arr = args.join(' ').match(this._commandPattern);
     if (!match_arr) {
-      message.reply('Invalid syntax.');
-      return;
+      return false;
     }
 
-    const [sub_command, user_handle, description] = match_arr.slice(1);
+    return true;
+  }
+
+  public async execute(message: IMessage, args: string[]) {
+    const [sub_command, user_handle, description] = args;
 
     try {
       if (sub_command === 'add') {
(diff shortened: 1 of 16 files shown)


src/app/plugins/modreport.plugin.ts
```diff
@@ -10,22 +10,25 @@ export class ModReportPlugin extends Plugin {
...
+  public _commandPattern: RegExp = /(add|list|warn|ban)\s+([^#]+#\d{4})\s*(.*)/;
 
   constructor(public container: IContainer) {
     super();
   }
 
-  public async execute(message: IMessage, args: string[]) {
-    const match_arr = args.join(' ').match(this._commandPattern);
+  public validate(message: IMessage, args: string[]) {
+    if (!args.length) return false;
```
CallumCoy: why didn't you just do the following?
`
const match_arr = args.join(' ').match(this._commandPattern)
return !!match_arr
`

CallumCoy: Also, is there a reason you overwrite the validate function for this one despite it looking like it does the same thing?

tanndlin: Already changed that in another commit

src/app/plugins/slowmode.plugin.ts
```diff
@@ -10,15 +10,12 @@ export class SlowModePlugin extends Plugin {
   public pluginAlias = ['slow'];
   public permission: ChannelType = ChannelType.Staff;
   public pluginChannelName: string = Constants.Channels.Staff.ModChat;
+  public commandPattern: RegExp = /(\d+\s){2}[a-z]{3}\d{4}_[a-z]+/;
```
c650, line 13: this wont always work because i dont think cop3100c_fadfdad will match

c650 (commented):
this is an interesting idea but i would separate the commandpattern feature from the use of it in many many plugins... it will be easy to introduce bug. 

instead, make minimal pr for commandPattern and then just as we go along we can refactor older plugins.

one cool thing (but may not add much value), is actually deliver the match groups to the plugin so that we don't have to run the regex twice

tanndlin:
waiting for #337 before continuing

tanndlin:
> one cool thing (but may not add much value), is actually deliver the match groups to the plugin so that we don't have to run the regex twice
I feel like this should be a different PR

c650:
> I feel like this should be a different PR

definitely

c650 (approved):

c650:
let me run a deploy before we cehck this in

c650:
good to check this in whenever

tanndlin:
tested every plugin I edited in this PR 🤞

PR #336, opened 2021-03-30
Message overhaul

CallumCoy:


CallumCoy:
I should note don't push it if it magically has nothing to improve on, I gotta spread it around the bot.

src/services/message.service.ts
```diff
@@ -31,6 +31,46 @@ export class MessageService {
...
+    const breakChar = argBreakChar || '';
+
+    const replyLength = reply ? ('<@' + message.member?.id + '>,  ').length : 0;
+    const miscLength = header.length + footer.length;
+
+    let messagesToSend: string[] = [header + content + footer];
+
+    if (content.length + miscLength + replyLength > Constants.MaxMessageLength) {
+      if (!breakChar) {
+        return this._failedToSend(message, `Message was over Discords character cap.`);
```
c650, line 53: we shouldn't be telling the user that we failed to send them the message they want. it should be caught by caller of this function who then must handle it somehow.

src/services/message.service.ts
```diff
@@ -75,6 +115,36 @@ export class MessageService {
...
+  private _splitMessage(message: string, argBreakChar: string, numOfMiscChars: number): string[] {
+    const numOfFreeChars = Constants.MaxMessageLength - numOfMiscChars;
+    const breakChar = argBreakChar === '*' ? '(.|\n)' : argBreakChar;
+
+    const messageSplitRegExp: RegExp = new RegExp(
+      `(.|\n){1,${numOfFreeChars}}(${breakChar}|$)`,
+      'gm'
+    );
+
+    return message.match(messageSplitRegExp) || [];
```
c650, line 132: shouldnt suppress this error

src/services/message.service.ts
```diff
@@ -31,6 +31,46 @@ export class MessageService {
     this._sendConstructedReport(report, { files: message.attachments.map((e) => e.url) });
   }
 
+  async sendMessage(
+    message: IMessage,
+    content: string,
+    reply: boolean,
+    argHeader?: string,
+    argFooter?: string,
+    argBreakChar?: string
```
joey-colon, line 40: when requirements change for `sendMessage` it will be simpler to extend if you accept an optional object called `options` that contains the struct for additional parameters you can send to send message

src/services/message.service.ts
```diff
@@ -31,6 +31,46 @@ export class MessageService {
...
+    if (content.length + miscLength + replyLength > Constants.MaxMessageLength) {
+      if (!breakChar) {
+        return this._failedToSend(message, `Message was over Discords character cap.`);
+      }
+
+      messagesToSend = this._splitMessage(content, breakChar, replyLength + miscLength);
+      if (!this._testMessages(messagesToSend, content.length, replyLength)) {
+        return this._failedToSend(message, `Failed to split the message up within the rule set.`);
+      }
+      messagesToSend = messagesToSend.map((message) => header + message + footer);
```
joey-colon, line 60: np: see above

src/services/message.service.ts
```diff
@@ -31,6 +31,46 @@ export class MessageService {
...
+    }
+
+    if (reply) {
+      message.reply(messagesToSend.shift() || '');
```
joey-colon, line 64: if `reply` is true, shouldn't we reply to the user with all chunks of the message?

src/services/message.service.ts
```diff
@@ -31,6 +31,46 @@ export class MessageService {
...
+    reply: boolean,
+    argHeader?: string,
+    argFooter?: string,
+    argBreakChar?: string
+  ): Promise<Boolean> {
+    const header = argHeader ? argHeader + '\n' : '';
+    const footer = argFooter ? '\n' + argFooter : '';
+    const breakChar = argBreakChar || '';
+
+    const replyLength = reply ? ('<@' + message.member?.id + '>,  ').length : 0;
```
tanndlin, line 46: You can `@someone` by using `Hello ${message.author}`

src/services/message.service.ts
```diff
@@ -31,6 +31,46 @@ export class MessageService {
...
+      }
+      messagesToSend = messagesToSend.map((message) => header + message + footer);
+    }
+
+    if (reply) {
+      message.reply(messagesToSend.shift() || '');
+    }
+
+    messagesToSend.forEach((val) => {
+      message.channel.send(val);
```
joey-colon, line 68: iirc `.send()` is a promise. can we use a regular for loop and await each message send? if any fail, we should let the client know via exception

src/services/message.service.ts
```diff
@@ -31,6 +31,46 @@ export class MessageService {
...
   }
 
+  async sendMessage(
+    message: IMessage,
+    content: string,
+    reply: boolean,
+    argHeader?: string,
+    argFooter?: string,
+    argBreakChar?: string
+  ): Promise<Boolean> {
```
joey-colon, line 41: im not really a fan of returning a boolean since we already have fulfillment/rejecting out of the box with promises. it makes more sense to make type `Promise<void>` and use reject/fufillment as flags for the status of the call

src/services/message.service.ts
```diff
@@ -31,6 +31,46 @@ export class MessageService {
...
+    const header = argHeader ? argHeader + '\n' : '';
+    const footer = argFooter ? '\n' + argFooter : '';
+    const breakChar = argBreakChar || '';
```
joey-colon, line 44: instead of `breakChar` can we use `delimiter`?

src/services/message.service.ts
```diff
@@ -31,6 +31,46 @@ export class MessageService {
...
+
+    const replyLength = reply ? ('<@' + message.member?.id + '>,  ').length : 0;
+    const miscLength = header.length + footer.length;
```
joey-colon, line 47: i think we can make this name a little more insightful. `templateLength` might give off more signal to what the relevance of this variable is

src/services/message.service.ts
```diff
@@ -31,6 +31,46 @@ export class MessageService {
     this._sendConstructedReport(report, { files: message.attachments.map((e) => e.url) });
   }
 
+  async sendMessage(
```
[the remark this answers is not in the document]
joey-colon: unless, this function will support both types of messages (by all means go for it!)

src/services/message.service.ts
joey-colon: an additional requirement that would be nice in the future is to support embeds, accepting relevant embed parameters in this `options` object would be make the interface much simpler for consumers.

CallumCoy: a separate function for embed I believe. @tanndlin am I correct on that?

tanndlin: Yeah, you can split it up into embeds then send it thru `messageService.sendPagedEmbed`
https://github.com/cs-discord-at-ucf/lion/blob/ff3e3a996aa603e78c4f5662a8d7616d98b29dde/src/services/message.service.ts#L42

src/services/message.service.ts
CallumCoy: we can, but I think it looks better not having your tag on all of them.  This is what it looks like when you do it the current way for reference.
image

src/services/message.service.ts
tanndlin: instead, since you already have a map. just
```
await Promise.all(messagesToSend.map((val) => {
    message.channel.send(val);
});
```

src/services/message.service.ts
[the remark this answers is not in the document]
CallumCoy: I think the audio one is a format that you can add to your message; nonetheless, I'll change it for clarity.

src/services/message.service.ts
```diff
@@ -75,6 +115,36 @@ export class MessageService {
...
+    messages.map((message) => {
+      messageCheckCount += message.length;
+    });
+
+    if (testLength != messageCheckCount) {
+      return false;
+    }
+    return true;
+  }
+
```
tanndlin: Since this PR is for a method on a `.service` all uses of these methods will be hard coded.
There will no variability in production so, there should never be an error creating the message, thus no reason to have this function

tanndlin (changes requested):

src/services/message.service.ts
CallumCoy: It can fail, if the returned text lacks the break character in the 2000 (- template and reply length) character limit.  Then it will parts of the message out.  Granted this case very unlikely, but it is still a possibility.

src/services/message.service.ts
CallumCoy: I have kinda added the "option" argument.  I had to make an interface for it since I could not figure out how to get it to work nicely with strings, without it bloating the code.  And the only sources I found on google were only for Java.  If you wanted it without an interface I would appreciate a link or example code.

src/services/message.service.ts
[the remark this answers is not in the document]
CallumCoy: is `return Promise.reject()` okay or would you prefer to throw an error?

CallumCoy:
Fun fact I found [this](https://discord.js.org/#/docs/main/stable/typedef/MessageOptions) towards the end of today XD.  The function I wrote only does one, possibly two things that it doesn't. Also, I'll figure out that DeepScan issue when I wake up.

src/services/message.service.ts
```diff
@@ -75,6 +117,35 @@ export class MessageService {
...
+    const messageSplitRegExp: RegExp = new RegExp(
+      `(.|\n){1,${numOfFreeChars}}(${delimiter}|$)`,
+      'gm'
+    );
+
+    return message.match(messageSplitRegExp);
+  }
+
+  private _testMessages(messages: string[], testLength: number): boolean {
```
c650, line 136: what is this for

tanndlin (commented):
There's definetly a better way to split messages, but work on the other comments, then we'll see

src/services/message.service.ts
```diff
@@ -31,6 +31,48 @@ export class MessageService {
...
+    argOptions: ITextMessageOptions
+  ): Promise<void> {
+    const options: ITextMessageOptions = {};
+    options.header = argOptions.header ? `${argOptions.header}\n` : '';
+    options.footer = argOptions.footer ? `\n${argOptions.footer}` : '';
+    options.reply = argOptions.reply || false;
+    options.delimiter = argOptions.delimiter || '';
+
+    const replyLength = options.reply ? `${message.author},  `.length : 0;
+    const templateLength = options.header.length + options.footer.length;
```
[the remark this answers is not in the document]
CallumCoy: Not sure where you got 2048 from, but in testing and what I have seen online 2000 is the solid cap for discord message length.  Also thought about it while I was coding it, but decided not to simply because this route only takes 1 more variable.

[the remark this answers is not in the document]
tanndlin: idk, you are probably right. Nonetheless, If you just give yourself some buffer, it makes things a lot less complex
```ts
const chunks = [];
const split = content.split('\n');

while (split.length > 0) {
  let temp = '';
  while (temp.length < 1000 && split.length > 0) {
    temp = `${split.pop()}\n` + temp;
  }
  chunks.push(temp);
}

await Promise.all(
  chunks.reverse().map((c) => {
    return message.channel.send(`${options.header}${c}${options.footer}`);
  })
);
```
Something along those lines

src/app/plugins/code.plugin.ts
```diff
@@ -10,6 +9,12 @@ export class CodePlugin extends Plugin {
   public pluginAlias = [];
   public permission: ChannelType = ChannelType.Public;
 
+  private codeMessageOptions: ITextMessageOptions = {
+    reply: false,
+    delimiter: '\n',
+    footer: '```',
+  };
```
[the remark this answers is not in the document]
tanndlin: My mistake, I thought this was in `message.service`.

src/services/message.service.ts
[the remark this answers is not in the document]
CallumCoy: I can do that if you think that is a better method, although I think this way is more like sweeping the problem under the rug.  Since I thought the key idea of this service was to catch when messages are too big, opposed to simply splitting them up which is the primary function of that method.  Although if you think the primary issue with the current code is it's too complex, then I can try to encapsulate it a bit more and change around some variables to try and make it read better.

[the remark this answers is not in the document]
tanndlin: How is this sweeping it under the rug? If the message is too big, it splits it. You can tweak it a little bit, you can make it 1900 if you want, Im just saying if you give yourself reasonable headroom the code becomes a lot simpler.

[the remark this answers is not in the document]
tanndlin: if(not over the limit){
dont nothing
}

src/services/message.service.ts
```diff
@@ -31,6 +31,55 @@ export class MessageService {
...
+  async sendTextMessage(
+    message: IMessage,
+    content: string,
+    argOptions: ITextMessageOptions
+  ): Promise<void> {
+    const options: ITextMessageOptions = {};
+    options.header = argOptions.header ? `${argOptions.header}\n` : '';
+    options.footer = argOptions.footer ? `\n${argOptions.footer}` : '';
+    options.reply = argOptions.reply || false;
+    options.delimiter = argOptions.delimiter === '*' ? '' : argOptions.delimiter;
```
c650, line 43: wont this sometimes be undefined?

CallumCoy: If you're u mean delimiter, yes it can at which point it's either not used, or is caught and a error is thrown on line 51.  If it's any other the other ones I thought the above code would deal with it.

src/services/message.service.ts
```diff
@@ -31,6 +31,55 @@ export class MessageService {
...
+
+    const replyLength = options.reply ? `${message.author},  `.length : 0;
+    const templateLength = replyLength + options.header.length + options.footer.length;
+
+    let messagesToSend: string[] | null = [`${options.header}${content}${options.footer}`];
+
+    if (content.length + templateLength > Constants.MaxMessageLength) {
+      if (options.delimiter === undefined) {
```
c650, line 51: prefer !options.delimiter

c650 (approved):

src/services/message.service.ts
CallumCoy: mb there, I was basically asleep when I finished up with it, and I thought there wasn't a way to achieve this.

CallumCoy:
@joey-colon if you could check out my last replies to your comments and let me know if you want anything else doing with that info.

src/app/handlers/require_url.handler.ts
```diff
@@ -21,9 +21,17 @@ export class RequireUrlHandler implements IHandler {
...
+          .sendTextMessage(
+            message,
+            `Hey ${message.author},\n We require for you to include a link to your message in the #${channelObj.name} channel.\n\n Here's your message content:\`\`\`${message.content}\`\`\` `,
+            { delimiter: ' ' }
+          )
+          .catch((err) => {
+            this.container.loggerService.warn(
+              `Error happened when asking ${message.author}, to insert a link inert a link in ${channelObj.name}. Error info:\n${err}`
+            );
+          });
```
tanndlin, line 34: imo there is no reason to add 10 lines of bloat for something that will never throw an error.
I feel like this is a useless abstraction

tanndlin (commented):
I feel like this is lots of bloat. If we do you this function we could catch in the function itself, rather than having a 4 line long catch in every plugin.
Plus, sending these messages will never throw an error in the first place. We should probably only use this function for when we are sending variably-lengthed  large messages

tanndlin:
> I feel like this is lots of bloat. If we do you this function we could catch in the function itself, rather than having a 4 line long catch in every plugin.
> Plus, sending these messages will never throw an error in the first place. We should probably only use this function for when we are sending variably-lengthed large messages

The only time we've ever had a message sending related error is when the message was too big, so lets only focus there

src/app/handlers/require_url.handler.ts
```diff
@@ -21,9 +21,17 @@ export class RequireUrlHandler implements IHandler {
     }
     try {
       if (!message.content.toLowerCase().match(this._url_regex)) {
-        message.author.send(
-          `Hey ${message.author},\n We require for you to include a link to your message in the #${channelObj.name} channel.\n\n Here's your message content:\`\`\`${message.content}\`\`\` `
-        );
+        this.container.messageService
+          .sendTextMessage(
+            message,
```
tanndlin, line 26: This was DMed before

src/app/handlers/require_url.handler.ts
c650: +1

src/app/handlers/require_url.handler.ts
c650: very good catch lol

src/app/handlers/require_url.handler.ts
CallumCoy: This specific one you highlighted I think is one of the major inputs that can be exploited since all you would have to do is send a long message in links which will cause an error.  As for the rest I chose based upon who has access to it and how likely the message is to get pretty big, so if you could list the ones that you want removing specifically I would appreciate that.  Also referring to the text message function dealing with errors itself, I thought c650 wanted it to be outside of the function mentioned [here](https://github.com/cs-discord-at-ucf/lion/pull/336#discussion_r604073697) if you could clarify where you want the catching/notifying if any, please and thank you @c650.

CallumCoy:
> > I feel like this is lots of bloat. If we do you this function we could catch in the function itself, rather than having a 4 line long catch in every plugin.
> > Plus, sending these messages will never throw an error in the first place. We should probably only use this function for when we are sending variably-lengthed large messages
> 
> The only time we've ever had a message sending related error is when the message was too big, so lets only focus there

Not sure what you mean by this, as I never even thought about anything else. With the exception of that word count I implemented as a check, which has since been removed.

CallumCoy:
Just so you are informed my weekend got busy out of nowhere so any changes probably won't happen till Monday evening.

src/app/handlers/require_url.handler.ts
tanndlin: What he meant was, don't tell the user that we messed up, if there was an error we need to handle it somehow without says `Hey man, we messed up and couldnt send your message`

tanndlin: > This specific one you highlighted I think is one of the major inputs that can be exploited since all you would have to do is send a long message in links which will cause an error.

`Hey ${message.author},\n We require for you to include a link to your message in the #${channelObj.name} channel.\n\n Here's your message content:\`\`\`${message.content}\`\`\` `
Oh, I see what you are saying now. You need to make sure it DM's the user though, it didn't used to be send to the channel.

src/app/plugins/addclasschans.plugin.ts
```diff
@@ -167,7 +167,15 @@ export class AddClassChannelsPlugin extends Plugin {
...
-    message.reply(response);
+    this.container.messageService
+      .sendTextMessage(message, response, {
+        reply: true,
+      })
+      .catch((err) => {
+        this.container.loggerService.warn(
+          `Failed to write message to ${message.author.username}, when they requested to create more classes. Error info:\n${err}`
+        );
+      });
```
tanndlin, line 178: Not necessary since the original message the user sent will be longer than this one.

src/app/plugins/addroles.plugin.ts
```diff
@@ -42,7 +42,16 @@ export class AddRolesPlugin extends Plugin {
...
+      this.container.messageService
+        .sendTextMessage(message, `Successfully added: ${roles_added.join(', ')}`, {
+          reply: true,
+          delimiter: ' ',
+        })
+        .catch((err) => {
+          this.container.loggerService.warn(
+            `Failed to list successfully added roles to ${message.author.username}. Error info:\n${err}`
+          );
+        });
```
tanndlin, line 54: Not necessary since we do not have enough roles to takes up nearly 2000 chars

src/app/plugins/broadcast.plugin.ts
```diff
@@ -127,11 +127,20 @@ export class BroadcastPlugin extends Plugin {
...
+
+    this.container.messageService
+      .sendTextMessage(message, content, {
+        reply: true,
+      })
+      .catch((err) => {
+        this.container.loggerService.warn(
+          `Failed to ask for comfirmation on an announcement from ${message.author.username}. Error info:\n${err}`
+        );
+      });
```
tanndlin, line 143: Not necessary. This plugin is usually sent to all classes so it just says `all` otherwise, just a select few of classes.

src/app/plugins/chanban.plugin.ts
```diff
@@ -50,9 +50,17 @@ export class ChanBanPlugin extends Plugin {
...
+          .sendTextMessage(
+            message,
+            `Banned user from ${successfully_banned_channels.map((c) => c.name).join(', ')}`,
+            { reply: true, delimiter: ' ' }
+          )
+          .catch((err) => {
+            this.container.loggerService.warn(
+              `Failed to inform ${message.author.username} on what channels ${username} was banned from successfully. Error info:\n${err}`
+            );
+          });
```
tanndlin, line 63: No one is going to get banned in 2000 chars worth of channels. They would just be outright banned before that 😆

src/app/plugins/checkclasses.plugin.ts
```diff
@@ -48,25 +47,18 @@ export class CheckClassesPlugin extends Plugin {
...
-    const splitChans = [chanNames.slice(0, middle), chanNames.slice(middle)]; //Split the classes into 2 equal groups
+    const chansForMessage = chansContainingUser.map((c) => c.name).join(' | ');
 
-    //For each group, join to string
-    return splitChans.map((chans) => chans.join(' | '));
+    this.container.messageService
+      .sendTextMessage(
+        message,
+        `User is registered for \`${chansContainingUser.length}\` classes:\n\`${chansForMessage}\``,
+        { reply: true, delimiter: ' ' }
```
tanndlin, line 56: image
This doesn't wrap in code across multiple messages correctly

src/app/plugins/delroles.plugin.ts
```diff
@@ -40,7 +40,16 @@ export class DelRolesPlugin extends Plugin {
...
+      this.container.messageService
+        .sendTextMessage(message, `Successfully deleted: ${roles_deleted.join(', ')}`, {
+          reply: true,
+          delimiter: ' ',
+        })
+        .catch((err) => {
+          this.container.loggerService.warn(
+            `Failed to inform ${message.author.username} what roles were successfully deleted from them. Error info:\n${err}`
+          );
+        });
```
tanndlin, line 52: same as `addroles.plugin`

src/app/plugins/dog.plugin.ts
```diff
@@ -29,12 +41,22 @@ export class DogPlugin extends Plugin {
...
+        .sendTextMessage(
+          message,
+          `Breeds supported: \n\`\`\`\n${this._breeds.join('\n')}\`\`\``,
+          this.messageOptions
+        )
+        .catch((err) => {
+          this.container.loggerService.warn(
+            `Failed to produce the dog breed list. Error info:\n${err}`
+          );
+        });
```
tanndlin, line 59: We don't have nearly that many breeds

src/app/plugins/register.plugin.ts
```diff
@@ -68,6 +68,12 @@ export class RegisterPlugin extends Plugin {
...
       messageForUser += `\nUnable to locate these classes: ${invalidClasses.join(' ')}`;
     }
-    message.reply(messageForUser);
+    this.container.messageService
+      .sendTextMessage(message, messageForUser, { reply: true, delimiter: ' ' })
+      .catch((err) =>
+        this.container.loggerService.warn(
+          `Failed to give feedback to ${message.author.username} abouts registering to classes. Error info:\n${err}`
+        )
+      );
```
tanndlin, line 77: Not necessary, the user is only allowed to register for 5 classes at max, at a time.

src/services/message.service.ts
```diff
@@ -31,6 +31,57 @@ export class MessageService {
...
+    options.footer = argOptions.footer ? `\n${argOptions.footer}` : '';
+    options.reply = argOptions.reply || false;
+    options.delimiter = argOptions.delimiter === false ? false : argOptions.delimiter || '\n';
+
+    const replyLength = options.reply ? `${message.author},  `.length : 0;
+    const templateLength = replyLength + options.header.length + options.footer.length;
+
+    let messagesToSend: string[] | null = [`${options.header}${content}${options.footer}`];
+
+    if (content.length + templateLength > Constants.MaxMessageLength) {
```
tanndlin, line 50: Check the negative and return

src/services/message.service.ts
```diff
@@ -31,6 +31,57 @@ export class MessageService {
...
+      if (!options.delimiter) {
+        return Promise.reject(`Message was over Discords character cap.`);
+      }
```
tanndlin, line 53: This is unreachable code since delimiter has a default value. I don't think its necessary anyways

src/services/message.service.ts
```diff
@@ -31,6 +31,57 @@ export class MessageService {
...
+        options.delimiter,
+        options.header,
+        options.footer
+      );
+
+      messagesToSend.forEach((data) => {
+        if (data.length > Constants.MaxMessageLength) {
+          return Promise.reject(`Unable to split the message up this much.`);
+        }
+      });
```
tanndlin, line 69: What's the point of the `Promise.reject` in here? Plus it does nothing because all it does is return out of the `forEach`

src/services/message.service.ts
```diff
@@ -75,6 +126,38 @@ export class MessageService {
...
+    header: string,
+    footer: string
+  ): string[] {
+    const messagesToSend = [];
+    const numOfFreeChars = Constants.MaxMessageLength - templateLength;
+
+    let temp = '';
+
+    while (splitMessage.length > 0) {
+      let nextCapsule = splitMessage.shift() || '';
```
tanndlin, line 142: `shift` is really inefficient, just pop, and reverse at the end. [Ref](https://github.com/cs-discord-at-ucf/lion/pull/336#discussion_r605117249)

src/services/message.service.ts
CallumCoy: `export interface ITextMessageOptions {
  header?: string;
  footer?: string;
  reply?: boolean;
  delimiter?: string | false;
}`

`options.delimiter = argOptions.delimiter === false ? false : argOptions.delimiter || '\n';`

can activate via false value.

src/services/message.service.ts
CallumCoy: idea was to catch if any messages from a set failed to send. requested [here](https://github.com/cs-discord-at-ucf/lion/pull/336#discussion_r604184280)  Honestly not sure how you would check to see all promises were fulfilled.

src/app/plugins/dog.plugin.ts
CallumCoy: I can hope someone volunteers to add to the possible breeds ;-;

src/app/plugins/checkclasses.plugin.ts
CallumCoy: I'll add a title option, that goes before the header on the first message.

src/app/plugins/chanban.plugin.ts
CallumCoy: Sounds like a challenge

src/app/plugins/broadcast.plugin.ts
CallumCoy: It doesn't even list the classes, I put it in since you can make a long announcement.  Granted I have never seen one so I don't even know what you guys usually use it for.

src/services/message.service.ts
tanndlin: The return just returns out of the forEach, so it doesn't do anything though

src/services/message.service.ts
CallumCoy: Why?  A really high percentage of the messages that reach this part need to run the code after the if statement either-or.

tanndlin: I makes the code more readable
```
if(dont need to edit){
    send original message
    return
}

split up...
```

src/services/message.service.ts
CallumCoy: The addition of the title option makes the code less readable and more repetitive so I don't think it is worth the performance gain.

tanndlin: What does this have to do with the title?
https://github.com/CallumCoy/lion/blob/b466497018653d926536b11f4aced512f05759af/src/services/message.service.ts#L58
reverse this
https://github.com/CallumCoy/lion/blob/b466497018653d926536b11f4aced512f05759af/src/services/message.service.ts#L142-L148
And pop instead of shift

CallumCoy: I know what you are telling me to do, but as I said it makes the code less readable since you end up adding the title after you split the message.  It also makes the code repetitive since you exit the while loops, but then have to enter a similar one to split apart the title.  Which as I said, I don't think adding that complexity is worth the potential performance boost.  And the other option is to make the while loops have branches lowering the readability of those while loops that are already not very readable.

CallumCoy: For reference, it would look something like [this](https://www.codepile.net/pile/kBlmM75r)  current code at line 70.

src/services/message.service.ts
```diff
@@ -36,50 +36,51 @@ export class MessageService {
...
-        return Promise.reject(`Message was over Discords character cap.`);
+    if (messagesToSend.length > Constants.MaxMessageLength) {
+      if (options.dm) {
+        return this._deliverMessageInDm(options.dm, messagesToSend);
       }
+      return this._deliverMessageInChan(message, messagesToSend, !!options.reply);
+    }
 
-      options.delimiter = argOptions.delimiter === '*' ? '' : options.delimiter;
+    if (!options.delimiter) {
```
tanndlin, line 58: Everything being sent into this function will be hard coded in a plugin, so there should never be an error.

src/services/message.service.ts
```diff
@@ -129,35 +130,68 @@ export class MessageService {
...
-    if (temp) {
-      messagesToSend.push(`${header}${temp}${footer}`);
+    if (temp != options.header) {
+      messagesToSend.push(`${temp}${options.footer}`);
     }
 
     return messagesToSend;
   }
 
+  private async _deliverMessageInChan(
```
tanndlin, line 160: lots of repeated code between this and `deliverMessageInDm`

tanndlin (commented):
3 things
1. I feel that this whole thing is a lot more complicated than it has to be. Way to much abstraction

2. There is no reason to throw around Promise.rejects without doing anything with them.

3. A huge PR across many files like this with no rhyme or reason to the commits make it hard to see what is going on / being changed

src/app/plugins/broadcast.plugin.ts
```diff
@@ -132,15 +132,9 @@ export class BroadcastPlugin extends Plugin {
...
-        reply: true,
-      })
-      .catch((err) => {
-        this.container.loggerService.warn(
-          `Failed to ask for comfirmation on an announcement from ${message.author.username}. Error info:\n${err}`
-        );
-      });
+    this.container.messageService.sendTextMessage(message, content, {
+      reply: true,
+    });
```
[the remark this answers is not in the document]
CallumCoy: It's not though it literally prints what a user just sent after the keyword.

src/services/message.service.ts
```diff
@@ -36,50 +36,51 @@ export class MessageService {
     content: string,
     argOptions: ITextMessageOptions
   ): Promise<void> {
-    const options: ITextMessageOptions = {};
+    const options: ITextMessageOptions = argOptions;
+
+    options.title = argOptions.title?.substring(0, 1000) || '';
```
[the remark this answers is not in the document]
CallumCoy: https://github.com/cs-discord-at-ucf/lion/pull/336#discussion_r606816262 how else do you think we can fix this if we don't separate content from intro information?  Unless you want the user is subscribed thing on every message

src/services/message.service.ts
CallumCoy: Not sure what you are getting at here, since we have proved some plugins can return messages that are too long already.

tanndlin: Using this function without a delimiter is literally the same as using the discord `send` function. So it should be required

c650:
> 3 things
> 
> I feel that this whole thing is a lot more complicated than it has to be. Way to much abstraction
> 
> There is no reason to throw around Promise.rejects without doing anything with them.
> 
> A huge PR across many files like this with no rhyme or reason to the commits make it hard to see what is going on / being changed

strongly agree with @tanndlin. this pr overwhelms me

src/services/message.service.ts
```diff
@@ -129,35 +130,68 @@ export class MessageService {
...
+          return Promise.reject();
+        });
+        return Promise.resolve();
+      })
+    );
+
+    return status[-1];
+  }
+
+  private async _deliverMessageInDm(target: User, content: string[]): Promise<void> {
```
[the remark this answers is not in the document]
CallumCoy: going to need to know what you want it as then since it doesn't like Promise<String>

CallumCoy:
> 3 things
> 
>     1. I feel that this whole thing is a lot more complicated than it has to be. Way to much abstraction
> 
>     2. There is no reason to throw around Promise.rejects without doing anything with them.
> 
>     3. A huge PR across many files like this with no rhyme or reason to the commits make it hard to see what is going on / being changed

1. Current commit I think is the most readable it can get without starting to remove features.
2. They are simply there so if a plugin wants to know why a message failed it can access that info for whatever reason.
3. The 3 changed files that look like nothing to do with the change each have a reason.
     code.ts: You specifically asked for that change.
     checkClasses.ts: the code was unnecessary with the messaging plugin,  however as I mentioned before if you are still planning on changing this plugin I can remove the file completely.
    dog.ts:  I'm too lazy to turn off fomat on save.

src/services/message.service.ts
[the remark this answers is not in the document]
tanndlin: Either make it return something useful or just remove it
thats like saying
```
function printSomething(a) : null{
   ...
   return null;
}

you might as well just remove it, or better yet, return the promise that is given by sending a message. ya know?
```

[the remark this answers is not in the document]
CallumCoy: it does do that currrently though

`
    const status = await Promise.all( //runs until a promise is rejected
      messagesContent.map((item) => {// sends all the messages into order until a message fails, at this point something really bad happened if it fails
        channel.send(item, { split: { char: delimiter } }).catch((err) => {
          this._loggerService.warn(`Failed to construct all messages.  Error info:\n${err}`);
          return Promise.reject(`Failed to construct all messages.  Error info:\n${err}`);
        });
        return Promise.resolve();
      })
    );

    //return last Promise which will be reject, if any sends failed
    return status[-1];
`

CallumCoy:
obsolete Found out you can use messageSplit from discord.js https://discord.js.org/#/docs/main/stable/typedef/SplitOptions

PR #338, opened 2021-03-31
Fixed overwrite of permissions

tanndlin:


tanndlin:
> great. while you are at it, any other instances (ctrl+f)?

Double checked, only place is in `addclasschans.plugin.ts`, which is where it should be

PR #342, opened 2021-04-14
Fixed DeepScan Issue in Register Plugin

tanndlin:
DeepScan must've gotten an update, and some old code no longer passed the scan.


c650 (approved):
Weird, but it’s a good point I guess

PR #344, opened 2021-04-16
Sped up paged embeds

tanndlin:


tanndlin:
Sped up the paged embeds cuz they were kinda slow. Also removed the emojis once the collector was done so that people know it wont work.

c650 (approved):
POG

PR #356, opened 2021-04-26
Return in Dog plugin if API Error

### LINKED ISSUE #351: dog API
when the dog API is down bot posts `supported breeds` with an empty list.  We probably should make it so it tells the caller the API is napping.

tanndlin:
fixes #351


tanndlin:
Merging in a different branch

PR #358, opened 2021-04-28
Better Logging for start/stop

### LINKED ISSUE #284: Logging Improvements
For all message handlers, we should have 

"Handler xx Dispatching Message Id #"

and then for plugins/etc it should be what it is but include the message ID

Also we should log as JSON, not a sentence, for easier queries

tanndlin:
Refs #284

### DIFF
--- a/src/app/handlers/command.handler.ts
+++ b/src/app/handlers/command.handler.ts
@@ -1,4 +1,4 @@
-import { IMessage, IContainer, IHandler } from '../../common/types';
+import { IMessage, IContainer, IHandler, IPluginEvent } from '../../common/types';
 import Constants from '../../common/constants';
 
 export class CommandHandler implements IHandler {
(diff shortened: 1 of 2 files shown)


src/app/handlers/command.handler.ts
```diff
@@ -34,12 +34,16 @@ export class CommandHandler implements IHandler {
 
       try {
         this.container.loggerService.info(
-          `Running ${plugin.name} with args: {${command.args.join(', ')}}'`
+          JSON.stringify({ event: 'starting', plugin: plugin.name, args: command.args })
         );
         await plugin.execute(message, command.args);
-        this.container.loggerService.info(`The ${plugin.name} has been fulfilled`);
+        this.container.loggerService.info(
+          JSON.stringify({ event: 'fulfill', plugin: plugin.name, args: command.args })
```
c650, line 41: fulfillCommand

c650 (commented):
I wouldnt say that this fixes the issue because there are lots of logs to convert to structured iirc

c650:
but ill happily accept this as a `Refs`

c650 (commented):
maybe we should introduce an Event type ? or PluginEvent?

tanndlin:
> maybe we should introduce an Event type ? or PluginEvent?

can you explain a little

c650:
> > maybe we should introduce an Event type ? or PluginEvent?
> 
> can you explain a little

Yeah. Maybe we should have a type so that typescript enforces certain fields that we want in all of our structured logs.

c650 (approved):

PR #359, opened 2021-04-29
made dynamic embed list function in message.service.

CallumCoy:
Also, put it into the cat plugin (currently the only plugin that uses it).

### DIFF
--- a/src/app/plugins/cat.plugin.ts
+++ b/src/app/plugins/cat.plugin.ts
@@ -46,7 +46,7 @@ export class CatPlugin extends Plugin {
     if (args[0].includes('breed')) {
       //Simply return the list of supported breeds
       await this._getListEmbed();
-      message.reply(this._embedBreeds || 'Failed to load breeds.');
+      message.reply((await this._getListEmbed()) || 'Failed to load breeds.');
       return;
     }
 
@@ -79,15 +79,18 @@ export class CatPlugin extends Plugin {
 
   private async _getListEmbed() {
     if (this._embedBreeds) {
-      return;
+      return this._embedBreeds;
     }
 
     const breedsAsArray = this._breeds.map((breedData: { name: string; id: string }) => {
       return breedData.name;
     });
+
     this._embedBreeds = await this.container.messageService.generateEmbedList(breedsAsArray, {
       title: 'Breeds',
     });
+
+    return this._embedBreeds;
   }
 
   // gets the commands and puts spaces between all words
--- a/src/app/plugins/cat.plugin.ts
+++ b/src/app/plugins/cat.plugin.ts
@@ -1,7 +1,7 @@
 import Constants from '../../common/constants';
 import { Plugin } from '../../common/plugin';
 import { ChannelType, IContainer, IHttpResponse, IMessage } from '../../common/types';
-import { Message, MessageEmbed } from 'discord.js';
+import { MessageEmbed } from 'discord.js';
 
 class Breed {
   public name: string = '';
--- a/src/app/plugins/cat.plugin.ts
+++ b/src/app/plugins/cat.plugin.ts
@@ -38,7 +38,7 @@ export class CatPlugin extends Plugin {
         const breedsAsArray = this._breeds.map((breedData: { name: string; id: string }) => {
           return breedData.name;
         });
-        this._embedBreeds = container.messageService.generateEmbedList('Breeds', breedsAsArray);
+        this._embedBreeds = container.messageService.generateEmbedList(breedsAsArray, 'Breeds');
       })
       .catch((err) => this.container.loggerService.warn(err));
   }
--- a/src/app/plugins/cat.plugin.ts
+++ b/src/app/plugins/cat.plugin.ts
@@ -1,6 +1,6 @@
 import Constants from '../../common/constants';
 import { Plugin } from '../../common/plugin';
-import { ChannelType, IContainer, IHttpResponse, IMessage } from '../../common/types';
+import { ChannelType, IContainer, IHttpResponse, IMessage, Maybe } from '../../common/types';
 import { MessageEmbed } from 'discord.js';
 
 class Breed {
@@ -18,7 +18,7 @@ export class CatPlugin extends Plugin {
 
   private _API_URL: string = 'https://api.thecatapi.com/v1/';
   private _breeds: Breed[] = [];
-  private _embedBreeds = new MessageEmbed();
+  private _embedBreeds: Maybe<MessageEmbed>;
 
   constructor(public container: IContainer) {
     super();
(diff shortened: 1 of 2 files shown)


src/services/message.service.ts
```diff
@@ -77,6 +77,36 @@ export class MessageService {
...
+    const embedItem = new MessageEmbed();
+    embedItem.setColor('#0099ff').setTitle(title);
+
+    // Splits the list into numCols as evenly as possible.
+    const columns = new Array(numCols).fill(0).map((_) => listItems.splice(0, numRows));
+
+    // Cycles through each column inserting them, and also notes the alphabetic range
+    columns.forEach((column) => {
+      const fromLetter = column[0].charAt(0);
+      const toLetter = column[column.length - 1].charAt(0);
```
c650, line 98: careful w this

src/services/message.service.ts
```diff
@@ -77,6 +77,36 @@ export class MessageService {
...
+    const maxCol = 3;
+    const maxRows = 10; // This is a soft limit
+
+    // Setting up number of columns and rows for the list
+    const numCols = Math.min(maxCol, Math.ceil(listItems.length / maxRows));
+    const numRows = Math.ceil(listItems.length / numCols);
+
+    // Sets up the embed
+    const embedItem = new MessageEmbed();
+    embedItem.setColor('#0099ff').setTitle(title);
```
tanndlin, line 90: Dont set the title in this function to make it simpler. bc each embed will need a different title, its best to set it in the plugin

c650: perhaps color also configurable. maybe make options Object w optional params

CallumCoy: Easier to let the plugin do it all @c650 when the messageEmbed is returned (I tried to send in a optional messageEmbed but the foreach didn't like it).  So I'll just take out the title like tanndlin said and the plugin can deal with it.  Just thought it was common enough to stay in it.

c650: gotcha.

c650 (approved):

src/services/message.service.ts
CallumCoy: In the end, I made it a dynamic choice.

src/app/plugins/cat.plugin.ts
```diff
@@ -81,6 +77,16 @@ export class CatPlugin extends Plugin {
       .catch((err) => this.container.loggerService.warn(err));
   }
 
+  private async _getListEmbed() {
+    const breedsAsArray = this._breeds.map((breedData: { name: string; id: string }) => {
+      return breedData.name;
+    });
+    this._embedBreeds = await this.container.messageService.generateEmbedList(
+      breedsAsArray,
+      new MessageEmbed()
```
tanndlin, line 86: What's the point in passing in `new MessageEmbed()`?

src/services/message.service.ts
```diff
@@ -77,25 +77,30 @@ export class MessageService {
...
-    const embedItem = new MessageEmbed();
-    embedItem.setColor('#0099ff').setTitle(title);
+    let embedItem: MessageEmbed;
+
+    if (typeof arg === 'object') {
+      embedItem = arg;
+    } else {
+      embedItem = new MessageEmbed();
+      embedItem.setColor('#0099ff').setTitle(arg);
+    }
```
tanndlin, line 95: What's this?

src/app/plugins/cat.plugin.ts
```diff
@@ -18,7 +18,7 @@ export class CatPlugin extends Plugin {
 
   private _API_URL: string = 'https://api.thecatapi.com/v1/';
   private _breeds: Breed[] = [];
-  private _embedBreeds = new MessageEmbed();
+  private _embedBreeds: Maybe<MessageEmbed>;
```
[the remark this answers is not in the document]
CallumCoy: it's so, I can check if it needs to be generated or not later on in the code.

src/app/plugins/cat.plugin.ts
CallumCoy: oops mb that was from testing

src/app/plugins/cat.plugin.ts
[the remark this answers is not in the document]
tanndlin: oh, my mistake

src/services/message.service.ts
```diff
@@ -77,25 +77,30 @@ export class MessageService {
...
     // Setting up number of columns and rows for the list
     const numCols = Math.min(maxCol, Math.ceil(listItems.length / maxRows));
     const numRows = Math.ceil(listItems.length / numCols);
 
-    // Sets up the embed
-    const embedItem = new MessageEmbed();
-    embedItem.setColor('#0099ff').setTitle(title);
+    let embedItem: MessageEmbed;
+
+    if (typeof arg === 'object') {
```
c650, line 90: i dont like this

CallumCoy: the idea in general or the typeof command?

tanndlin:
Having the letters at the top assumes the list is sorted, but its not necessarily sorted, i feel that should be an option

src/services/message.service.ts
tanndlin: typeof, make an options object as optional not a multi type arg

src/app/plugins/cat.plugin.ts
```diff
@@ -50,7 +45,10 @@ export class CatPlugin extends Plugin {
 
     if (args[0].includes('breed')) {
       //Simply return the list of supported breeds
-      message.reply(this._embedBreeds);
+      if (!this._embedBreeds) {
+        await this._getListEmbed();
+      }
+      message.reply(this._embedBreeds || 'Failed to load breeds.');
```
c650, line 51: nit: do the `if` check in the `_getListEmbed` func so that the caller doesnt have to think about when to call it

c650 (approved):

src/services/message.service.ts
```diff
@@ -77,6 +77,51 @@ export class MessageService {
...
+
+    // Title was sent inserts it into the embed.
+    if (options.title) {
+      embedItem.setColor('#0099ff').setTitle(options.title);
+    }
+
+    // finds out if the list is sorted or not
+    const temp = listItems;
+    const sortedList = !!temp.reduce(
+      (prevRes: any, item: any) => prevRes !== false && item >= prevRes && item
```
tanndlin, line 99: Interesting

src/app/plugins/cat.plugin.ts
```diff
@@ -45,9 +45,7 @@ export class CatPlugin extends Plugin {
 
     if (args[0].includes('breed')) {
       //Simply return the list of supported breeds
-      if (!this._embedBreeds) {
-        await this._getListEmbed();
-      }
+      await this._getListEmbed();
```
tanndlin, line 48: I think corgo said this on the last pr, this should return the list, and just stick it in the `message.reply`

src/app/plugins/cat.plugin.ts
```diff
@@ -45,9 +45,7 @@ export class CatPlugin extends Plugin {
...
       message.reply(this._embedBreeds || 'Failed to load breeds.');
```
tanndlin, line 49: `await`

tanndlin (commented):
other than those, lgtm

src/app/plugins/cat.plugin.ts
CallumCoy: what do you mean await?

tanndlin: await the message reply

PR #361, opened 2021-04-29
Check ban list if resolveUser fails

### LINKED ISSUE #236: ModerationService use Client.fetch_user instead of Guild user lookup
https://github.com/joey-colon/lion/blob/b4b28e8790bf9c62b1c0f6d8b133450634a75dd3/src/services/moderation.service.ts#L13-L19

currently when a user is banned we can no longer look them up.

tanndlin:
fixes #236
We can look up banned users now

### DIFF
--- a/src/services/moderation.service.ts
+++ b/src/services/moderation.service.ts
@@ -386,15 +386,15 @@ export class ModService {
     username: string,
     channels: GuildChannel[]
   ): Promise<GuildChannel[]> {
-    const uid = await Moderation.Helpers.resolveUser(guild, username);
+    const id = await Moderation.Helpers.resolveUser(guild, username);
     const successfulBanChannelList: GuildChannel[] = [];
 
-    if (!uid) {
+    if (!id) {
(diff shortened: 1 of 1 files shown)


src/services/moderation.service.ts
```diff
@@ -383,7 +384,7 @@ export class ModService {
     username: string,
     channels: GuildChannel[]
   ): Promise<GuildChannel[]> {
-    const uid = Moderation.Helpers.resolveUser(guild, username);
+    const uid = await Moderation.Helpers.resolveUser(guild, username);
```
c650, line 387: either make them all `uid` or all `id` while ur at it.

c650 (approved):
lgtm w one small comment

src/services/moderation.service.ts
```diff
@@ -17,6 +17,8 @@ export namespace Moderation {
           return id;
         }
 
+        //If the lookup didnt work, they may be banned
```
c650, line 20: nit: spaces

c650 (approved):

Issue #360, opened 2021-04-29
Time stamps on the weather plugin


ColonelToad:
Lion needs to learn how to properly tell time and increment time, while it has been brought up before it's never been looked at. 
https://github.com/cs-discord-at-ucf/lion/commits/master/src/app/plugins/weather.plugin.ts

PR #363, opened 2021-04-29
time correction in weather plugin

CallumCoy:
go even easier on me :sob:

### DIFF
--- a/src/app/plugins/weather.plugin.ts
+++ b/src/app/plugins/weather.plugin.ts
@@ -104,7 +104,7 @@ export class WeatherPlugin extends Plugin {
 
       if (hours > 0) {
         const dt = new Date(timestamp * 1000);
-        let strf: string = dt.toLocaleTimeString([], { hour12: true, hour: 'numeric' });
+        const strf: string = dt.toLocaleTimeString([], { hour12: true, hour: 'numeric' });
 
         const weather_id = tdata.weather[0].id;
 
--- a/src/app/plugins/weather.plugin.ts
+++ b/src/app/plugins/weather.plugin.ts
@@ -104,8 +104,7 @@ export class WeatherPlugin extends Plugin {
 
       if (hours > 0) {
         const dt = new Date(timestamp * 1000);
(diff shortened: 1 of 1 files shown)


src/app/plugins/weather.plugin.ts
```diff
@@ -104,7 +104,7 @@ export class WeatherPlugin extends Plugin {
 
       if (hours > 0) {
         const dt = new Date(timestamp * 1000);
-        let strf: string = dt.getHours().toString();
+        let strf: string = (dt.getHours() % 12).toString();
```
tanndlin, line 107: Dont think this will work, this Gives a range from 0-11 o'clock

tanndlin: Make sure you test too

CallumCoy:
oops forgot, solves #360

tanndlin:
Not as simple as % 12, its other wise itll not change the AM/PM

CallumCoy:
> Not as simple as % 12, its other wise itll not change the AM/PM

we'll see in 20 minutes, can't really test it manually, but I think it will work looking at how it has posted in the past

src/app/plugins/weather.plugin.ts
```diff
@@ -104,7 +104,7 @@ export class WeatherPlugin extends Plugin {
 
       if (hours > 0) {
         const dt = new Date(timestamp * 1000);
-        let strf: string = dt.getHours().toString();
+        let strf: string = (dt.getHours() % 12 || 12).toString();
         strf += Math.floor(dt.getHours() / 12) < 1 ? 'AM' : 'PM';
```
c650, line 108: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString

CallumCoy:
> can we make `strf` const and better name?

Was const, the lint picked it up first :eyes:. as for the name sure, don't even know what it is meant to stand for.

tanndlin (approved):

PR #378, opened 2021-05-06
Notfiy user when warned and DMs off

tanndlin:
Opening this for now, thinking about alternate ways to notify a user when they have been warned

### DIFF
--- a/src/services/moderation.service.ts
+++ b/src/services/moderation.service.ts
@@ -232,14 +232,12 @@ export class ModService {
     });
 
     const wasAbleToDM = await this._sendModMessageToUser('A warning has been issued. ', report);
+    const response = `User warned: ${Moderation.Helpers.serialiseReportForMessage(report)}`;
     if (wasAbleToDM) {
-      return `User warned: ${Moderation.Helpers.serialiseReportForMessage(report)}`;
+      return response;
     }
 
-    return (
-      `User warned: ${Moderation.Helpers.serialiseReportForMessage(report)}\n` +
-      `User's DM are off, and was not notified of warning`
-    );
+    return response + `\nUser's DM are off, and was not notified of warning`;
   }
 
   public async fileBan(report: Moderation.Report) {
@@ -577,7 +575,7 @@ export class ModService {
         });
       return true;
     } catch (e) {
-      this._loggerService.warn(`Couldnt warn ${rep.user} about warn. ${e}`);
+      this._loggerService.warn(`Couldnt DM ${rep.user} about warn. ${e}`);
       return false;
     }
   }
--- a/src/services/moderation.service.ts
+++ b/src/services/moderation.service.ts
@@ -1,7 +1,6 @@
 import { Guild, Snowflake, MessageEmbed, GuildChannel, TextChannel, User } from 'discord.js';
 import { StorageService } from './storage.service';
 import { ObjectId } from 'mongodb';
-import Environment from '../environment';
 import { ClientService } from './client.service';
 import { GuildService } from './guild.service';
 import { LoggerService } from './logger.service';
--- a/src/services/moderation.service.ts
+++ b/src/services/moderation.service.ts
@@ -231,8 +231,15 @@ export class ModService {
       reportId: fileReportResult,
     });
 
-    await this._sendModMessageToUser('A warning has been issued. ', report);
-    return `User warned: ${Moderation.Helpers.serialiseReportForMessage(report)}`;
+    const wasAbleToDM = await this._sendModMessageToUser('A warning has been issued. ', report);
+    if (wasAbleToDM) {
+      return `User warned: ${Moderation.Helpers.serialiseReportForMessage(report)}`;
+    }
+
+    return (
+      `User warned: ${Moderation.Helpers.serialiseReportForMessage(report)}\n` +
+      `Was not user's DM are off, and was not notified of warning`
+    );
   }
 
   public async fileBan(report: Moderation.Report) {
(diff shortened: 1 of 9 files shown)


src/services/moderation.service.ts
```diff
@@ -561,12 +568,17 @@ export class ModService {
...
+  private async _sendModMessageToUser(message: string, rep: Moderation.Report): Promise<boolean> {
+    try {
+      await this._clientService.users.cache
+        .get(rep.user)
+        ?.send(`${message} Reason: ${rep.description || '<none>'}`, {
+          files: rep.attachments && JSON.parse(JSON.stringify(rep.attachments)),
+        });
+      return true;
+    } catch (e) {
+      this._loggerService.warn(`Couldnt warn ${rep.user} about warn. ${e}`);
```
c650, line 580: "warn about warn"

src/services/moderation.service.ts
```diff
@@ -231,8 +231,15 @@ export class ModService {
...
 
-    await this._sendModMessageToUser('A warning has been issued. ', report);
-    return `User warned: ${Moderation.Helpers.serialiseReportForMessage(report)}`;
+    const wasAbleToDM = await this._sendModMessageToUser('A warning has been issued. ', report);
+    if (wasAbleToDM) {
+      return `User warned: ${Moderation.Helpers.serialiseReportForMessage(report)}`;
+    }
+
+    return (
+      `User warned: ${Moderation.Helpers.serialiseReportForMessage(report)}\n` +
```
c650, line 240: same string twice, save in `response` variable or something?

c650 (approved):

c650 (commented):
so if bot restarts before someone ack;s a warning they may not?

tanndlin:
> so if bot restarts before someone ack;s a warning they may not?

The Listeners only last for a day, and if the bot resets the listeners go away bc its in temp memory
I could make it so that it recreates the listeners on restart

Currently there is no punishment for not acknowledging, do you want me to give the `@Suspended` role until they acknowledge

VijayStroup:
What I do is check to see if the reaction was added in the category.
Hence, you can have a category called "warnings" and have a bunch of channels like "465026352777789441_warn_1" and "392203144085504011_warn_3" and if a reaction gets added in the "warnings" category, check to see if the id of the user who added it matches the id in the channel name, if so, they acknowledged the warning.

This method requires no cache.

tanndlin:
> What I do is check to see if the reaction was added in the category.
> Hence, you can have a category called "warnings" and have a bunch of channels like "465026352777789441_warn_1" and "392203144085504011_warn_3" and if a reaction gets added in the "warnings" category, check to see if the id of the user who added it matches the id in the channel name, if so, they acknowledged the warning.
> 
> This method requires no cache.

I see what you are saying, but the message needs to be in the cache to listen for reactions the normal way
Which isnt too bad of an issue, you can manually cache them

c650:
> Currently there is no punishment for not acknowledging, do you want me to give the @Suspended role until they acknowledge

i thought that was the point

c650:
> Not sure if there is anything like that in discord.js

The closest thing is listening to the react event but it sounds like it only notices the reacts on cached messages.

tanndlin:
> > Not sure if there is anything like that in discord.js
> 
> 
> 
> The closest thing is listening to the react event but it sounds like it only notices the reacts on cached messages.

It's not hard to make a job that ensures the messages are cached

tanndlin:
> > Currently there is no punishment for not acknowledging, do you want me to give the @Suspended role until they acknowledge
> 
> 
> 
> i thought that was the point 

Lol ok

VijayStroup:
> The closest thing is listening to the react event but it sounds like it only notices the reacts on cached messages.

discord.js

src/services/warning.service.ts
```diff
@@ -0,0 +1,91 @@
...
+    //Give user Supsended Role until they acknowledge
+    await member.roles.add(this._guildService.getRole('Suspended'));
+  }
+
+  private async _getChanForUser(rep: Moderation.Report, warnCat: CategoryChannel) {
+    if (this._chanMap.has(rep.user)) {
+      return this._chanMap.get(rep.user) as GuildChannel;
+    }
+
+    return this._guildService.get().channels.create(rep.user, {
```
c650, line 48: probably also let Moderators see the chan

src/services/warning.service.ts
```diff
@@ -0,0 +1,91 @@
...
+    } else {
+      //If the bot restated, it wont be in the map
+      await this._guildService
+        .get()
+        .channels.cache.filter((c) => c.name === id)
+        .first()
+        ?.delete('User acknowledged warning');
+    }
+
+    await this._guildService
```
c650, line 86: try to remove the role before we delete the channel, so that the user could @ a mod if it doesnt work

src/services/warning.service.ts
```diff
@@ -0,0 +1,91 @@
...
+    const chan = this._chanMap.get(id);
+    if (chan) {
+      await chan.delete('User acknowledged warning');
+      this._chanMap.delete(id);
+    } else {
+      //If the bot restated, it wont be in the map
+      await this._guildService
+        .get()
+        .channels.cache.filter((c) => c.name === id)
+        .first()
```
c650, line 82: prefer

```
if (!chan) {
 chan = .. // get channel from guild service
}

await chan.delete(...)
```

src/services/warning.service.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (!member) {
+      return;
+    }
+
+    const warnChan = await this._getChanForUser(rep, this._warnCategory);
+    this._chanMap.set(rep.user, warnChan);
+
+    await (warnChan as TextChannel).send(member.toString());
+    const embed = await (warnChan as TextChannel).send(this._serializeToEmbed(message, rep));
+    await embed.react('👍');
```
c650, line 37: nit: hardcoded emoji

PR #382, opened 2021-05-13
more role reactions

VijayStroup:


c650 (commented):
Lot of copy pasta here. Make a clear, reusable way, like a function

src/app/plugins/addroles.plugin.ts
```diff
@@ -16,6 +16,15 @@ export class AddRolesPlugin extends Plugin {
     return !!args.length;
   }
 
+  react(emojiName: string, message: IMessage) {
```
c650, line 19: private

src/app/plugins/addroles.plugin.ts
```diff
@@ -33,22 +42,8 @@ export class AddRolesPlugin extends Plugin {
...
-        else if (role.name.toLowerCase() === 'gradstudent') {
-          const knightEmoji = this.container.guildService.get().emojis.cache
-            .filter(n => n.name.toLowerCase() === 'knight').first();
-  
-          if (knightEmoji) {
-            message.react(knightEmoji);
-          }
-        }
+        if (role.name.toLowerCase() === 'alumni') this.react('okboomer', message);
+        else if (role.name.toLowerCase() === 'gradstudent') this.react('knight', message);
```
c650, line 46: handle the `if` logic in the `react` func instead of out here

tanndlin: Cache the emojis as a property as like a record so you have
```
private emojis: Record<string, IRoleEmoji> = {
    alumni: { emojiName: 'knight', emoji: null },
    gradstudent: { emojiName: 'okboomer', emoji: null },
};
```
and assign them in the constructor that way we dont have to find it, and make sure it exists each time

then you can do something like
```
await Promise.all(
  Object.keys(this.emojis).map((key) => {
   if (role.toLowerCase() !== key) {
      return;
   }

   return message.react(this.emojis[key].emojiName);
  })
);
```
That way its easily scalable

Sorry if that not correct

tanndlin: also make sure you are awaiting promises

tanndlin: for that example `IRoleEmoji` would be a simple interface with a string for emojiName and ReactionEmoji, which is the actual emoji

c650: This tbh

src/app/plugins/addroles.plugin.ts
```diff
@@ -53,3 +64,8 @@ export class AddRolesPlugin extends Plugin {
     }
   }
 }
+
+interface IRoleEmoji{
+  emojiName: string,
+  emoji: undefined | GuildEmoji
```
tanndlin, line 70: use `Maybe<GuildEmoji>` instead

src/app/plugins/addroles.plugin.ts
```diff
@@ -32,14 +50,7 @@ export class AddRolesPlugin extends Plugin {
...
         roles_added.push(role.name);
-
-        if (role.name.toLowerCase() === 'alumni') {
-          const knightEmoji = this.container.guildService.get().emojis.cache
-            .filter(n => n.name === 'knight').first();
-          if (knightEmoji) {
-            message.react(knightEmoji);
-          }
-        }
+        this.react(role.name.toLowerCase(), message);
```
tanndlin, line 53: await this

src/app/plugins/addroles.plugin.ts
```diff
@@ -16,6 +17,23 @@ export class AddRolesPlugin extends Plugin {
...
+
+  private react(role: string, message: IMessage) {
+    // check to see if emoji has been instantiated
+    if (!this.emojis[role].emoji) {
+      this.emojis[role].emoji = this.container.guildService.get().emojis.cache
+        .filter(n => n.name.toLowerCase() === this.emojis[role].emojiName).first();
+    }
+
+    if (this.emojis[role].emoji) {
+      message.react(this.emojis[role].emoji as EmojiIdentifierResolvable)
```
tanndlin, line 33: await this

tanndlin (commented):
Does this work for roles that aren't `almuni` or `GradStudent`?

src/app/plugins/addroles.plugin.ts
tanndlin: nevermind, instead just make it `GuildEmoji` then cast GuildEmoji when you [retrieve it](https://github.com/VijayStroup/lion/blob/13fc2364601d1bbd470c34039f8be30ed6a25cd3/src/app/plugins/addroles.plugin.ts#L29)

VijayStroup:
> Does this work for roles that aren't `almuni` or `GradStudent`?
Yup, just add more records to `emojis` with the same format.

src/app/plugins/addroles.plugin.ts
```diff
@@ -16,6 +17,23 @@ export class AddRolesPlugin extends Plugin {
     return !!args.length;
   }
 
+  private emojis: Record<string, IRoleEmoji> = {
+    alumni: { emojiName: 'okboomer', emoji: undefined },
+    gradstudent: { emojiName: 'knight', emoji: undefined },
+  }
```
tanndlin, line 23: Move this between public stuff and constructor

src/app/plugins/addroles.plugin.ts
```diff
@@ -16,6 +17,23 @@ export class AddRolesPlugin extends Plugin {
...
+
+  private async react(role: string, message: IMessage) {
+    // check to see if emoji has been instantiated
+    if (!this.emojis[role].emoji) {
+      this.emojis[role].emoji = this.container.guildService.get().emojis.cache
+        .filter(n => n.name.toLowerCase() === this.emojis[role].emojiName).first();
+    }
+
+    if (this.emojis[role].emoji) {
+      await message.react(this.emojis[role].emoji as EmojiIdentifierResolvable)
```
tanndlin, line 33: remove this cast and put `...first() as GuildEmoji` on line 29

tanndlin (approved):

PR #401, opened 2021-05-18
Ariana Grande

VijayStroup:


src/app/plugins/ariana.plugin.ts
```diff
@@ -0,0 +1,79 @@
+import { Plugin } from '../../common/plugin';
+import { IContainer, IMessage, ChannelType, IHttpResponse } from '../../common/types';
+
+export class ArianaPlugin extends Plugin {
+  public name: string = 'Ariana Plugin';
+  public description: string = 'Get a random picture of Ariana Grande.';
+  public usage: string = 'ariana';
+  public pluginAlias = ['ari'];
+  public permission: ChannelType = ChannelType.Bot;
```
CallumCoy, line 9: I think channel Media, or just type Public. Is better suited.

src/app/plugins/ariana.plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+      .get(this._API_URL + '1')
+      .then((res: IHttpResponse) => {
+        const data: string = res.data;
+        const pagesSearch: RegExpMatchArray | null = data.match(this.rePages);
+
+        if (!pagesSearch) {
+          this.container.loggerService.warn('Failed to get Ariana page in constructor.');
+          return;
+        }
+        const pageSearch: RegExpMatchArray | null = pagesSearch[0].match(this.rePage);
```
CallumCoy, line 32: use type Maybe from types.ts for lines 32, 36, and 46.
https://github.com/cs-discord-at-ucf/lion/blob/master/src/common/types.ts < line 176

src/app/plugins/ariana.plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+          return;
+        }
+      })
+      .catch((err) => this.container.loggerService.warn(err));
+
+    return ari;
+  }
+
+  public async execute(message: IMessage, args?: string[]) {
+    const page = Math.floor(Math.random() * this.nPages) + 1;
```
CallumCoy, line 67: Can you just comment that page 0 is invalid on the API.  just so it's noted why the +1 is there please.

src/app/plugins/ariana.plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+
+        if (!imgSearch) {
+          return;
+        }
+
+        const count: number = Math.floor(Math.random() * imgSearch.length);
+
+        const url = imgSearch[count].match(this.reThumb);
+        if (url) {
+          ari = url[0].replace(/amp;|"/g, '');
```
CallumCoy, line 56: reverse and remove the else.

` if (!url) {
   return
}`

src/app/plugins/ariana.plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+  public async execute(message: IMessage, args?: string[]) {
+    const page = Math.floor(Math.random() * this.nPages) + 1;
+    const ari = await this.getAri(page);
+    if (ari) {
+      await message.reply('', {
+        files: [{
+          attachment: ari,
+          name: 'ari.jpg'
+        }]});
+      } else {
```
CallumCoy, line 75: Reverse and return, then remove the else.

src/app/plugins/ariana.plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+        }
+        ari = url[0].replace(/amp;|"/g, '');
+      })
+      .catch((err) => this.container.loggerService.warn(err));
+
+    return ari;
+  }
+
+  public async execute(message: IMessage, args?: string[]) {
+    const page = Math.floor(Math.random() * this.nPages) + 1; // valid pages 1-100
```
joey-colon, line 66: is there a reason why we expose `page` in execute method? this seems like an implementation detail that only `getAri` cares about

src/app/plugins/ariana.plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+        const pageSearch: Maybe<RegExpMatchArray> = pagesSearch[0].match(this.rePage);
+        if (pageSearch) {
+          this.nPages = parseInt(pageSearch[0].replace(/>|</g, ''));
+        }
+      })
+      .catch((err) => this.container.loggerService.warn(err));
+  }
+
+  private async getAri(page: number) {
+    let ari;
```
joey-colon, line 41: im not really a fan of this, but if u feel strongly about it, u can keep it as is.

an alternative way (and more readble imo) could be;

```
const response = await this.container.httpService.get(this._API_URL + page).then(res => res.data);

// all logic within your `then` function can be moved outside since you can directly interact with `response`.
```
 this will get rid of the `ari` variable

src/app/plugins/ariana.plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+  public description: string = 'Get a random picture of Ariana Grande.';
+  public usage: string = 'ariana';
+  public pluginAlias = ['ari'];
+  public permission: ChannelType = ChannelType.Public;
+
+  private _API_URL: string = 'https://www.gettyimages.com/photos/ariana-grande?page=';
+  private reImgs: RegExp = new RegExp(/<img class="gallery-asset__thumb gallery-mosaic-asset__thumb"(.*?)>/g);
+  private reThumb: RegExp = new RegExp(/https:\/\/(.*?)"/g);
+  private rePages: RegExp = new RegExp(/<span class="PaginationRow-module__lastPage___2pChH">(.*?)<\/span>/g);
+  private rePage: RegExp = new RegExp(/>(.*?)</g);
```
joey-colon, line 15: if someone ever wants to contribute to this, it might be difficult given these variable names. could we use more descriptive variable names or add a comment for what each pattern is referring to?

src/app/plugins/ariana.plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+  private nPages: number = 0;
```
joey-colon, line 16: np: `totalPages` reads better

src/app/plugins/ariana.plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  constructor(public container: IContainer) {
+    super();
+
+    // get total number of pages of Ariana
+    this.container.httpService
+      .get(this._API_URL + '1')
+      .then((res: IHttpResponse) => {
+        const data: string = res.data;
+        const pagesSearch: Maybe<RegExpMatchArray> = data.match(this.rePages);
```
joey-colon, line 26: fyi, with array destructuring, we can do something along the following lines:

```
const array = [1,2,3]
const [firstElement] = array; // 1
```

we can translate the same idea to `pagesSearch`

src/app/plugins/ariana.plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+
+        if (!pagesSearch) {
+          this.container.loggerService.warn('Failed to get Ariana page in constructor.');
+          return;
+        }
+        const pageSearch: Maybe<RegExpMatchArray> = pagesSearch[0].match(this.rePage);
```
joey-colon, line 32: similar to the variable names comment above. its sort of confusing to make meaning of the difference as a reviewer between `pageSearch` and `pagesSearch`

src/app/plugins/ariana.plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+          return;
+        }
+
+        const count: number = Math.floor(Math.random() * imgSearch.length);
+
+        const url = imgSearch[count].match(this.reThumb);
+        if (!url) {
+          return;
+        }
+        ari = url[0].replace(/amp;|"/g, '');
```
joey-colon, line 58: re: array restructuring comment above

src/app/plugins/ariana.plugin.ts
VijayStroup: On every execute, it grabs a random page and a random image from that page.
The selection for the random page can be in the `getAri` function if you'd like.

tanndlin:
This whole thing seems kinda odd/creepy?
Can we make it a celebrity plugin? Then you can look up Ariana grande all you want

VijayStroup:
> This whole thing seems kinda odd/creepy?
> Can we make it a celebrity plugin? Then you can look up Ariana grande all you want

I do not know how to make it so open, it would still have to be restricted to a list of celerity names so you can't search up any picture you want.

tanndlin:
> > This whole thing seems kinda odd/creepy?
> 
> > Can we make it a celebrity plugin? Then you can look up Ariana grande all you want
> 
> 
> 
> I do not know how to make it so open, it would still have to be restricted to a list of celerity names so you can't search up any picture you want.

Well yeah, is there an API for￼ celebs in general?

c650:
this gave me a chuckle

PR #403, opened 2021-05-20
Implemented fix for no classes.

### LINKED ISSUE #397: Bot may crash on servers with no classes
Bot may crash on servers with no classes.  Register and unregister reacts sends a invalid class to the functions it provides.

CallumCoy:
fixes #397

### DIFF
--- a/src/app/plugins/register.plugin.ts
+++ b/src/app/plugins/register.plugin.ts
@@ -1,7 +1,6 @@
 import { User } from 'discord.js';
 import { Plugin } from '../../common/plugin';
 import { IContainer, IMessage, ChannelType, IEmbedData, ClassType } from '../../common/types';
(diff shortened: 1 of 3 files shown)


src/app/plugins/register.plugin.ts
```diff
@@ -110,6 +110,12 @@ export class RegisterPlugin extends Plugin {
       invalidClasses
     );
 
+    // Emoji data will be empty if the server has no classes.
+    if (!embedData.emojiData.length) {
+      message.reply('No classes found at this time.');
+      return;
+    }
+
```
tanndlin, line 118: This isnt obvious why this is true, with the way this is setup
I feel like it would be more obvious if you did `classService.getClasses(ClassType.ALL)` and check if 0

tanndlin (approved):

PR #416, opened 2021-05-25
Prom atlas/crumbl plugin

PromAtlas:


src/app/plugins/crumbl.plugin.ts
```diff
@@ -0,0 +1,134 @@
...
+import { MessageEmbed } from 'discord.js';
+
+export class CrumblPlugin extends Plugin {
+  public name: string = 'Crumbl Cookies Plugin';
+  public description: string = 'Returns the cookies available that week at Crumbl';
+  public usage: string = 'crumbl';
+  public pluginAlias = [];
+  public permission: ChannelType = ChannelType.Public;
+  public pluginChannelName: string = Constants.Channels.Public.Food;
+  public buildId: string = '';
```
tanndlin, line 15: Should this be `private`?

src/app/plugins/crumbl.plugin.ts
```diff
@@ -0,0 +1,134 @@
...
+  private _lastPost: Maybe<IMessage> = null;
```
tanndlin, line 16: nice

src/app/plugins/crumbl.plugin.ts
```diff
@@ -0,0 +1,134 @@
...
+          const embed = new MessageEmbed()
+          .setTitle(cookie.name)
+          .setDescription(cookie.description)
+          .setThumbnail(cookie.image);
+          return embed;
+        });
+  }
+
+  public async execute(message: IMessage, args: string[]) {
+    const regex: RegExp = /"buildId":"(.*?)"/;
```
tanndlin, line 50: make this a class constant

src/app/plugins/crumbl.plugin.ts
```diff
@@ -0,0 +1,134 @@
...
+  {
+    const matches = data.match(regex);
+
+    if (matches !== null && matches.length > 1)
+    {
+      return matches[1];
+    }
+    
+    return '';
+  }
```
tanndlin, line 32: There is some magic going on here, maybe do some array destructuring, or at least some comments
Also, we prefer to check the negative and return
```
if(!matches.length){
    return '';
}
//Do stuff
```

src/app/plugins/crumbl.plugin.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  private _createEmbed(cookies: any, message: IMessage): MessageEmbed[]
+  {
+    return  cookies
+      .filter (
+        (c: any) => c.name !== 'Milk Chocolate Chip' && c.name !== 'Chilled Sugar'
+      )
```
tanndlin, line 39: Why not these 2?

tanndlin: also we prefer not to use any when you can, so make an interface at the bottom of the plugin for cookies parameter

src/app/plugins/crumbl.plugin.ts
```diff
@@ -0,0 +1,134 @@
...
+      .map((cookie: any, i: number) => {
+          const embed = new MessageEmbed()
+          .setTitle(cookie.name)
```
tanndlin, line 42: just return the new embed rather than storing in a new variable

src/app/plugins/crumbl.plugin.ts
```diff
@@ -0,0 +1,134 @@
...
+        this.buildId = this._matchId(data, regex); 
+      })
+      .catch((err) => this.container.loggerService.warn(err));      
+
+    await this.container.httpService
+      .get(`https://crumblcookies.com/_next/data/${this.buildId}/index.json`)
+        .then(async (blob: IHttpResponse) => {
+          if (this.buildId === '')
+          {
+            console.log('The build id is empty');
```
tanndlin, line 66: prefer `this.container.loggerService.warn(...)` if its useful to know.
Also when we catch errors, its best to let the user know something happened, at least give them a response

src/app/plugins/crumbl.plugin.ts
```diff
@@ -0,0 +1,134 @@
...
+      const last = batch.last();
+      if (last) {
+        last_id = last.id;
+      }
+
+      buffer.push(...batch.array());
+    }
+    return buffer;
+  }
+}
```
tanndlin, line 134: kinda unnecessary
I wrote this to get > 100 messages at once when looking for listings in the channel, but just for deleting old messages is over kill.
Just remove this function and use
```
const messages = await message.channel.messages.fetch({limit:100});
```

src/app/plugins/crumbl.plugin.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    await this._fetchMessages(listCall, 100).then((messages) => {
+      const botMsgs = messages.filter(
+        (msg) =>
+          msg.author.bot && //From bot
+          msg.embeds.length && //Contains an embed
+          newPosting.id != msg.id //Not the new listing
+      );
+      if (botMsgs.length === 0) {
```
tanndlin, line 95: 0 is false, so you can just do `if(!botMsgs.length)`

src/app/plugins/crumbl.plugin.ts
```diff
@@ -0,0 +1,134 @@
...
+        return;
+      }
+
+      this._lastPost = botMsgs[0];
```
tanndlin, line 99: use array destructuring
`[this._lastPost] = botMsgs;`

tanndlin (commented):
Looks pretty good, just make sure you have prettier installed and make sure to format it with prettier, some of the formatting looks odd

tanndlin:
Looks pretty dang good, and I like the use of `messageService.sendPagedEmbed` :)

src/app/plugins/crumbl.plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+    const matches = data.match(this._regex);
+    if (!matches) {
+      return '';
+    }
+
+    const [, id] = matches;
+    return id;
+  }
+
+  private _createEmbed(cookies: any, message: IMessage): MessageEmbed[] {
```
c650, line 35: why `any`?

src/app/plugins/crumbl.plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    const messages = await listCall.channel.messages.fetch({ limit: 100 });
+    const botMsgs = messages.filter(
+      (msg) =>
+        msg.author.bot && //From bot
+        msg.embeds.length !== 0 && //Contains an embed
+        newPosting.id != msg.id //Not the new listing
+    );
+    if (!botMsgs.size) {
```
c650, line 90: newline above

Issue #418, opened 2021-05-25
Register plugin getting error with space in input


tanndlin:
`!register eel3552_wei  cnt4714_llewellyn` causes an error with no reply
No specific error given either
```
error {"status":"error","pluginName":"Register Plugin","args":["eel3552_wei","","cnt4714_llewellyn"],"user":"Tanndlin#4450","error":{}} 
```
Works with each class individually, but can't recreate error with any other combo of classes

tanndlin:
Bruh, there was a space in the input 😂
Still shouldnt give an error when there is a space in the input

tanndlin:
Fixed by not allowing space as args
#419

PR #419, opened 2021-05-25
Sanitize input

tanndlin:


joey-colon (approved):

tanndlin:
refs #418
We shouldnt be getting errors, so we need to handle them
This doesnt solve #418

PR #431, opened 2021-05-29
Prom atlas/lion presence

PromAtlas:


tanndlin (approved):

src/app/plugins/lionpresence.plugin.ts
```diff
@@ -0,0 +1,36 @@
...
+import { Plugin } from '../../common/plugin';
+import { IContainer, IMessage, ChannelType } from '../../common/types';
+
+export class LionPresence extends Plugin {
+  public name: string = 'Lion Presence';
+  public description: string = 'Plugin to set the presence of the lion bot.';
+  public usage: string = 'activity <activity_type> <message>';
+  public pluginAlias = ['setactivity', 'setact'];
+  public permission: ChannelType = ChannelType.Staff;
+  private _types: string[] = ['PLAYING', 'STREAMING', 'LISTENING', 'WATCHING', 'COMPETING'];
```
joey-colon, line 11: looks like this should have type ActivityType[]

tanndlin: ahhh, we were trying to find a way to check if a string followed that custom type, and couldn't think of a good way

joey-colon: ahh gotcha

PR #435, opened 2021-05-29
Tylermathis/pluginsearch

TylerMathis:
Allow a user to execute a [greplike](https://man7.org/linux/man-pages/man1/grep.1.html) search on all user-level plugins.

Does not allow queries on Staff/Admin plugins.

image
image

### DIFF
--- a/src/app/plugins/commandsearch.plugin.ts
+++ b/src/app/plugins/commandsearch.plugin.ts
@@ -1,11 +1,11 @@
 import { Plugin } from '../../common/plugin';
 import { ChannelType, IContainer, IMessage, IPlugin } from '../../common/types';
 
-export class PluginSearchPlugin extends Plugin {
-  public name: string = 'Plugin Search';
-  public description: string = 'Search our plugins database';
-  public usage: string = 'search <phrase>';
-  public pluginAlias = ['search', 'searchplugin', 'searchplugins'];
+export class CommandSearchPlugin extends Plugin {
+  public name: string = 'Command Search';
+  public description: string = 'Search our commands on a key word or phrase';
+  public usage: string = 'commands <word or phrase>';
+  public pluginAlias = ['searchcommands', 'grep'];
   public permission: ChannelType = ChannelType.Bot;
   public commandPattern: RegExp = /[^]+/;
 
--- a/src/app/plugins/commandsearch.plugin.ts
+++ b/src/app/plugins/commandsearch.plugin.ts
@@ -47,13 +47,13 @@ export class CommandSearchPlugin extends Plugin {
 
   private _grep(plugin: IPlugin, query: string): boolean {
     const pluginMeta = `
-      ${plugin.name}
-      ${plugin.description}
-      ${plugin.usage}
-      ${plugin.pluginAlias?.join(' ')}`.toLowerCase();
+      name: ${plugin.name}
+      description: ${plugin.description}
+      usage: ${plugin.usage}
+      aliases: ${plugin.pluginAlias?.join(' ')}`;
 
     // Find greplike matches within plugin's metadata.
-    const grepRegex = new RegExp(`^.*(${query.toLowerCase()}).*$`, 'mg');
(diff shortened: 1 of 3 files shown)


src/app/plugins/pluginsearch.plugin.ts
```diff
@@ -0,0 +1,55 @@
...
+  constructor(public container: IContainer) {
+    super();
+  }
+
+  public async execute(message: IMessage, args: string[]) {
+    const query = args.join(' ');
+
+    // For every plugin, evaluate it's match
+    const results: string[] = [];
+    Object.entries(this.container.pluginService.plugins).forEach(([pluginName, plugin]) => {
```
TylerMathis, line 21: If somebody has a more jslike way of populating this results array please chime in. Maybe a reduce? I couldn't get the syntax out.

joey-colon: i wouldn't sweat it but something like
```
Object.entries(this.container.pluginService.plugins).reduce((results, [pluginName, plugin]) => {
  if (
     plugin.permission === ChannelType.Admin || 
     plugin.permission === ChannelType.Staff ||
     !this._grep(plugin, query)
   ) {
          return results;
  }

  results.push(pluginName);
  return results;
}, []);
```

TylerMathis: Sick! Thank you :D

tanndlin:
Maybe change the name to `!commands`

TylerMathis:
The name `commands` feels like it might be reserved for something else later on, so I just threw the word search on top :+1:

src/app/plugins/commandsearch.plugin.ts
```diff
@@ -0,0 +1,61 @@
...
+  private _grep(plugin: IPlugin, query: string): boolean {
+    const pluginMeta = `
+      ${plugin.name}
+      ${plugin.description}
+      ${plugin.usage}
+      ${plugin.pluginAlias?.join(' ')}`.toLowerCase();
+
+    // Find greplike matches within plugin's metadata.
+    const grepRegex = new RegExp(`^.*(${query.toLowerCase()}).*$`, 'mg');
+    const match = pluginMeta.match(grepRegex);
```
c650, line 57: no need to build a regex for this search

src/app/plugins/commandsearch.plugin.ts
```diff
@@ -0,0 +1,61 @@
...
+    const query = args.join(' ');
+
+    // For every plugin, evaluate it's match
+    const results = Object.entries(this.container.pluginService.plugins).reduce(
+      (results: string[], [pluginName, plugin]) => {
+        if (
+          plugin.permission === ChannelType.Admin ||
+          plugin.permission === ChannelType.Staff ||
+          !this._grep(plugin, query)
+        ) {
```
c650, line 26: this doesnt quite make it so that the response always only returns commands that are available in the calling channel

src/app/plugins/commandsearch.plugin.ts
tanndlin: I think the idea was that a user could make their own regex as input too
What would be nice, is if when you build the pluginMeta, you did something like ` \`name: ${plugin.name}\`` and so on,
that way each element is query-able with regex @TylerMathis

src/app/plugins/commandsearch.plugin.ts
tanndlin: That was a different PR, this command is only in #bot_commands

TylerMathis: Yeah, we want users to be able to query all commands.

src/app/plugins/commandsearch.plugin.ts
TylerMathis: It would be cool to support custom regex, why not? :D

tanndlin (approved):
<3

tanndlin:
need to rebase

PR #437, opened 2021-05-30
tylermathis/ai/ttt

TylerMathis:
If a user challenges Lion to TicTacToe, Lion will now play back. Difficulty is all the way up, the user will always either tie or lose ( but we don't have to tell them that ;) )

### DIFF
--- a/src/app/plugins/tictactoe.plugin.ts
+++ b/src/app/plugins/tictactoe.plugin.ts
@@ -198,15 +198,16 @@ class TTTGame {
       })
     );
 
-    // Sort the moves
-    moves.sort((a, b) => a.val - b.val);
+    // Sort the moves in decreasing value
+    moves.sort((a, b) => b.val - a.val);
     return { bestRow: moves[0].row, bestCol: moves[0].col };
   }
 
   // Recursive algorithm to find the strength of each possible move
   private _evaluate(currentPlayer: number) {
+    // If we reached a win state, the LAST move won.
     if (this._checkWin()) {
-      return currentPlayer;
+      return -currentPlayer;
     }
 
     if (this._checkTie()) {
--- a/src/app/plugins/tictactoe.plugin.ts
+++ b/src/app/plugins/tictactoe.plugin.ts
@@ -46,7 +46,7 @@ export class TicTacToe extends Plugin {
     const game = new TTTGame(
       message.author,
       oppMember.user,
-      message.mentions.members?.first()?.roles.cache.some((role) => role.name === 'Chatbot')
+      oppMember.id === this.container.clientService.user?.id
     );
     const msg = await message.reply(game.showBoard());
     await Promise.all(this._moves.map((emoji) => msg.react(emoji)));
@@ -113,10 +113,10 @@ class TTTGame {
   // -1 is playerA
   private currentPlayer: number = -1;
 
-  constructor(playerA: User, playerB: User, playingLion: Maybe<boolean>) {
+  constructor(playerA: User, playerB: User, playingLion: boolean) {
     this._playerA = playerA;
     this._playerB = playerB;
-    this._playingLion = playingLion ? true : false;
+    this._playingLion = playingLion;
 
     // Make 3x3 board of 0
     this._board = [
@@ -163,8 +163,8 @@ class TTTGame {
     this._flipTurn();
     this.reset();
     await msg.edit(this.showBoard());
-    // Reset where the player is choosing
 
+    // Make Lion's move if necessary.
     if (!this._gameOver && this.currentPlayer === 1 && this._playingLion) {
       this._lionMove();
       this._checkAndUpdateWin();
@@ -225,6 +225,7 @@ class TTTGame {
       })
     );
 
+    // Return average value of all possible moves from position.
     return this._sumArray(moves) / moves.length;
   }
 
--- a/src/app/plugins/tictactoe.plugin.ts
+++ b/src/app/plugins/tictactoe.plugin.ts
@@ -192,7 +192,7 @@ class TTTGame {
 
         // Make the move and evaluate the board state
         this._board[row][col] = this.currentPlayer;
-        moves.push({ row, col, val: this._minimax(this.currentPlayer * -1) });
+        moves.push({ row, col, val: this._evaluate(this.currentPlayer * -1) });
         // Backtrack.
         this._board[row][col] = 0;
       })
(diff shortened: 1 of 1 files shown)


src/app/plugins/tictactoe.plugin.ts
```diff
@@ -43,7 +43,11 @@ export class TicTacToe extends Plugin {
   }
 
   private async _createGame(message: IMessage, oppMember: GuildMember) {
-    const game = new TTTGame(message.author, oppMember.user);
+    const game = new TTTGame(
+      message.author,
+      oppMember.user,
+      message.mentions.members?.first()?.roles.cache.some((role) => role.name === 'Chatbot')
```
tanndlin, line 49: There is a `MemberUtils.hasRole(member:GuildMember, roleName:string)`

src/app/plugins/tictactoe.plugin.ts
```diff
@@ -92,24 +96,27 @@ class TTTGame {
...
   private _col = -1;
 
   // -1 is playerA
   private currentPlayer: number = -1;
 
-  constructor(playerA: User, playerB: User) {
+  constructor(playerA: User, playerB: User, playingLion: Maybe<boolean>) {
     this._playerA = playerA;
     this._playerB = playerB;
+    this._playingLion = playingLion ? true : false;
```
tanndlin, line 119: Why is playing Lion `Maybe<...>`
Then you could just do `this._playingLion = playingLion`

src/app/plugins/tictactoe.plugin.ts
```diff
@@ -128,43 +135,110 @@ class TTTGame {
...
-    if (this._checkWin()) {
-      this._winner = this.currentPlayer;
-    }
+    this._board[this._row][this._col] = this.currentPlayer;
+    this._checkAndUpdateWin();
 
     this._flipTurn();
     this.reset();
     await msg.edit(this.showBoard());
     // Reset where the player is choosing
```
tanndlin, line 166: Comment looks out of place

src/app/plugins/tictactoe.plugin.ts
```diff
@@ -128,43 +135,110 @@ class TTTGame {
...
+          return;
+        }
+
+        // Make the move and evaluate the board state
+        this._board[row][col] = this.currentPlayer;
+        moves.push({ row, col, val: this._evaluate(this.currentPlayer * -1) });
+        // Backtrack.
+        this._board[row][col] = 0;
+      })
+    );
```
tanndlin, line 199: Does this not side effect the board?

src/app/plugins/tictactoe.plugin.ts
```diff
@@ -128,43 +135,110 @@ class TTTGame {
...
+    // For every location
+    this._board.forEach((_, row) =>
+      _.forEach((_, col) => {
+        if (this._board[row][col] !== 0) {
+          return;
+        }
+
+        // Make the move and evaluate the board state
+        this._board[row][col] = this.currentPlayer;
+        moves.push({ row, col, val: this._evaluate(this.currentPlayer * -1) });
```
tanndlin, line 195: <3 the flipping `this.currentPlayer * -1`

src/app/plugins/tictactoe.plugin.ts
```diff
@@ -128,43 +135,110 @@ class TTTGame {
...
+    this._board.forEach((_, row) =>
+      _.forEach((_, col) => {
+        if (this._board[row][col] !== 0) {
+          return;
+        }
+        this._board[row][col] = currentPlayer;
+        moves.push(this._evaluate(currentPlayer * -1));
+        this._board[row][col] = 0;
+      })
+    );
```
tanndlin, line 226: `.reduce`

src/app/plugins/tictactoe.plugin.ts
```diff
@@ -128,43 +135,110 @@ class TTTGame {
...
+
+    return this._sumArray(moves) / moves.length;
```
tanndlin, line 228: Should probably comment why you do this

src/app/plugins/tictactoe.plugin.ts
tanndlin: Probably better to make sure that `message.author.id === this.container.clientService.id`
That way in case there is ever someone else with Chatbot, it won't mess it up

src/app/plugins/tictactoe.plugin.ts
TylerMathis: PlayingLion is maybe because of the null ability of the members cache. Depending on that helper function mentioned above though we might be able to fix this.

src/app/plugins/tictactoe.plugin.ts
TylerMathis: No, it backtracks. Make change, recurse, undo change. All changes will be undone.

src/app/plugins/tictactoe.plugin.ts
TylerMathis: I'll try my best 😂

src/app/plugins/tictactoe.plugin.ts
TylerMathis: This is much easier, thanks!

src/app/plugins/tictactoe.plugin.ts
TylerMathis: Resolved with the change mentioned above

src/app/plugins/tictactoe.plugin.ts
TylerMathis: We don't really care about the author, so I swapped to `oppMember.id === this.container.clientService.user?id`

src/app/plugins/tictactoe.plugin.ts
TylerMathis: <3

src/app/plugins/tictactoe.plugin.ts
TylerMathis: For some reason, using reduce  in this context is modifying the content of `this._board` I don't understand it at all. I also don't see the advantage of reduce in this context, other than removing a single line of code. Why do we prefer reduce over a foreach in this instance?

tanndlin: I made the reduce, but it comes out uglier than what is already there

tanndlin (approved):

PR #438, opened 2021-05-30
Tylermathis/ai/connectfour

TylerMathis:
Maybe I just suck at Connect Four, but I have yet to beat it :o

Check it out and try for yourself.
In order to recognize that you are playing a bot, the tagged user must have the @Chatbot role

### DIFF
--- a/src/app/plugins/connectfour.plugin.ts
+++ b/src/app/plugins/connectfour.plugin.ts
@@ -39,7 +39,11 @@ export class ConnectFourPlugin extends Plugin {
   }
 
   private async _createGame(message: IMessage, oppMember: GuildMember) {
-    const game = new ConnectFourGame(message.author, oppMember.user);
+    const game = new ConnectFourGame(
+      message.author,
+      oppMember.user,
+      message.mentions.members?.first()?.roles.cache.some((role) => role.name === 'Chatbot')
+    );
     const msg = await message.reply(game.showBoard());
     await Promise.all(ConnectFourPlugin.moves.map((emoji) => msg.react(emoji)));
 
@@ -65,7 +69,7 @@ export class ConnectFourPlugin extends Plugin {
 
       await game.move(ConnectFourPlugin.moves.indexOf(react.emoji.name), msg);
 
-      if (game.getWinner() || game.getTie()) {
+      if (game.getGameOver()) {
         collector.stop();
         return;
       }
@@ -95,16 +99,21 @@ class ConnectFourGame {
   private _cols: number = 7;
 
   private currentPlayer: number = -1;
+  private _playingLion: boolean;
+  private _aiDepth = 4;
 
-  private winner: Maybe<number> = null;
-  private tie: boolean = false;
+  private _winner: Maybe<number> = null;
+  private _tie: boolean = false;
+  private gameOver: boolean = false;
 
   private _searchDX: number[] = [-1, -1, 0, 1, 1, 1, 0, -1];
   private _searchDY: number[] = [0, -1, -1, -1, 0, 1, 1, 1];
 
-  constructor(playerA: User, playerB: User) {
+  constructor(playerA: User, playerB: User, playingLion: Maybe<boolean>) {
     this._playerA = playerA;
     this._playerB = playerB;
+
+    this._playingLion = playingLion ? true : false;
 
     this._board = Array.from(Array(this._rows), (_) => Array(this._cols).fill(0));
   }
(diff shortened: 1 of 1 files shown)


src/app/plugins/connectfour.plugin.ts
```diff
@@ -96,112 +100,214 @@ class ConnectFourGame {
...
 
   private _searchDX: number[] = [-1, -1, 0, 1, 1, 1, 0, -1];
   private _searchDY: number[] = [0, -1, -1, -1, 0, 1, 1, 1];
 
-  constructor(playerA: User, playerB: User) {
+  constructor(playerA: User, playerB: User, playingLion: Maybe<boolean>) {
     this._playerA = playerA;
     this._playerB = playerB;
 
+    this._playingLion = playingLion ? true : false;
```
c650, line 119: silly ternary.

src/app/plugins/connectfour.plugin.ts
```diff
@@ -96,112 +100,214 @@ class ConnectFourGame {
...
+  private _getNormalizedDifficulty() {
+    return 1 / this._aiDifficulty;
+  }
+
+  // Evaluate the strength of the current board state,
+  // as it relates to Player 2's success.
+  private _evaluate(currentPlayer: number, depth: number) {
+    // If we have reached a win state, then the LAST move won.
+    if (this._checkWin()) {
+      return -4 * currentPlayer;
```
c650, line 181: -4?

src/app/plugins/connectfour.plugin.ts
```diff
@@ -96,112 +100,214 @@ class ConnectFourGame {
...
+    }
+    return row;
+  }
+
+  private _longestChainOnBoard(currentPlayer?: number): number {
+    let longestChain = 0;
+    this._board.forEach((rowObj, rowIdx) => {
+      rowObj.forEach((_, colIdx) => {
+        longestChain = Math.max(
+          longestChain,
```
c650, line 249: use `reduce` so that you can do

```js
const longestchain = this._board.reduce(...)
```

c650 (approved):

src/app/plugins/connectfour.plugin.ts
tanndlin: Better yet
```ts
return this._board.reduce(...)
```

src/app/plugins/connectfour.plugin.ts
TylerMathis: Oops, looked like I dumped this change before I pushed.

src/app/plugins/connectfour.plugin.ts
TylerMathis: Addressed with comment.

src/app/plugins/connectfour.plugin.ts
TylerMathis: reduce will always return a value that is one dimension less than the object passed into it. How can I get around that for this case?

TylerMathis: As reference, this is what does not currently work.
`  private _longestChainOnBoard(currentPlayer?: number): number {
    return this._board.reduce((longestChain, rowObj, row) => {
      rowObj.forEach((_, col) => {
        longestChain = Math.max(
          longestChain,
          this._longestChainAtLocation(row, col, currentPlayer);
        );
      });
      return this._longestChainOnBoard;
    }, 0);`

tanndlin: > reduce will always return a value that is one dimension less than the object passed into it. How can I get around that for this case?

 Not necessarily, you can make it return whatever you want

tanndlin: > As reference, this is what does not currently work.
> 
> ```javascript
> 
> private _longestChainOnBoard(currentPlayer?: number): number {
> 
>     return this._board.reduce((longestChain, rowObj, row) => {
> 
>       rowObj.forEach((_, col) => {
> 
>         longestChain = Math.max(
> 
>           longestChain,
> 
>           this._longestChainAtLocation(row, col, currentPlayer);
> 
>         );
> 
>       });
> 
>       return this._longestChainOnBoard;
> 
>     }, 0);```

Foreach, doesn't return anything, you need a double reduce, or take the max of all 3 rows, so like
```ts 
return Math.max(this._board.map(row=> row.reduce((acc, rowObj, row) => Math.max(calculate the max, acc)
));
```

TylerMathis: Makes sense, reduce will always just feel useless to me I suppose ahahah

src/app/plugins/connectfour.plugin.ts
```diff
@@ -96,112 +100,211 @@ class ConnectFourGame {
...
+    // We use a very simple evaluation: -(longest player one chain)
+    // This causes Lion to simply try to minimize Player 1's success,
+    // While our _checkWin case takes over when it's time to clinch victory.
+    if (depth === this._aiDepth) {
+      return this._longestChainOnBoard(1) - this._longestChainOnBoard(-1);
     }
-    // Subtract one to place piece one slot above obstruction.
+
+    const moves: number[] = [];
+    for (let col = 0; col < this._cols; col++) {
```
tanndlin, line 197: Can we use a foreach?

src/app/plugins/connectfour.plugin.ts
```diff
@@ -96,112 +100,211 @@ class ConnectFourGame {
...
+    // Find first obstruction from top down in column.
+    let row = 0;
+    while (row < this._rows && this._board[row][col] === 0) {
+      row++;
+    }
+    return row;
+  }
+
+  private _longestChainOnBoard(currentPlayer?: number): number {
+    return this._board.reduce((longestChainOnBoard, rowObj, row) =>
```
tanndlin, line 247: This variable name is super long lmao, maybe `longestChain`

src/app/plugins/connectfour.plugin.ts
```diff
@@ -96,112 +100,211 @@ class ConnectFourGame {
...
+          chains[direction]++;
         }
       }
+    }
 
-      // Return whether or not we found a chain of four.
-      return chains.includes(4);
-    };
+    // Return whether or not we found a chain of four.
+    return Math.max(...chains);
```
tanndlin, line 271: Do we need the spread operator? I thought `Math.max` could take in a `number[]` as is

TylerMathis: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max

src/app/plugins/connectfour.plugin.ts
TylerMathis: Why? We only need an index here, and the index needs to be column oriented. In order to pull columns indices that are semantically correct out of a forEach we would need

```javascript
this._board.forEach((_row, row) => row.forEach((_col, col) => {
    // do the thing.
})
```

This takes more time, and more memory.


If I just misinterpreted what you meant let me know.

src/app/plugins/connectfour.plugin.ts
TylerMathis: We have a `longestChainOnBoard` and a `longestChainAtLocation`.

For the sake of people extending this class I feel like having the distinction is important no?

src/app/plugins/connectfour.plugin.ts
tanndlin: oh is it column oriented?
Then nevermind

src/app/plugins/connectfour.plugin.ts
tanndlin: Fair enough

PR #444, opened 2021-06-02
Game Plugin leaderboards

### LINKED ISSUE #434: Connect Four and TicTacToe Leaderboard
would have to use MongoDB

tanndlin:
fixes #434

### DIFF
--- a/src/app/plugins/connectfour.plugin.ts
+++ b/src/app/plugins/connectfour.plugin.ts
@@ -3,7 +3,7 @@ import moment from 'moment';
 import Constants from '../../common/constants';
 import { Plugin } from '../../common/plugin';
 import { ChannelType, IContainer, IMessage, Maybe } from '../../common/types';
-import { GameResult, Games } from '../../services/gameleaderboard.service';
+import { GameResult, GameType } from '../../services/gameleaderboard.service';
 
 export class ConnectFourPlugin extends Plugin {
   public name: string = 'Connect Four';
@@ -94,7 +94,7 @@ export class ConnectFourPlugin extends Plugin {
       // update the leaderboard for the author of the game
       await this.container.gameLeaderboardService.updateLeaderboard(
         message.author,
-        Games.ConnectFour,
+        GameType.ConnectFour,
         {
           opponent: oppMember.user.id,
           result: convertToResult(-1),
@@ -104,7 +104,7 @@ export class ConnectFourPlugin extends Plugin {
       // update the leaderboard of the opponent
       await this.container.gameLeaderboardService.updateLeaderboard(
         oppMember.user,
-        Games.ConnectFour,
+        GameType.ConnectFour,
         {
           opponent: message.author.id,
           result: convertToResult(1),
--- a/src/app/plugins/connectfour.plugin.ts
+++ b/src/app/plugins/connectfour.plugin.ts
@@ -3,6 +3,7 @@ import moment from 'moment';
 import Constants from '../../common/constants';
 import { Plugin } from '../../common/plugin';
 import { ChannelType, IContainer, IMessage, Maybe } from '../../common/types';
+import { GameResult, Games } from '../../services/gameleaderboard.service';
 
 export class ConnectFourPlugin extends Plugin {
   public name: string = 'Connect Four';
@@ -77,7 +78,39 @@ export class ConnectFourPlugin extends Plugin {
       await react.users.remove(user);
     });
 
-    collector.on('end', () => {
+    collector.on('end', async () => {
+      const convertToResult = (x: number) => {
+        if (game.getWinner() === x) {
+          return GameResult.Won;
+        }
+
+        if (game.checkTie() === true) {
+          return GameResult.Tie;
+        }
+
+        return GameResult.Lost;
+      };
+
+      // update the leaderboard for the author of the game
+      await this.container.gameLeaderboardService.updateLeaderboard(
+        message.author,
+        Games.ConnectFour,
+        {
+          opponent: oppMember.user.id,
+          result: convertToResult(-1),
+        }
+      );
+
+      // update the leaderboard of the opponent
+      await this.container.gameLeaderboardService.updateLeaderboard(
+        oppMember.user,
+        Games.ConnectFour,
+        {
+          opponent: message.author.id,
+          result: convertToResult(1),
+        }
+      );
+
       msg.reactions.removeAll();
     });
   }
@@ -127,6 +160,19 @@ class ConnectFourGame {
 
   public getGameOver() {
     return this.gameOver;
+  }
+
+  public getWinner() {
+    return this._winner;
+  }
+
+  public getLoser() {
+    // -1 means playerA won
+    return this.getWinner() === -1 ? this._playerB : this._playerA;
+  }
+
+  public getTie() {
+    return this._tie;
   }
 
   public async move(col: number, msg: IMessage) {
@@ -182,7 +228,7 @@ class ConnectFourGame {
     if (this._checkWin()) {
       return -4 * currentPlayer;
     }
-    if (this._checkTie()) {
+    if (this.checkTie()) {
       return 0;
     }
     // If we reached depth, then evaluate the board.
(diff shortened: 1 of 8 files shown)


src/app/plugins/leaderboard.plugin.ts
```diff
@@ -20,18 +20,78 @@ export class LeaderboardPlugin extends Plugin {
...
-    if (!opponent) {
+    if (!opponentOne) {
       const embed = await this.container.gameLeaderboardService.createOverallLeaderboardEmbed(
         message.author,
         gameEnum
       );
+
+      message.channel.send(embed);
+      return;
+    } else if (!opponentTwo) {
```
tanndlin, line 38: Dont need `else` since we return after the preceding `if`

src/app/plugins/leaderboard.plugin.ts
```diff
@@ -20,18 +20,78 @@ export class LeaderboardPlugin extends Plugin {
...
+      const match = opponentOne.match(/<@!?&?(\d+)>/);
```
tanndlin, line 39: Make this regex a class constant

src/app/plugins/leaderboard.plugin.ts
```diff
@@ -20,18 +20,78 @@ export class LeaderboardPlugin extends Plugin {
...
+      if (!match) {
+        message.reply('Invalid <matchup> argument');
+        return;
+      }
+      const uID = match[1];
```
tanndlin, line 44: prob could use a comment why its the second one
Also array destructuring

src/app/plugins/leaderboard.plugin.ts
```diff
@@ -20,18 +20,78 @@ export class LeaderboardPlugin extends Plugin {
...
+
+      const embed = await this.container.gameLeaderboardService.createMatchupLeaderboardEmbed(
+        message.author,
+        oppUser.user,
+        gameEnum
+      );
+
+      message.channel.send(embed);
+      return;
+    } else {
```
tanndlin, line 65: saa
no need to use `else` if we return out of the previous `if`

Reduces indentation, thus complexity

src/app/plugins/leaderboard.plugin.ts
```diff
@@ -20,18 +20,78 @@ export class LeaderboardPlugin extends Plugin {
...
+      const matchOne = args[1].match(/<@!?&?(\d+)>/);
+      const matchTwo = args[2].match(/<@!?&?(\d+)>/);
```
tanndlin, line 67: saa make these a class constant

src/app/plugins/leaderboard.plugin.ts
```diff
@@ -20,18 +20,78 @@ export class LeaderboardPlugin extends Plugin {
...
+      }
+
+      const oppUserOne = guild.members.cache.get(uIDOne);
+      const oppUserTwo = guild.members.cache.get(uIDTwo);
+
+      // user could not be found
+      if (!oppUserOne || !oppUserTwo) {
+        message.channel.send('User could not be found');
+        return;
+      }
```
tanndlin, line 87: Sorta duplicate code, maybe there is a way to combine these into functions for 
```
input => regex => check match => get user with match
```

src/services/gameleaderboard.service.ts
```diff
@@ -183,7 +183,7 @@ export class GameLeaderboardService {
     const [userOneEntry] = entries.filter((e) => e.userId == userOne.id);
 
     if (!userOneEntry) {
-      return 'User not found';
+      return `User with id ${userOne.id} not found`;
```
tanndlin, line 186: prob should return their nickname/username
The user wont know what the id means since its just a bunch of random numbers

src/app/plugins/connectfour.plugin.ts
```diff
@@ -129,6 +162,19 @@ class ConnectFourGame {
     return this.gameOver;
   }
 
+  public getWinner() {
+    return this._winner;
+  }
+
+  public getLoser() {
+    // -1 means playerA won
+    return this.getWinner() === -1 ? this._playerB : this._playerA;
```
joey-colon, line 171: can we avoid magic numbers?

src/app/plugins/leaderboard.plugin.ts
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    return this.container.gameLeaderboardService.createMatchupLeaderboardEmbed(
+      oppUserOne.user,
+      oppUserTwo.user,
+      gameEnum
+    );
+  }
+
+  private _convertGameNameToEnum(gameName: string): Maybe<GameType> {
```
joey-colon, line 121: np: `_getGameType()`

src/app/plugins/leaderboard.plugin.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  private _getMatchUpEmbed(
+    message: IMessage,
+    playerOneString: string,
+    playerTwoString: string,
+    guild: Guild,
+    gameEnum: GameType
+  ) {
+    const matchOne = playerOneString.match(this._MENTION_REGEX);
+    const matchTwo = playerTwoString.match(this._MENTION_REGEX);
```
joey-colon, line 98: live ur best life but i would write it as:
`const [matchOne, matchTwo] = [playerOneString, playerTwoString].map(p => p.match(this._MENTION_REGEX))`

src/services/gameleaderboard.service.ts
```diff
@@ -0,0 +1,280 @@
...
+
+    // If we were able to find user's data, put them at the top
+    const userFieldData = this._createPlayerFieldData(user, entries);
+    if (userFieldData) {
+      const { name, value } = userFieldData;
+      embed.addField(name, value); // Caller's rank
+    }
+
+    entries
+      .slice(0, 24) // Max of 25 fields per embed
```
joey-colon, line 110: remove comment and use a descriptive var name

src/services/gameleaderboard.service.ts
```diff
@@ -0,0 +1,280 @@
...
+      `${userOne.username} vs ${userTwo.username}`,
+      `__Wins:__ ${userOneWins} - ${userTwoWins}\n` +
+        `__Loses:__ ${userTwoWins} - ${userOneWins}\n` +
+        `__Ties:__ ${ties}`
+    );
+    return embed;
+  }
+
+  private _getOverallStats(entry: IGameLeaderBoardEntry): Maybe<IUserOverallEntry> {
+    const user = this._guildService.get().members.cache.get(entry.userId)?.user;
```
joey-colon, line 222: what if cache miss? can we `||` this declaration with an actual fetch query to discord?

src/services/gameleaderboard.service.ts
```diff
@@ -0,0 +1,280 @@
...
+
+    if (!userOneEntry) {
+      return `User \`${userOne.username}\` not found`;
+    }
+
+    const matchupGames = userOneEntry.games.filter((game: IGame) => game.opponent == userTwo.id);
+
+    let userOneWins = 0,
+      userTwoWins = 0,
+      ties = 0;
```
joey-colon, line 193: can u use `let` for each declaration since you did it below

src/services/gameleaderboard.service.ts
```diff
@@ -0,0 +1,280 @@
...
+  Lost,
+  Tie,
+}
+
+export enum GameType {
+  TicTacToe = 1,
+  ConnectFour,
+}
+
+type GameCollectionTypes = 'tttLeaderboard' | 'connectFourLeaderboard';
```
joey-colon, line 280: np: can u move these types to be at the top of file below imports?

src/services/gameleaderboard.service.ts
joey-colon: also i thought we added some sort of util that has this fallback functionality?

src/services/gameleaderboard.service.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  private _getCollection(
+    gameType: GameCollectionTypes
+  ): (game: GameType) => Promise<Maybe<Collection<IGameLeaderBoardEntry>>> {
+    const collections = this._storageService.getCollections();
+    return async () => (await collections)[gameType];
+  }
+
+  // return the appropriate mongo collection used for the given game
+  private async _gameToCollection(game: GameType) {
```
joey-colon, line 81: do we need this method if we are already maintaining `_gameEnumToCollection`? looks like we could just hydrate it with the collections initially in the constructor

src/app/plugins/tictactoe.plugin.ts
```diff
@@ -87,7 +88,41 @@ export class TicTacToe extends Plugin {
...
+
+        if (game.checkTie() === true) {
+          return GameResult.Tie;
+        }
+
+        return GameResult.Lost;
+      };
+
+      // update the leaderboard for the author of the game
+      await this.container.gameLeaderboardService.updateLeaderboard(
```
joey-colon, line 105: corgo would tell u to `Promise.all` this

src/app/plugins/connectfour.plugin.ts
tanndlin: We need it in one place, just inside of getWinner, but we can make `getWinner` return a user so that its not a magic number anymore

src/app/plugins/leaderboard.plugin.ts
tanndlin: agree

src/services/gameleaderboard.service.ts
tanndlin: members.cache is the only way afaik

src/services/gameleaderboard.service.ts
tanndlin: We have to await getting the collections, that was why we were having a hard time finding a way to map these to string or enums in the constructor

Cant have async in constructor

PR #445, opened 2021-06-03
Add twitter plugin

suneettipirneni:
Adds a plugin to return the latest tweets from UCF accounts, results display in embeds as shown below:

Screen Shot 2021-06-03 at 12 34 48 AM

### Supported UCF Twitter Accounts:

Currently Five UCF accounts are supports

- @UCF,
- @UCF_Football,
- @UCFCECS,
- @KnightHacks
- @UCF_Knights

### Env File changes:

The `TWITTER_BEARER_TOKEN` env var is now required for twitter access, this information can be obtained here: https://developer.twitter.com/en/apply-for-access


src/app/plugins/twitter.plugin.ts
```diff
@@ -0,0 +1,103 @@
...
+    public get description(): string {
+        return 'Gets the latest twitter timelines from UCF accounts';
+    }
+    public get usage(): string {
+        throw 'twitter <UCF account>';
+    }
+
+    public get permission(): ChannelType {
+        return ChannelType.Public
+    }
```
tanndlin, line 33: Look at the other plugins for reference, we just make these public fields

src/app/plugins/twitter.plugin.ts
```diff
@@ -0,0 +1,103 @@
...
+    knighthacks: '3122136832',
+    cecs: '2292877801',
+}
+
+const twitterIconURL = 'https://images-ext-1.discordapp.net/external/bXJWV2Y_F3XSra_kEqIYXAAsI3m1meckfLhYuWzxIfI/https/abs.twimg.com/icons/apple-touch-icon-192x192.png';
+
+export class TwitterPlugin extends Plugin {
+
+    // The twitter API returns a min of 5 tweets, but thats a bit much for our bot, so we'll do 3 instead.
+    maxSize = 3;
```
tanndlin, line 19: make this a private and put it down with the other class property `twitter`

src/app/plugins/twitter.plugin.ts
```diff
@@ -0,0 +1,103 @@
...
+        super();
+    }
+
+    public async execute(message: IMessage, args: string[]) {
+        const [param] = args;
+
+        let accountId;
+
+        // Default to UCF account if no args provided.
+        if (args.length === 0) {
```
tanndlin, line 47: I'd prefer `if(param)`

src/app/plugins/twitter.plugin.ts
```diff
@@ -0,0 +1,103 @@
...
+
+const accounts: Record<string, string> = {
+    ucf: '18999501',
+    knights: '21306514',
+    football: '30282826',
+    knighthacks: '3122136832',
+    cecs: '2292877801',
+}
+
+const twitterIconURL = 'https://images-ext-1.discordapp.net/external/bXJWV2Y_F3XSra_kEqIYXAAsI3m1meckfLhYuWzxIfI/https/abs.twimg.com/icons/apple-touch-icon-192x192.png';
```
tanndlin, line 14: make these 2 private class constants

src/app/plugins/twitter.plugin.ts
```diff
@@ -0,0 +1,103 @@
...
+
+        message.react('👍');
+        message.reply('Sure thing! Getting latest tweets!');
+
+        // Fetch respective tweets.
+        const response = await this.twitter.getLatestTweets(accountId, this.maxSize);
+        const embeds = await this._createEmbedList(response, accountId);
+
+        // Send out all of the fetched tweets.
+        embeds.forEach(embed => message.reply(embed));
```
tanndlin, line 70: How many embeds can this be?
We may want to combine them into one embed so that we dont spam / get rate limited

src/services/twitter.service.ts
```diff
@@ -0,0 +1,118 @@
...
+      nextToken: string;
+  };
+  includes?: {
+    media?: TwitterMedia[]
+  }
+}
+
+export class TwitterService {
+  private bearerToken = Environment.TwitterBearerToken!;
+  private http = axios.create();
```
tanndlin, line 55: Did you try using `container.httpService` first?

src/services/twitter.service.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  public constructor() {
```
tanndlin, line 57: The way that we create services, is they are instantiated in `container.ts` then we called them from within the plugin with
```
this.container.twitterService...
```
Look at some other ones for an example

tanndlin (commented):
Pretty elegant code :)

tanndlin:
Your diff looks good, but there are a lot of commits in here, which is fine because we squash PRs
But next use rebase so that the history for the PR is clean 👍

src/app/plugins/twitter.plugin.ts
suneettipirneni: This has a max of 3 embeds. Combining them all together is less than ideal since we'll lose a lot of rich content. However I've now adjusted the code to use a webhook in order to send all three in a single payload as opposed to 3 different payloads, so it shouldn't affect rate limits.

src/services/twitter.service.ts
suneettipirneni: Yeah, I did. With the twitter API every request needs the bearer access token in headers I find using a custom axios instance that automatically intercepts this for every request more convenient.

src/app/plugins/twitter.plugin.ts
```diff
@@ -0,0 +1,99 @@
...
+        const [param] = args;
+
+        let accountId;
+
+        // Default to UCF account if no args provided.
+        if (!param) {
+            accountId = TwitterPlugin.accounts.ucf;
+        } else {
+            accountId = TwitterPlugin.accounts[param.toLowerCase()];
+        }
```
c650, line 41: i dont like this pattern

src/app/plugins/twitter.plugin.ts
```diff
@@ -0,0 +1,99 @@
...
+        if (!accountId) {
+            let options = '\n';
+            Object.keys(TwitterPlugin.accounts).forEach(key => options += `🔸 ${key}\n`);
+            message.react('❌')
+            message.reply(`Invalid UCF Twitter Account \'${param}\', possible options are:${options}`);
+            return;
+        }
+
+
+        message.react('👍');
```
c650, line 53:
```
await Promise.all([ message.react, message.reply ])
```

src/app/plugins/twitter.plugin.ts
```diff
@@ -0,0 +1,99 @@
...
+        const response = await this.twitter.getLatestTweets(accountId, this.maxSize);
+        const embeds = await this._createEmbeds(response, accountId);
+
+
+        // The reason a webhook is used here is because traditional bot messages don't allow you
+        // to send multiple embeds at once, with a webhook you can send 10 at a time.
+        const webhook = await (message.channel as TextChannel).createWebhook('UCF Twitter', {
+            avatar: 'https://about.twitter.com/content/dam/about-twitter/en/brand-toolkit/brand-download-img-1.jpg.twimg.1920.jpg'
+        });
+        webhook.send({ embeds });
```
c650, line 66: await

src/services/twitter.service.ts
```diff
@@ -0,0 +1,118 @@
+import Environment from "../environment";
+import axios, { AxiosRequestConfig } from 'axios';
+
+export type Tweet = {
```
c650, line 4: y not interface?

c650 (approved):

src/app/plugins/twitter.plugin.ts
suneettipirneni: This is the last statement in the block and the promised value isn't used, the `send()` message will execute async but we don't care about the result, so prefixing with `await` here wouldn't make a difference.

src/services/twitter.service.ts
suneettipirneni: I doesn't make any difference in this instance to use a `type` over an `interface`, but I can change it if this is part of the style.

src/app/plugins/twitter.plugin.ts
suneettipirneni: Could you explain what you don't like?

tanndlin: prob use a ternary
We try to avoid `let` as much as possible

tanndlin:
gg

PR #446, opened 2021-06-03
Coin toss or selector

### LINKED ISSUE #409: Head or Tails plugin

ElbertTse:
Let me know if the image for picking should be changed. Right now I'm using a wand emoji but I'm open to swapping the image to something else. Closes #409.

- Can call the plugin without arguments which does a coin toss
- Can call the plugin with arguments and it will choose among the given arguments


src/app/plugins/cointoss.plugin.ts
```diff
@@ -0,0 +1,36 @@
...
+  constructor(public container: IContainer) {
+    super();
+  }
+
+  public async execute(message: IMessage, args?: string[]) {
+    let _responses: string[] = [];
+    let _image: string = '';
+
+    if (args?.length === 0) {
+      _image = 'https://emojipedia-us.s3.dualstack.us-west-1.amazonaws.com/thumbs/240/apple/285/coin_1fa99.png';
```
tanndlin, line 22: Make this a class constant

src/app/plugins/cointoss.plugin.ts
```diff
@@ -0,0 +1,36 @@
...
+  public permission: ChannelType = ChannelType.Public;
+  private _embed: MessageEmbed = new MessageEmbed();
+
+  constructor(public container: IContainer) {
+    super();
+  }
+
+  public async execute(message: IMessage, args?: string[]) {
+    let _responses: string[] = [];
+    let _image: string = '';
```
tanndlin, line 19: Make the array a const, you can push to an array that is constant
then you can just push with the spread operator
```
repsonses.push(...args)
```

tanndlin: Also we name things with a hanging underscore when they are a private property/method on a class, so no need to do that here

src/app/plugins/cointoss.plugin.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    if (args?.length === 0) {
+      _image = 'https://emojipedia-us.s3.dualstack.us-west-1.amazonaws.com/thumbs/240/apple/285/coin_1fa99.png';
+      this._embed.setAuthor('Lion flipped a coin and it lands on...', _image);
+      _responses = ['Heads', 'Tails']
+    } else {
+      _image = 'https://emojipedia-us.s3.dualstack.us-west-1.amazonaws.com/thumbs/120/apple/285/magic-wand_1fa84.png';
```
tanndlin, line 26: saa

src/app/plugins/cointoss.plugin.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  public async execute(message: IMessage, args?: string[]) {
+    let _responses: string[] = [];
+    let _image: string = '';
+
+    if (args?.length === 0) {
+      _image = 'https://emojipedia-us.s3.dualstack.us-west-1.amazonaws.com/thumbs/240/apple/285/coin_1fa99.png';
+      this._embed.setAuthor('Lion flipped a coin and it lands on...', _image);
+      _responses = ['Heads', 'Tails']
```
tanndlin, line 24: You should prob just make this a class constant as the default
Then if the user provides args, then just make a new variable for it

src/app/plugins/cointoss.plugin.ts
```diff
@@ -0,0 +1,34 @@
...
+  constructor(public container: IContainer) {
+    super();
+  }
+
+  public async execute(message: IMessage, args?: string[]) {
+    if (args?.length === 0) {
+      this._embed.setAuthor('Lion flipped a coin and it lands on...', this._coinImg);
+      this._responses.push(...this._defaultArgs);
+    } else {
+      const wandImg = 'https://emojipedia-us.s3.dualstack.us-west-1.amazonaws.com/thumbs/120/apple/285/magic-wand_1fa84.png';
```
tanndlin, line 24: when i say class constant, we mean like `private _IMG = ...`

src/app/plugins/cointoss.plugin.ts
```diff
@@ -0,0 +1,34 @@
...
+export class CoinToss extends Plugin {
+  public name: string = 'Coin Toss Plugin';
+  public description: string = 'Ask Lion to toss a coin or if given arguments, choose among the arguments.';
+  public usage: string = '!cointoss || !pick <arg_1> <arg_2> ... <arg_n>';
+  public pluginAlias: string[] = ['pick', 'select', 'cointoss', 'coin', 'toss', 'ct'];
+  public permission: ChannelType = ChannelType.Public;
+  private _embed: MessageEmbed = new MessageEmbed();
+  private _coinImg: string = 'https://emojipedia-us.s3.dualstack.us-west-1.amazonaws.com/thumbs/240/apple/285/coin_1fa99.png';
+  private _defaultArgs: string[] = ['Heads', 'Tails'];
+  private _responses: string[] = [];
```
tanndlin, line 14: If you make this a private property, it will affect the next time you call it, this one doesn't need to be a class constant

ElbertTse: For clarification, are you talking about _responses on line 14?

tanndlin: Yes

src/app/plugins/cointoss.plugin.ts
```diff
@@ -0,0 +1,40 @@
...
+      responses.push(...args);
+    }
+
+    const choice = Math.floor(Math.random() * responses.length)
+    this._embed.setColor('#0099ff').setTitle(responses[choice]);
+    
+    message.reply(this._embed);
+
+    // Need to reset args list
+    responses = [];
```
tanndlin, line 38: I see what you meant, but this isnt necessary because this is only scoped to the function
 So it gets erased when the function ends

ElbertTse: You're right, I think I may have been declaring responses with let instead of const but its fixed now

tanndlin (approved):

PR #449, opened 2021-06-03
Member milestone

### LINKED ISSUE #119: New online user record message
Send a message to some channel when we achieve a new "online users" record

Armintron:
Emits a message when 100 new members have joined since the last message. Fixes #119

### DIFF
--- a/src/app/handlers/membercount.handler.ts
+++ b/src/app/handlers/membercount.handler.ts
@@ -1,0 +1,62 @@
+import { GuildMember, MessageEmbed, TextChannel } from 'discord.js';
+import Constants from '../../common/constants';
+import { IContainer, IHandler, IServerInfo } from '../../common/types';
+
+export class MemberCountHandler implements IHandler {
+  private _memberMilestoneInterval = 100;
+
+  constructor(public container: IContainer) {}
+  public async execute(member: GuildMember) {
+    const collections = await this.container.storageService.getCollections();
+    const { serverInfo: serverInfoCollection } = collections;
+    const knightEmoji = member.guild.emojis.cache.find((e) => e.name === 'knight');
+
+    if (!serverInfoCollection) {
+      return;
+    }
+
+    const res = await serverInfoCollection
+      .find<IServerCount>({ name: 'MemberCount' })
+      .toArray();
+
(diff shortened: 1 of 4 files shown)


tanndlin (dismissed):

src/app/handlers/membercount.handler.ts
```diff
@@ -0,0 +1,62 @@
...
+    const { serverInfo: serverInfoCollection } = collections;
+    const knightEmoji = member.guild.emojis.cache.find((e) => e.name === 'knight');
+
+    if (!serverInfoCollection) {
+      return;
+    }
+
+    const res = await serverInfoCollection
+      .find<IServerCount>({ name: 'MemberCount' })
+      .toArray();
```
c650, line 20: why are you keeping an array?

tanndlin: Why have more than one entry of the record?
I was planning on storing it, so later on we can see the growth overtime, maybe eventually make a graph or something

src/app/handlers/membercount.handler.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    const lastSavedMemberCount = res[res.length - 1].count;
+
+    if (!res.length) {
+      return;
+    }
```
tanndlin, line 26: Should be above the preceding line

tanndlin (commented):
It needs to store the initial count, right now it returns if the length is 0
So make it so that once we hit count % 100 == 0
add it just once, if the collection is empty

src/app/handlers/membercount.handler.ts
```diff
@@ -15,17 +15,29 @@ export class MemberCountHandler implements IHandler {
...
-    const lastSavedMemberCount = res[res.length - 1].count;
+    const countToInsert: IServerCount = {
+      name: 'MemberCount',
+      count: currentCount,
+      dateUpdated: new Date(),
+    };
+
+    if (!memberCountDocs.length) {
+      // we want to add the initial count on a nice number
+      if (currentCount % 100 != 0) {
```
tanndlin, line 32: `!==`

tanndlin (approved):

PR #455, opened 2021-06-05
Create CONTRIBUTING.md

tanndlin:
anything i missed?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,6 +18,10 @@ This is not required but will help you get started.
 
 Make sure to assign yourself to an issue when you start on it, or plan on taking it, so that we don't work on the same one.
 
+For the most part, the admins will be merging the pull requests, but if your pull request reaches 2 approvals, then feel free to do so when you are ready.
+
+Make sure to PR onto the `develop` branch as this one is always even or ahead of master. When we are ready to deploy, we will merge develop onto master so that the bot can auto redeploy
+
 ## Testing Lion
 
 In order to test your code, you will need to set up a test server and [Discord Application](https://discord.com/developers/applications) to run Lion in. You can get the [UCF CS Discord Server](https://discord.gg/uXBmTd9) Template from one of our admins in the `#lion_project` channel.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,13 @@
+# Contributing to Lion
+## Branch names
+Make sure that your branch names follow our naming scheme `author/feature_name`
+For exmaple, `tanndlin/fixing_a_bug`, note that this is in snakecase
+## Necessary extensions
+1. Prettier
+2. ESLint
+Both of these extensions will make sure most of your code follows our conventions.
+## Variable names
+1. Camel Case
+2. Private properties on classes should start with an underscore. ie: `private _foo: number = 0`, `public bar: number = 1`
+## Github
+Make sure to asign yourself to an issue when you start on it, or plan on taking it, so that we don't work on the same one.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,11 +14,6 @@ This is not required but will help you get started.
 1. Prettier
 2. ESLint
 
-## Variable names
-
-1. Camel Case
-2. Private properties on classes should start with an underscore. ie: `private _foo: number = 0`, `public bar: number = 1`
-
 ## Github
 
 Make sure to assign yourself to an issue when you start on it, or plan on taking it, so that we don't work on the same one.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,13 @@
+# Contributing to Lion
+## Branch names
+Make sure that your branch names follow our naming scheme `author/feature_name`
+For exmaple, `tanndlin/fixing_a_bug`, note that this is in snakecase
+## Necessary extensions
+1. Prettier
+2. ESLint
+Both of these extensions will make sure most of your code follows our conventions.
```
c650, line 8: Put extra blank line. Did u check preview

CONTRIBUTING.md
```diff
@@ -0,0 +1,13 @@
+# Contributing to Lion
+## Branch names
+Make sure that your branch names follow our naming scheme `author/feature_name`
+For exmaple, `tanndlin/fixing_a_bug`, note that this is in snakecase
+## Necessary extensions
```
c650, line 5: Should mention that we all use vs code but it shouldn’t be mandated

CONTRIBUTING.md
```diff
@@ -0,0 +1,13 @@
+# Contributing to Lion
+## Branch names
+Make sure that your branch names follow our naming scheme `author/feature_name`
```
c650, line 3: I still like dev/author/thing

joey-colon: i like `author--thing` :)

TylerMathis:
I feel like a full solid `contributing.md` should go pretty in to depth on our entire development process. A lot of people are starting to hop onto the Lion dev train, and I think we want to make sure that the process is as clear as possible, all the way to down to plugin flow conventions.

ElbertTse:
I'm wondering if there should be anything on how to test lion aside from saying `npm run watch`. The only issue with adding in "how to run lion" is that there are plenty of resources on how to run a bot in your own discord server and honestly it might not even be w/in scope for this.

CONTRIBUTING.md
tanndlin: We've been doing author/thing for the most part so we should prob stick to that
Im not against `dev/author/thing`, but is the `dev` not redundant? Is there ever a time someone shouldn't put `dev`?

suneettipirneni: I think something that hasn't been brought up is the situation where branches have multiple authors. A lot of times people will ask for help after they create a branch, and it seems a bit unfair for it to be labeled only under one person.

tanndlin:
Made some edits, lmk what you think

CONTRIBUTING.md
tanndlin: I agree with this, what a good naming convention for when it's multiple people?
Putting each name is too long

ElbertTse: I'd say maybe rely on the participants/assignee list on github or listing the authors in the pull request? Still probably not the best way though

suneettipirneni: Expanding on what Elbert said not only are people assigned for the ticket they want to work on but the assigned list is dynamic. Branch names are not dynamic, you have to manually check another one out in order to name it appropriately, and I think everyone can agree that branch names with multiple authors isn't the nicest way to go about doing things. Luckily git history fills this in for us users are automatically audited and tagged in PR based on the git history, in other words, it's dynamic, there's no question of who did what. I just feel like keeping author names in the branch name causes more issues than they solve.

TylerMathis: I'd agree with this. I'm a fan of `category/specific`, with as many levels of depth as you need.

For an example, if you were fixing a bug in the AI side of Connect Four, you'd open the branch `bugfix/ai/connect_four`, this leaves no ambiguity, and allows got'a branch name autocomplete to do its job nicely in the CLI.

Got already has capabilities to manage authors in both commits and PRs.

ElbertTse: If we want to stick with `author/thing` then we would need a way to make sure that the first person that assigns themselves to the issue is the one who makes the branch. I do like the idea of `category/specific` though. We would just need a list of categories which would may or may not be all encompassing and can maybe correlate to the tags we have for issues.

suneettipirneni:
Another suggestion would be making the `develop` branch the default branch. This will automatically set it to the base branch when opening a PR

tanndlin:
> Another suggestion would be making the `develop` branch the default branch. This will automatically set it to the base branch when opening a PR

I've been debating this, but I'm not sure if there will be any other consequences that I don't know about

suneettipirneni:
Wait why was this merged? Did we ever reach an agreement on how would should name branches?

tanndlin:
I kept it as what we are currently, and have been, doing. If we reach an agreement it should be changed, then we can make the change. I just wanted something up there

@c650 and @joey-colon will have the final say since this is their project

suneettipirneni:
> I kept it as what we are currently, and have been, doing. If we reach an agreement it should be changed, then we can make the change. I just wanted something up there
> 

Ok that's fine and I understand, but it's hard for an outsider to tell what's going on if you just merge a PR that had an unresolved issue, without any context given.

PR #457, opened 2021-06-05
Fix TicTacToe AI

TylerMathis:
The AI I wrote before had some small flaws. Now it's perfect. Finds all moves on the board that will lead to a win and picks a random one. Unbeatable Lion, hurah!


src/app/plugins/tictactoe.plugin.ts
```diff
@@ -192,19 +192,24 @@ class TTTGame {
...
         this._board[row][col] = 0;
       })
     );
 
     // Sort the moves in decreasing value
     moves.sort((a, b) => b.val - a.val);
-    return { bestRow: moves[0].row, bestCol: moves[0].col };
+
+    // Calculate all moves with equal value, and return one.
+    const bestMoves = moves.filter(move => move.val === moves[0].val);
```
tanndlin, line 205: Is it guaranteed that `moves.length !== 0`?

TylerMathis: Yes

TylerMathis: This function only runs if the game is not over.

tanndlin (approved):

PR #460, opened 2021-06-06
Enforce Style Guide via Eslint

suneettipirneni:
Correlated to #455 

Enforces private class members/methods to be prefixed with an underscore:

Screen Shot 2021-06-06 at 12 57 35 AM

Globally accessible class declarations must be declared in UPPER_CASE:

Screen Shot 2021-06-06 at 12 59 20 AM

Interface Declarations must be Prefixed with "I":

Screen Shot 2021-06-06 at 1 01 10 AM

Some other notable rules:

- Triple equals enforcement (where applicable)

All code that wasn't adhering to these rules have now been reformatted to conform to these rules.

### DIFF
--- a/src/app/plugins/connectfour.plugin.ts
+++ b/src/app/plugins/connectfour.plugin.ts
@@ -100,7 +100,7 @@ export class ConnectFourPlugin extends Plugin {
       });
 
       await Promise.all(updates);
-      void msg.reactions.removeAll();
+      msg.reactions.removeAll();
     });
   }
 }
--- a/src/app/plugins/connectfour.plugin.ts
+++ b/src/app/plugins/connectfour.plugin.ts
@@ -78,7 +78,7 @@ export class ConnectFourPlugin extends Plugin {
     });
 
     collector.on('end', () => {
-      msg.reactions.removeAll();
+      void msg.reactions.removeAll();
     });
   }
 }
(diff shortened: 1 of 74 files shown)


tanndlin (commented):
This is pog
Does this auto fix for you though? In most cases, if the rule is set to warn rather than error it will auto-fix on save

suneettipirneni:
> This is pog
> Does this auto fix for you though? In most cases, if the rule is set to warn rather than error it will auto-fix on save

No, I'm guessing the reason autofixes aren't available is because if you use a symbol in multiple areas all of those areas will have to be renamed as well. But once everyone starts typing and sees the red underlines they'll be quick to fix any errors.

tanndlin (approved):
Thanks this was much needed, didn't know ESLint had these rules

suneettipirneni:
> Why put void infront of almost every method call?

Sometimes when async methods are invoked we don't want to always handle their result. However in doing this it introduces some confusion as to where the method being called is async or not. The `void` operator basically indicates "this method is async but we're not handling the result"

joey-colon:
i feel like `void` rule is more confusing than it adds benefits

suneettipirneni:
> i feel like `void` rule is more confusing than it adds benefits

The floating promises rule is now removed.

tanndlin (approved):

src/app/plugins/connectfour.plugin.ts
```diff
@@ -100,7 +100,7 @@ export class ConnectFourPlugin extends Plugin {
       });
 
       await Promise.all(updates);
-      msg.reactions.removeAll();
+      void msg.reactions.removeAll();
```
tanndlin, line 103: Still some left over `void`s in here

joey-colon (approved):

PR #461, opened 2021-06-06
Create a 'Game' Abstract to Modularize Leaderboard

TylerMathis:
Game's are now abstracted out.

`getWinner`, `getLoser`, and `checkTie` are now required to be implemented in all children.

After a game is over, log a result with `game.recordResult();`


tanndlin:
next time use rebase so that you dont have merge commits

src/app/abstracts/game.abstract.ts
```diff
@@ -0,0 +1,54 @@
...
+      ),
+      this._container.gameLeaderboardService.updateLeaderboard(
+        loser,
+        this._game,
+        {
+          opponent: winner.id,
+          result: GameResult.Lost,
+        }
+      )
+    ];
```
tanndlin, line 35: This doesnt handle ties

tanndlin: try this (and change accordingly to fit your abstract)
```ts
      const updates = [
        this.container.gameLeaderboardService.updateLeaderboard(
          message.author,
          GameType.TicTacToe,
          {
            opponent: oppMember.user.id,
            result: convertToResult(message.author),
          }
        ),
        this.container.gameLeaderboardService.updateLeaderboard(
          oppMember.user,
          GameType.TicTacToe,
          {
            opponent: message.author.id,
            result: convertToResult(oppMember.user),
          }
        ),
      ];
```

src/app/abstracts/game.abstract.ts
```diff
@@ -0,0 +1,54 @@
+import { User } from 'discord.js';
+import { IContainer } from '../../common/types';
+import { GameResult, GameType } from '../../services/gameleaderboard.service';
+export default abstract class Game {
+  private _container: IContainer;
+  private _game: GameType;
+
+  constructor(container: IContainer, game: GameType) {
```
tanndlin, line 8: container is normally public in here, that way you dont have to set it to a private field

tanndlin:
Wouldnt it make sense to do game.ts and put it in common, just like plugin.ts?

src/app/abstracts/game.abstract.ts
TylerMathis: How would we give the container to the game?

tanndlin: This is what it normally looks like
```
  super();
}
```
And you can just access it with this.container

TylerMathis: What would the declaration look like? If container wasn't declared, then it'll just say that's an unused variable.

src/common/game.ts
```diff
@@ -0,0 +1,64 @@
...
+          result: playerResult(this.playerB),
+        }
+      )
+    ];
+
+    await Promise.all(updates);
+  }
+
+  /**
+   * @returns User: The winner of the game
```
suneettipirneni, line 51: JSDoc type annotations are in the format:
```
@returns {Type} description.
```

You could leave the type in the jsdoc out since it's in typescript anyways.

src/app/abstracts/game.abstract.ts
tanndlin: You mean in your case? Because the code I put is what we've always been using

src/app/plugins/connectfour.plugin.ts
```diff
@@ -171,12 +133,12 @@ class ConnectFourGame {
 
   public getWinner(): User {
     // -1 is playerA
-    return this._winner === -1 ? this._playerA : this._playerB;
+    return this._winner === -1 ? this.playerA : this.playerB;
```
tanndlin, line 136: This assumes that someone won
There can be a tie, and should prob return null if so

TylerMathis: Agreed

src/app/abstracts/game.abstract.ts
TylerMathis: Yes in this case.

TylerMathis: I guess a better way to ask the question is.

Do the variables live in the abstract? If so, how do we get them there without passing them through super().

tanndlin: I'm need to brush up on inheritance. It's been a while since I really had to do this stuff.
Is there and issue with passing it in super?

suneettipirneni: > I guess a better way to ask the question is.
> 
> Do the variables live in the abstract? If so, how do we get them there without passing them through super().

Yeah, if you initialize/instantiate a variable in an abstract class that value is carried down to extending classes. So in the TicTacToe plugin you can simply use `this.container`

TylerMathis: Awesome this is how I thought things worked! This is how it is currently currently implemented, with necessary data living in the Abstract class, and being passed in through the constructor.

So the question still lies though, @tanndlin  what changes are you suggesting here.

suneettipirneni: I think he wants your container field to be either public or protected, as it stands right now, subclasses can't access the container field.

TylerMathis: Ah I see, I currently only have playerA and playerB as protected members, and keep everything else as private.

What is the benefit of publicizing the container? Isn't the whole point of the abstraction to keep game logic in one place and system logic in another?

suneettipirneni: In this case we aren't really concerned with encapsulation when it comes to inheritance. We are only concerned about it when it's used externally. When I say "externally" I mean as value within your code, not by extending it. Your code is encapsulated from the outside. But that same encapsulation doesn't apply for inheritance. Extending classes still have to override the abstract methods. And they have access to the protected fields. Contrast this with using an instance of `Game` externally, all of the protected fields are successfully hidden.

This is more of a template than an encapsulation. Almost every property in `Game` can be overwritten anyways just by extending it. This is fine because extending classes sometimes need that flexibility depending on their use case. But if they don't need that flexibility the `Game` class offers reasonable defaults (e.g. the non-abstract methods).

TylerMathis: Ah understood! So along that same vein, we probably should keep Game variables protected rather than fully public, as to discourage changes being made from outside of a Game instance.

What are your thoughts?

suneettipirneni: Sounds good to me!

tanndlin:
#460 got merged, so make sure to rebase or merge that to make sure your names follow the style guide
You also have conflicts

I changed the stuff for updating the leaderboard to remember to use that new code for `resultResult` since that's the one that works now

src/common/game.ts
```diff
@@ -0,0 +1,64 @@
...
+        this._game,
+        {
+          opponent: this.playerA.id,
+          result: playerResult(this.playerB),
+        }
+      )
+    ];
+
+    await Promise.all(updates);
+  }
```
tanndlin, line 48: Should check that the game actually ended before updating leaderboard. The users could have let the time run out

TylerMathis:
> #460 got merged, so make sure to rebase or merge that to make sure your names follow the style guide
> 
> You also have conflicts
> 
> 
> 
> I changed the stuff for updating the leaderboard to remember to use that new code for `resultResult` since that's the one that works now

Yup, just waiting for feedback on how we want to remove conflicts moving forward before I do anything else. :)

src/common/game.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * @returns The loser of the game
+   */
+  abstract getLoser(): User | undefined;
+
+  /**
+   * @returns Whether the game was a tie or not
+   */
+  abstract checkTie(): boolean;
```
tanndlin, line 63: do these include logic for finding the winner? Or do they just return a cached winner field?

tanndlin:
> > #460 got merged, so make sure to rebase or merge that to make sure your names follow the style guide
> > You also have conflicts
> > I changed the stuff for updating the leaderboard to remember to use that new code for `resultResult` since that's the one that works now
> 
> Yup, just waiting for feedback on how we want to remove conflicts moving forward before I do anything else. :)

That code that is in the `collector.end` that's creating the conflict. Make sure you use that or similar to `game.recordResult` since its the working code. Not sure if there was any other conflicts

TylerMathis:
> > > #460 got merged, so make sure to rebase or merge that to make sure your names follow the style guide
> 
> > > You also have conflicts
> 
> > > I changed the stuff for updating the leaderboard to remember to use that new code for `resultResult` since that's the one that works now
> 
> > 
> 
> > Yup, just waiting for feedback on how we want to remove conflicts moving forward before I do anything else. :)
> 
> 
> 
> That code that is in the `collector.end` that's creating the conflict. Make sure you use that or similar to `game.recordResult` since its the working code. Not sure if there was any other conflicts

Oh I mean in regards to rebase vs merge.

tanndlin:
> > > > #460 got merged, so make sure to rebase or merge that to make sure your names follow the style guide
> > 
> > 
> > > > You also have conflicts
> > 
> > 
> > > > I changed the stuff for updating the leaderboard to remember to use that new code for `resultResult` since that's the one that works now
> > 
> > 
> > > 
> > 
> > 
> > > Yup, just waiting for feedback on how we want to remove conflicts moving forward before I do anything else. :)
> > 
> > 
> > That code that is in the `collector.end` that's creating the conflict. Make sure you use that or similar to `game.recordResult` since its the working code. Not sure if there was any other conflicts
> 
> Oh I mean in regards to rebase vs merge.

Do whatever you are comfortable with for now. I was waiting on Corgo or Joey for a say

src/common/game.ts
TylerMathis: No, they are an abstract method that should only return the winner, if it exists.

src/common/game.ts
TylerMathis: Adding this now, thank you!

TylerMathis:
^^ These changes were me, forgot to associate an identity with myself on different laptop.

TylerMathis:
Changes have been addressed, please let me know if anything can be improved 👍

src/common/game.ts
```diff
@@ -0,0 +1,56 @@
+import { User } from 'discord.js';
+import { IContainer, Maybe } from './types';
+import { GameResult, GameType } from '../services/gameleaderboard.service';
+export default abstract class Game {
+  protected container: IContainer;
+  protected game: GameType;
+  protected playerA: User;
+  protected playerB: User;
```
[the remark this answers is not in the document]
suneettipirneni: I assumed that only private fields would use a `_` prefix, hence why the linter doesn't enforce it, but we can modify the linter later if need be.

src/app/plugins/tictactoe.plugin.ts
```diff
@@ -311,10 +265,15 @@ class TTTGame {
     this._currentPlayer *= -1;
   }
 
-  private _checkAndUpdateWin() {
+  private _updateGameState() {
     if (this._checkWin()) {
-      this._winner = this._currentPlayer;
-      this._gameOver = true;
+      this.winner = this._convertPlayerToUser(this._currentPlayer);
+      this.loser = this._convertPlayerToUser(this._currentPlayer * -1);
```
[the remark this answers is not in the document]
suneettipirneni: Perhaps we can use an enum and leverage the type system:

```typescript
enum ResultType {
  won,
  tied,
}
...

type GameResult = { type: ResultType.tied } | { type: ResultType.won, user: User }

// Method
public get result(): GameResult;

...

const result = game.result;

// Checking
switch(result.type) {
  case ResultType.won:
    // We don't need to check for `undefined` on `result.user` here, because typescript's
    // type system is smart enough to figure out that `user` can *never* be undefined when
    // when result.type == ResultType.won.
    console.log(result.user.name);
  case ResultType.tie:
    // This is compile time error, same process as above but ts knows that
    // if the result type == .tied user doesn't exist.
    console.log(result.user);
}
```

tanndlin:
Have all the comments been addressed? No rush, i just hate for you to be waiting on me

TylerMathis:
Nope! Sorry, been crazy busy lately. What direction do we want to move with this? I see there are two trains of thought on the best nice forward.

src/app/plugins/tictactoe.plugin.ts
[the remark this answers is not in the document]
tanndlin: bump

tanndlin:
Closing until this gets picked back up

PR #467, opened 2021-06-08
Convert C4 AI To Actual Minimax

TylerMathis:
C4 wasn't using real minimax before, and actually had a pretty significant bug in it's main evaluation module. This has been addressed, and now Lion is tough to beat! (but possible, remember that C4 is a solved game for white)

### DIFF
--- a/src/app/plugins/connectfour.plugin.ts
+++ b/src/app/plugins/connectfour.plugin.ts
@@ -141,8 +141,6 @@ class ConnectFourGame {
 
   private _playingLion: boolean;
   private _aiDepth = 4;
-  // Difficulty should be given in [1, 100];
-  private _aiDifficulty = 50;
 
   private _winner: Maybe<number> = null;
   private _tie: boolean = false;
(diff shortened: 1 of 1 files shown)


src/app/plugins/connectfour.plugin.ts
```diff
@@ -204,38 +202,36 @@ class ConnectFourGame {
...
   private _lionMove() {
+    const bestCol = this._getBestMove();
+    this._dropPiece(bestCol);
+  }
+
+  // Turn on the beast.
+  private _getBestMove(): number {
     const moves: { col: number; val: number }[] = [];
     for (let col = 0; col < this._cols; col++) {
       if (!this._dropPiece(col)) {
```
tanndlin, line 214: does `this._dropPiece` affect the state of the board?

tanndlin: oh, i see the `this._removeTopPiece(col);`
You should prob comment that this change gets reversed at the bottom of the function

TylerMathis: Yeah I can notate that this is a backtracking algorithm.

suneettipirneni (approved):
lgtm

tanndlin (approved):

PR #468, opened 2021-06-09
Make Fields in leaderboard inline

tanndlin:
No more cowboy coding


tanndlin:
Gotta merge #469 First

PR #470, opened 2021-06-09
Refactored how C4 got opponent

tanndlin:


Armintron (approved):
Looks good to me

TylerMathis:
What is the purpose of this? Does it make the regex mobile/desktop cross compat?

tanndlin:
> What is the purpose of this? Does it make the regex mobile/desktop cross compat?

Creating it as a string, hardcodes the length
Sometimes when someone @'s someone, there is an `!` infront, and sometimes there is not. This guarantees that it get the right ID

TylerMathis:
> > What is the purpose of this? Does it make the regex mobile/desktop cross compat?
> 
> Creating it as a string, hardcodes the length
> Sometimes when someone @'s someone, there is an `!` infront, and sometimes there is not. This guarantees that it get the right ID

Weeeeird, thank you!

TylerMathis (approved):
<3

PR #472, opened 2021-06-10
User Service

tanndlin:
What else would be useful in userService?


src/services/user.service.ts
```diff
@@ -0,0 +1,48 @@
...
+    const user = this._guildService
+      .get()
+      .members.cache.filter((member) => {
+        const { nickname } = member;
+        const { tag, username, id } = member.user;
+        return [nickname, tag, username, id].some((t) => t === target || t === strippedID);
+      })
+      .first();
+
+    return user;
```
suneettipirneni, line 27: You probably just inline this return

```suggestion
    return user = this._guildService
      .get()
      .members.cache.filter((member) => {
        const { nickname } = member;
        const { tag, username, id } = member.user;
        return [nickname, tag, username, id].some((t) => t === target || t === strippedID);
      })
      .first();
```

src/services/user.service.ts
```diff
@@ -0,0 +1,48 @@
...
+      const roleNameLower = roleName.toLowerCase();
+      return member.roles.cache.filter((r) => r.name.toLowerCase() === roleNameLower).size !== 0;
+    }
+
+    return member.roles.cache.filter((r) => r === roleName).size !== 0;
+  }
+
+  public getAgeThreshold() {
+    return this._ageThreshold;
+  }
```
suneettipirneni, line 47: Since this getter is only being used to get the private field and nothing more, you can change your `_ageThreshold` field to be `readonly public` which has the same effect, but a more concise syntax: `userService.ageThreshold` vs `useService.getAgeThreshold()`:

```typescript
public readonly ageThreshold = moment.duration(2, 'days');
```

suneettipirneni: Furthermore you can also probably make the field `static` since it's always the same for each instance of `UserService`

suneettipirneni (approved):

tanndlin:
Imma merge this, we can add more stuff when we think about it

PR #473, opened 2021-06-10
Use `Message#mentions` instead of regex for leaderboard

suneettipirneni:
Discord.js has built-in support for mentions so we can use that instead, also in Discord.js v13 using a regex for mentions will make it harder to conform to discord.js types.

### DIFF
--- a/src/app/plugins/leaderboard.plugin.ts
+++ b/src/app/plugins/leaderboard.plugin.ts
@@ -71,14 +71,14 @@ export class LeaderboardPlugin extends Plugin {
   }
 
   private _getMatchUpEmbed(
-    playerOneString: User,
-    playerTwoString: User,
+    playerOne: User,
+    playerTwo: User,
     gameEnum: GameType
   ) {
 
     return this.container.gameLeaderboardService.createMatchupLeaderboardEmbed(
-      playerOneString,
-      playerTwoString,
+      playerOne,
+      playerTwo,
       gameEnum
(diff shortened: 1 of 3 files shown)


src/app/plugins/leaderboard.plugin.ts
```diff
@@ -12,9 +12,7 @@ export class LeaderboardPlugin extends Plugin {
   public permission: ChannelType = ChannelType.Public;
   public pluginChannelName = Constants.Channels.Public.Games;
 
-  private readonly _MENTION_REGEX = /<@!?&?(\d+)>/;
-
-  public validate(message: IMessage, args: string[]) {
+  public validate(_message: IMessage, args: string[]) {
```
tanndlin, line 15: why change `message` to `_message`

suneettipirneni: In typescript you mark parameters you don't use by prefixing them with an underscore, this tells them to be skipped by the compiler.

tanndlin (approved):
I like red diffs

suneettipirneni:
> I like red diffs

Ayyyyyy can I get one more approval on my PR: 🥺👉👈 😂😂😂 🥺👉👈

TylerMathis:
I can't figure out how to approve this on mobile lol.

- [ ] Approved!

TylerMathis (approved):
KACHOW

PR #475, opened 2021-06-13
update to discord.js v13 and fix breaking changes.

suneettipirneni:
draft, pending discord.js v13 release


suneettipirneni:
I've let this get stale and outdated, if anyone is motivated to finish this lmk

PR #476, opened 2021-06-14
Add more eslint rules

suneettipirneni:
Adds enforcement for semicolons, single quote strings (where applicable) and errors for unused variables.


.eslintrc.json
```diff
@@ -14,8 +14,16 @@
...
     "prefer-const": "error",
+    // note you must disable the base rule as it can report incorrect errors
+    "semi": "off",
+    "@typescript-eslint/semi": ["error"],
     "space-before-blocks": "error",
     "spaced-comment": "warn",
+    "quotes": "off",
+    "@typescript-eslint/quotes": ["error", "single", { "avoidEscape": true }],
+    // note you must disable the base rule as it can report incorrect errors
+    "no-unused-vars": "off",
```
tanndlin, line 25: That's odd, bc i swear we had this rule before, but i dont see it in the current eslint.json

tanndlin (approved):
I like this because we had prettier rules for this, but sometimes a new contributor doesnt have prettier.

src/test.ts
```diff
@@ -1,7 +0,0 @@
-class MyClass {
-  private _myVar = 'hello';
-
-  foo() {
-    console.log(this._myVar)
-  }
-}
```
tanndlin, line 7: Whats this about?

tanndlin (approved):

src/test.ts
suneettipirneni: It was some leftover testing code that leaked through

PR #477, opened 2021-06-14
Use `Message#mentions` instead of regex in `LionPingHandler`

suneettipirneni:
Title says all.

### DIFF
--- a/src/app/handlers/lionping.handler.ts
+++ b/src/app/handlers/lionping.handler.ts
@@ -2,7 +2,6 @@ import { EmojiIdentifierResolvable } from 'discord.js';
 import { IContainer, IHandler, IMessage, Maybe } from '../../common/types';
 
 export class LionPingHandler implements IHandler {
-  public mentionRegex: RegExp = /<@!?\d+>/g;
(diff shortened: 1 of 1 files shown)


src/app/handlers/lionping.handler.ts
```diff
@@ -13,23 +12,17 @@ export class LionPingHandler implements IHandler {
     }
 
     if (!this._reactEmoji) {
-      this._reactEmoji =
-        message.guild?.emojis.cache.filter((e) => e.name === 'lion').first() || '👍';
+      this._reactEmoji = message.guild?.emojis.cache
+        .filter((e) => e.name === 'lion')
+        .first() ?? '👍';
```
tanndlin, line 17: What is `||` vs `??`

suneettipirneni: `??` Is the null coalescing operator, it's preferred over using || for checking for undefined or null, because || can give false positives for anything that evaluates to false. For example if an empty string is returned `||` will think the left hand side is false. `??` Would evaluate to true, which is the expected behavior because `""` isn't null or undefined.

tanndlin (approved):

c650 (approved):

PR #478, opened 2021-06-14
Don't allow people to play themselves in games.

suneettipirneni:
This breaks the leaderboard, because you get not 1, but 2 wins if you beat *yourself* in a game. I also refactored some areas that were still using mention regex.


tanndlin (approved):

Armintron (approved):
Looks good to me

PR #482, opened 2021-06-16
Check for everyone ping and add code plugin

### LINKED ISSUE #481: Check for everyone ping in command handler

tanndlin:
Fixes #481

### DIFF
--- a/src/app/plugins/code.plugin.ts
+++ b/src/app/plugins/code.plugin.ts
@@ -39,9 +39,11 @@ export class CodePlugin extends Plugin {
       .fetch(messageID)
       .then((targMessage) => {
 
+        console.log(targMessage.mentions);
+
         // Check if the target language has '@everyone' in it.
         if (targMessage.content.match(/@everyone/) != null) {
-          message.channel.send("Please omit '@everyone' from you code snippet.")
+          message.channel.send("Please don't ping everyone from you code snippet.")
           return;
         }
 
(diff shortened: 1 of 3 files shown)


tanndlin:
Can bring back code plugin after this
Though, I still feel like it needs to be revamped or something

suneettipirneni:
This may work for other plugins but it won't work for the code plugin, because it references another message

tanndlin:
> This may work for other plugins but it won't work for the code plugin, because it references another message

You right

suneettipirneni:
> I dont think this works bc it doesnt show up as @everyone.
> It's like a normal mention `<@4085743750>`
> 
> You can just do message.mentions?.everyone

I tested it, and it prevented the code plugin from repeating `@everyone`. The reason is because the code formatting converts everything into plain text, so transforming mentions like, `<@4085743750>` isn't a thing that discord does.

src/app/plugins/code.plugin.ts
```diff
@@ -39,9 +39,11 @@ export class CodePlugin extends Plugin {
       .fetch(messageID)
       .then((targMessage) => {
 
+        console.log(targMessage.mentions);
+
         // Check if the target language has '@everyone' in it.
         if (targMessage.content.match(/@everyone/) != null) {
```
tanndlin, line 45: `if(!targMessage.content.match(/@everyone/))`

src/app/plugins/code.plugin.ts
```diff
@@ -39,9 +39,11 @@ export class CodePlugin extends Plugin {
       .fetch(messageID)
       .then((targMessage) => {
 
+        console.log(targMessage.mentions);
```
tanndlin, line 42: left over log

suneettipirneni (approved):

PR #487, opened 2021-06-18
Create admin commands for enabling/disabling plugins

### LINKED ISSUE #484: Admin command for enable/disabling a plugin

suneettipirneni:
closes #484

### DIFF
--- a/src/app/plugins/controller.plugin.ts
+++ b/src/app/plugins/controller.plugin.ts
@@ -37,7 +37,7 @@ export class PluginControl extends Plugin {
 
     try {
       await pluginStateData
-        ?.updateOne({ name:  plugin.name }, 
+        .updateOne({ name:  plugin.name }, 
           { $set: { isActive: method === 'activate' }},
           { upsert: true });
     } catch(error) {
--- a/src/app/plugins/controller.plugin.ts
+++ b/src/app/plugins/controller.plugin.ts
@@ -1,10 +1,10 @@
-import { Plugin } from "../../common/plugin";
-import { IContainer, ChannelType, IMessage } from "../../common/types";
+import { Plugin } from '../../common/plugin';
+import { IContainer, ChannelType, IMessage } from '../../common/types';
 
 export class PluginControl extends Plugin {
   public name: string = 'controller';
   public description: string = 'Controls activating and deactivating plugins.';
-  public usage: string = "!controller <activate | deactive> <plugin name>";
+  public usage: string = '!controller <activate | deactive> <plugin name>';
   public permission: ChannelType = ChannelType.Admin;
   public commandPattern: RegExp = /^(deactivate|activate) (?!\s*$).+/;
 
@@ -36,7 +36,7 @@ export class PluginControl extends Plugin {
     }
 
     try {
-      const result = await pluginStateData
+      await pluginStateData
         ?.updateOne({ name:  plugin.name }, 
           { $set: { isActive: method === 'activate' }},
           { upsert: true });
--- a/src/app/plugins/controller.plugin.ts
+++ b/src/app/plugins/controller.plugin.ts
@@ -20,16 +20,22 @@ export class PluginControl extends Plugin {
 
     const [method, pluginName] = args;
 
-    switch (method) {
-      case 'deactivate':
-        this.container.pluginService.setPluginActive(pluginName, false);
-        break;
-      case 'activate':
-        this.container.pluginService.setPluginActive(pluginName, true);
-        break;
-      default:
-        await message.channel.send(`${method} is invalid, possible options are activate or deactivate.`)
+    try {
+      switch (method) {
+        case 'deactivate':
+          this.container.pluginService.setPluginActive(pluginName, false);
+          break;
+        case 'activate':
+          this.container.pluginService.setPluginActive(pluginName, true);
+          break;
+        default:
+          await message.channel.send(`${method} is invalid, possible options are activate or deactivate.`)
+      }
+    } catch(e) {
+      await message.channel.send(e.message);
+      return;
     }
+
+    message.channel.send(`${pluginName} has been ${method}d`);
   }
-    
 }
--- a/src/app/plugins/controller.plugin.ts
+++ b/src/app/plugins/controller.plugin.ts
@@ -4,7 +4,7 @@ import { IContainer, ChannelType, IMessage } from '../../common/types';
 export class PluginControl extends Plugin {
   public name: string = 'controller';
   public description: string = 'Controls activating and deactivating plugins.';
-  public usage: string = '!controller <activate | deactive> <plugin name>';
+  public usage: string = 'controller <activate | deactive> <plugin name>';
   public permission: ChannelType = ChannelType.Admin;
   public commandPattern: RegExp = /^(deactivate|activate) (?!\s*$).+/;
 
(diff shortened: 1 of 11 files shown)


src/app/plugins/controller.plugin.ts
```diff
@@ -0,0 +1,41 @@
...
+  public description: string = 'Controls activating and deactivating plugins.';
+  public usage: string = "!controller <activate | deactive> <plugin name>";
+  public permission: ChannelType = ChannelType.Admin;
+
+  constructor(public container: IContainer) {
+    super();
+  }
+
+
+  public async execute(message: IMessage, args?: string[]): Promise<void> {
```
tanndlin, line 15: Remove the `?` args is always defined, just not always populated, that way you dont have to check if args exists

src/app/plugins/controller.plugin.ts
```diff
@@ -0,0 +1,41 @@
...
+    if (!args || args!.length < 2) {
+      await message.channel.send(`Usage: ${this.usage}`);
```
tanndlin, line 17: Override validate for this `validate`.
When that returns false it'll take care of sending the usage

src/services/plugin.service.ts
```diff
@@ -81,4 +82,35 @@ export class PluginService {
       return page;
     });
   }
+
+  public setPluginActive(plugin: Plugin | string, active: boolean) {
```
tanndlin, line 86: Since you are only using the plugin name, it would be easier to just make the parameter a string. Especially since the only way this is called is through user input, which is already a string

src/app/plugins/controller.plugin.ts
```diff
@@ -0,0 +1,41 @@
...
+        case 'activate':
+          this.container.pluginService.setPluginActive(pluginName, true);
+          break;
+        default:
+          await message.channel.send(`${method} is invalid, possible options are activate or deactivate.`)
+      }
+    } catch(e) {
+      await message.channel.send(e.message);
+      return;
+    }
```
tanndlin, line 37: Could just do 
```
this.container.pluginService.setPluginActive(pluginName, method === 'activate');
```
You could also enforce only `activate` and `deactivate` with the `public commandPattern`

src/services/plugin.service.ts
```diff
@@ -81,4 +82,35 @@ export class PluginService {
...
+    let myPlugin;
+
+    if (!(typeof plugin === 'string')) {
+      myPlugin = plugin;
+    } else {
+      let fetchedPlugin = this.plugins[plugin];
```
tanndlin, line 92: aliases map to the actual name (and the name itself is an alias)
so you can do 
```
const fetchedPlugin = this.plugins[this.aliases[plugin]]
```

src/services/plugin.service.ts
joey-colon: not really a fan of this name. can we do something like `setPluginState` or `setPluginStatus` ?

joey-colon:
can we persist our overall plugin state to db so we don't lose it to on bot restart?

joey-colon:
> can we persist our overall plugin state to db so we don't lose it to on bot restart?

and so we can hydrate on bot startup with the last state from db

tanndlin:
> can we persist our overall plugin state to db so we don't lose it to on bot restart?

I was thinking about that, would not be hard to put in the db

src/app/plugins/controller.plugin.ts
```diff
@@ -6,31 +6,22 @@ export class PluginControl extends Plugin {
...
   public usage: string = "!controller <activate | deactive> <plugin name>";
   public permission: ChannelType = ChannelType.Admin;
+  public commandPattern: RegExp = /^(deactivate|activate) (?!\s*$).+/;
 
   constructor(public container: IContainer) {
     super();
   }
 
+  public validate(message: IMessage, args: string[]) {
+    return (args && args.length >= 2 && this.commandPattern.test(args.join(' ')));
```
tanndlin, line 16: no need to override validate if you use the command pattern. Sorry for the conflicting reviews
because of the commandPattern, it takes care of the length too

src/services/plugin.service.ts
```diff
@@ -3,26 +3,51 @@ import { PluginLoader } from '../bootstrap/plugin.loader';
...
 
   private readonly _NUM_DISPLAY = 10;
 
+  public async initPluginState(container: IContainer): Promise<void> {
+    if (this._pluginState) {
+      return;
+    }
+
+    const stateCollection = (await container.storageService.getCollections()).pluginState;
+    this._pluginState = await stateCollection?.find().toArray();
```
tanndlin, line 24: nit: not a great name for a collection

src/app/plugins/controller.plugin.ts
```diff
@@ -13,20 +13,40 @@ export class PluginControl extends Plugin {
...
+    // Save data in persistently.
+    const pluginStateData = (await this.container.storageService.getCollections()).pluginState;
+    if (!pluginStateData) {
+      message.channel.send('Error connecting to the DB');
+      return;
+    }
+
+    try {
+      await pluginStateData
+        .updateOne({ name:  plugin.name },
```
tanndlin, line 40: scope it by guild ID too, we have test servers on the same collection and scope by guildID

src/services/plugin.service.ts
```diff
@@ -82,7 +107,7 @@ export class PluginService {
     });
   }
 
-  public setPluginState(plugin: string, active: boolean) {
+  public setPluginState(plugin: string, active: boolean): IPlugin {
```
tanndlin, line 110: I don't like that this function fetches and sets the state of a plugin. You should prob make another function to fetch a plugin, then a `void` function to `setPluginState`

suneettipirneni: The reason that this function is wrapped like this is because, whenever a state change occurs in plugin, it also needs to be saved. This isolates plugins from having to manually save plugin states. Basically, the business logic is abstracted out.

suneettipirneni: My bad, that's actually not the case, I'll go ahead move the save logic into there though.

src/services/plugin.service.ts
```diff
@@ -3,12 +3,37 @@ import { PluginLoader } from '../bootstrap/plugin.loader';
...
+    // Set all of the plugins to the persisted state.
+    Object.values(this.plugins).forEach(plugin => {
+      fetchedStates.forEach(state => {
+        if (state.name === plugin.name && state.guildID === container.guildService.get().id) {
+          plugin.isActive = state.isActive;
+        }
+      });
+    });
+  }
+
```
c650, line 36: mmm persistence!!

c650 (approved):

src/app/plugins/controller.plugin.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  public async execute(message: IMessage, args: string[]): Promise<void> {
+    const [method, pluginName] = args;
+
+    try {
+      await this.container.pluginService.setPluginState(this.container, pluginName, method === 'activate') ;
+    } catch(e) {
+      await message.channel.send(e.message);
+      return;
+    }
```
tanndlin, line 23: instead of try catch, just do `this.container.pluginService.setPluginState(this.container, pluginName, method === 'activate').catch((e) => ...`

src/services/plugin.service.ts
```diff
@@ -3,12 +3,37 @@ import { PluginLoader } from '../bootstrap/plugin.loader';
...
+    }
+
+    const fetchedStates = await pluginStates.find().toArray();
+
+    // Set all of the plugins to the persisted state.
+    Object.values(this.plugins).forEach(plugin => {
+      fetchedStates.forEach(state => {
+        if (state.name === plugin.name && state.guildID === container.guildService.get().id) {
+          plugin.isActive = state.isActive;
+        }
```
tanndlin, line 32: I would prefer to either filter the plugins with guildID or better yet, pass in the guildID for `.find()` from mongo

tanndlin (commented):
What is the default state in the DB? is the collection populated automatically with all the plugins or is it filled in as the controller plugin is called?

suneettipirneni:
> What is the default state in the DB? is the collection populated automatically with all the plugins or is it filled in as the controller plugin is called?

It's populated as the controller is called. All plugins default to being active so if an entry in the DB isn't found it is loaded in as active.

tanndlin:
> > What is the default state in the DB? is the collection populated automatically with all the plugins or is it filled in as the controller plugin is called?
> 
> It's populated as the controller is called. All plugins default to being active so if an entry in the DB isn't found it is loaded in as active.

Ok cool

src/app/plugins/controller.plugin.ts
suneettipirneni: The issue here is that I want to return from the function after I catch an error, this isn't possible with the `.catch()` syntax.

tanndlin (approved):

PR #488, opened 2021-06-19
Automated Testing

tanndlin:
Not really sure the best way to make this code so that you can have your own tests and be ignored by git


src/test/common/pluginTester.ts
```diff
@@ -0,0 +1,11 @@
+import { IMessage } from '../../common/types';
+
+export abstract class PluginTester {
+  public abstract channelName: string;
+  public abstract args: string;
+
+  constructor() {}
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public onResponse(message: IMessage): void {}
```
suneettipirneni, line 10: I think this should be marked as `abstract` because it seems like a requirement. I believe this would also prevent you from having to suppress the unused-var error.

tanndlin: its not required, lots of plugins wouldn't need something done to the result such as a react or something. Most of the time i was using this, i didnt need the `onReponse`

suneettipirneni:
```typescript
if (!testCase.onResponse) {
      return;
} 
```

I see this code, but shouldn't this never be false since `onResponse` is always inherited?

tanndlin: oh yeah, i think its left over. I had tried
```ts
public abstract onReponse? (...)
```
but didn't get that to work

suneettipirneni: Now that I think of it PluginTester is better off just as an interface. Since the base class doesn't contain any business logic. You can set your `onResponse` type to `onResponse?: (message: IMessage) => void`

tanndlin: That's a good idea. Thank you :)

suneettipirneni: Ok cool, correct me if I'm wrong but the ts error suppression comment isn't needed anymore, right?

src/test/tester.ts
```diff
@@ -0,0 +1,66 @@
+import { Client, TextChannel } from 'discord.js';
+import moment from 'moment';
+import { IMessage } from '../common/types';
+import Environment from '../environment';
+import { CASES } from './bootstrap/tester.loader';
+import { PluginTester } from './common/pluginTester';
```
suneettipirneni, line 6: CircleCI seems to be failing because of this. Should be `IPluginTester`.

suneettipirneni (changes requested):

c650 (approved):
so this is for integration tests?

tanndlin:
> so this is for integration tests?

Correct, requires a separate bot token, to send a message mimicking a user. Just need to run `npm run test` in another terminal

tanndlin:
This is broke from #511 
And idk how to get the token to work now because its in a separate folder

tanndlin:
Reverted [this commit](https://github.com/cs-discord-at-ucf/lion/commit/a248fb0b7e8c0c7a271e928e91bbf80ad53c2470) because it broke this PR, and I feel it was very unnecessary to begin with.

suneettipirneni:
> Reverted [this commit](https://github.com/cs-discord-at-ucf/lion/commit/a248fb0b7e8c0c7a271e928e91bbf80ad53c2470) because it broke this PR, and I feel it was very unnecessary to begin with.

That PR didn't change any of the logic behind the env vars, so it shouldn't be affecting this PR.

tanndlin:
> > Reverted [this commit](https://github.com/cs-discord-at-ucf/lion/commit/a248fb0b7e8c0c7a271e928e91bbf80ad53c2470) because it broke this PR, and I feel it was very unnecessary to begin with.
> 
> That PR didn't change any of the logic behind the env vars, so it shouldn't be affecting this PR.

I get `INVALID_TOKEN` error and cant seem to pass it in. It was more straightforward before that PR

suneettipirneni:
Ok let me see if I can get it working, the only difference between my PR on the previous code is process.env.{token} instead of Environment.{token}

suneettipirneni:
You're not loading in the environment vars when you run the test, it's a different entrypoint to `npm run watch`. If you change the line in your package.json:

from
```json
"test:serve": "node -r ts-node/register --inspect src/test/tester.ts",
```
to
```json
"test:serve": "node -r ts-node/register -r dotenv/config --inspect src/test/tester.ts",
```

It should work

tanndlin:
Merging for now, works fine as is
We can tweak it as we see fit

PR #495, opened 2021-06-21
Add `--noImplicitOverride` compiler option

suneettipirneni:
This adds compiler errors if inherited properties aren't marked with `override`. 

### Why is this needed?
If you want to view the formal spec on this you visit the [TS docs](https://www.typescriptlang.org/tsconfig#noImplicitOverride).

tldr; 
> When working with classes which use inheritance, it’s possible for a sub-class to get “out of sync” with the functions it overloads when they are renamed in the base class.

This compiler option prevents confusion of either:
1) unknowingly overriding an inherited property
2) thinking you're overriding an inherited property that has since been refactored.

In addition it provides a visual identifier of what is inherited and what isn't which is nice for newer contributors.


tanndlin:
Need to fix build and conflicts and what not
There are so many commits in here I have no idea where to look in the diff

suneettipirneni:
> Need to fix build and conflicts and what not
> There are so many commits in here I have no idea where to look in the diff

Ignore the diff, it's a product of the git history rewrite we did last night. As for the diff, its just basically adding `override` to inherited properties.

src/app/plugins/addclasschans.plugin.ts
```diff
@@ -12,7 +12,7 @@ export class AddClassChannelsPlugin extends Plugin {
   public name: string = 'Add many classes';
   public description: string = 'creates a bunch of class channels';
   public usage: string = 'addclasschans';
-  public pluginAlias = [];
+  public override pluginAlias = [];
   public permission: ChannelType = ChannelType.Admin;
```
tanndlin, line 16: Why is only the alias an override for all plugins?

suneettipirneni: Because a lot of contributors have been overriding the default value given in the base class. The other fields aren't marked with override because they're required.

src/app/jobs/inactivevoice.job.ts
```diff
@@ -2,14 +2,14 @@ import { IContainer } from '../../common/types';
...
-  public name: string = 'Inactive Voice';
+  public override interval: number = 1000 * 60 * 10; // Every 10 mintues
+  public override name: string = 'Inactive Voice';
 
   constructor() {
     super();
   }
 
-  public async execute(container: IContainer) {
+  public override async execute(container: IContainer) {
```
tanndlin, line 12: `interval`, `name`, and `execute` should all be required in a job

suneettipirneni: > `interval`, `name`, and `execute` should all be required in a job

They aren't, `job.ts` provides default values for all of those fields:
```typescript
export abstract class Job implements IJob {
  public name: string = '';
  public interval: number = 60000;
  public execute(container?: IContainer) {}
}
```

tanndlin (approved):

suneettipirneni:
Is this good to go? it's getting stale

suneettipirneni:
Oh I didn't see the merge conflicts, I'll take of those

suneettipirneni:
Can I get another review pretty please?

tanndlin:
@joey-colon @c650 whats your opinions

suneettipirneni:
This PR is also getting stale, I would like to get another review 🥺

Armintron (approved):
Very clever rule. I likey

c650:
> ... @c650 whats your opinions

compiler rule more strict -- find w me if it catches bugs

c650 (approved):
approving this but not actually reading it because others already did

PR #499, opened 2021-06-23
New ESlint Rule Party

suneettipirneni:
This adds *6* new rules to the linter, I'll explain why I added each of them:

### `@typescript-eslint/await-thenable`:
Enforces people to only use `await` on `async` expressions, and not mistakenly use it for synchronous expressions.

### `@typescript-eslint/no-extra-non-null-assertion`:
Prevents statements like these: 
```typescript
return bar!?.n
```

### `@typescript-eslint/no-unnecessary-type-assertion`:
Prevents unnecessary casting, and/or null assertions if it a given property is known to be a certain type, or non-nullish:
```typescript
function foo(bar: string): string {
  return foo! + "bar"; // Error: bar is known to be non-nullish so the `!` operator doesn't do anything.
}
```

### `@typescript-eslint/prefer-nullish-coalescing`:
This is one of the more important rules, it prevents people from using `||` for nullish checks and prefers `??` instead:
```typescript
myValue: string | undefined;
...
// Wrong
const output = myValue || '';

// Correct
const output = myValue ?? '';
```
### `eol-last`:
Enforces a blank line at the end of every file.

### `@typescript-eslint/require-await`:
A lot of our functions were marked as `async` even though they didn't have any `await` statements. This rule prevents this common mistake from happening.
```typescript
// Incorrect
async function foo(): Promise<string> {
  return "bar";
}

// Correct
function foo(): string {
  return "bar";
}
```
___
**As per usual any of these rules can be commented and added and removed during the review process if need be.**


tanndlin:
same as the other one, there are so many commits, the diff is a mess

suneettipirneni:
> same as the other one, there are so many commits, the diff is a mess

As with my other pr's git history is borked bc of the squash thing last night.

tanndlin (approved):

Armintron (approved):
Looks good to me

PR #503, opened 2021-06-25
Create Weather Events Warning Job

Clev7:


src/app/jobs/weatherevents.job.ts
```diff
@@ -0,0 +1,83 @@
...
+  SAME?: (string)[] | null;
+  UGC?: (string)[] | null;
+}
+export interface Parameters {
+  PIL?: (string)[] | null;
+  NWSheadline?: (string)[] | null;
+  BLOCKCHANNEL?: (string)[] | null;
+  VTEC?: (string)[] | null;
+  eventEndingTime?: (string)[] | null;
+}
```
tanndlin, line 83: Try to prune as many of these fields you don't actually need/use

src/app/jobs/weatherevents.job.ts
```diff
@@ -0,0 +1,83 @@
...
+  public async execute(container: IContainer) {
+    const channel = (container.guildService.getChannel("weather_events")) as TextChannel;
+    const resp = await container.httpService.get("https://api.weather.gov/alerts/active?area=FL");
+    const data: Weather_Event = resp.data;
+
+    if (!data.features) {
+      return;
+    }
+
+    const map: string[] = data.features.map(f => f.properties.headline);
```
tanndlin, line 24: You should filter by severities, we dont need to know every time there is a little bit of thunder

Clev7: That's true. I've tried the API on quite a few states and it seems that the severity level can be one of 'Minor', 'Moderate', 'Severe', 'Extreme'. It can also be 'Unknown', but I think that's just because the weather's still developing at that point. I think Moderate is when it starts catching thunderstorms and past that things get worse. Should probably draw the line on Severe or Extreme.

Clev7:
I also cleared up a lot of the interface clutter.

Clev7:
Should we add a role for those who want to be pinged on severe or extreme weather?

src/app/jobs/weatherevents.job.ts
```diff
@@ -17,67 +18,31 @@ export class WeatherEventsJob extends Job {
...
 
-    const map: string[] = data.features.map(f => f.properties.headline);
+    const relevantData: FeaturesEntity[] = data.features.filter(rD => {
+      const highSeverity = rD.properties.severity === 'Severe' 
+                          || rD.properties.severity === 'Extreme';
 
-    if (map.length == 0) {
-      return;
-    }
+      return highSeverity && rD.properties.headline.includes("Orlando")
```
tanndlin, line 29: Is there not a field that tells the location, rather than looking through the headline?

tanndlin (commented):
Can you put pic of what it looks like into this PR?
We should probably make it look nice with an embed rather than plain text

src/app/jobs/weatherevents.job.ts
```diff
@@ -0,0 +1,48 @@
...
+  features?: (FeaturesEntity)[] | null;
+}
+
+export interface FeaturesEntity {
+  properties: Properties;
+}
+export interface Properties {
+  severity: string;
+  headline: string;
+}
```
suneettipirneni, line 48: Is your branch up to date with our develop branch? It should be giving you a linting error for not prefixing interface names with "I"

src/app/jobs/weatherevents.job.ts
Clev7: After taking a second look through some JSONs, senderName seems to help with that too. It usually says something like NWS <city>. In all of the headlines I've seen, the sendername is always mentioned in the headline too, but for correctness sake, I'll use the senderName instead since that makes more sense.

src/app/jobs/weatherevents.job.ts
Clev7: I did git pull inside the repo a couple of days ago before I worked on it. Not sure why it's not working past that I'm not working on the develop branch. I'll change it manually, but can you elaborate on this further?

suneettipirneni: Oh, your branch should be based from `develop` not `master.`. Your PR should also be edited to be merging into `develop`.

Clev7:
Side note, I think using only the headline can be a little problematic. At 2:49AM, when I was testing this on Jacksonville instead of Orlando, I got duplicate headlines for warnings with slightly different descriptions. If you follow the descriptions for each of the JSON objects from bottom to top at [](https://api.weather.gov/alerts/active?area=FL) , you'll see that the same warning can get multiple updates over the course of 24 hours, slightly changing the description and headlines accordingly.

I think we should either make one message that shows all of the updates in the order that they were issued, or we find a way to only get only the most recent updates for the same warning at a time.

src/app/jobs/weatherevents.job.ts
Clev7: Oh I see. I changed it to develop.

Clev7: It says it can't find definitions for the linting rules.
image

src/app/jobs/weatherevents.job.ts
Clev7: This is what it looks like for one ping to the API. Here, you can see the issue that I described below about duplicate headlines and what not.
image

src/app/jobs/weatherevents.job.ts
```diff
@@ -0,0 +1,53 @@
...
+                          || rD.properties.severity === 'Extreme';
+
+      return highSeverity && rD.properties.senderName.includes("Orlando")
+    });
+
+    const relevantHeadlines: string[] = relevantData.map(rD => rD.properties.headline)
+
+    // The warnings are automatically sorted from most to least recent, which
+    // is why I reversed it.
+    channel.send(relevantHeadlines.reverse().join("\n")).catch(err => console.log(err));
```
joey-colon, line 38: you'll need to `await` this call in order for your `catch` to be activated (if the call decides to throw).

src/app/jobs/weatherevents.job.ts
```diff
@@ -0,0 +1,53 @@
...
+import { Job } from '../../common/job';
+import { IContainer } from '../../common/types';
+import { TextChannel } from 'discord.js';
+
+export class WeatherEventsJob extends Job {
+  // Run once an hour
+  public interval: number = 1000 * 60 * 60;
+
+  // For debugging purposes only
+  // public interval: number = 1000 * 10;
```
joey-colon, line 11: remove

src/app/jobs/weatherevents.job.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export interface IFeaturesEntity {
+  properties: IProperties;
+}
+export interface IProperties {
+  severity: string;
+  headline: string;
+  senderName: string;
+}
```
joey-colon, line 53: is there a reason we want to export these interfaces? if not, we shouldn't be.

we also tend to include types at the top of the file (right below the imports)

src/app/jobs/weatherevents.job.ts
```diff
@@ -0,0 +1,53 @@
...
+  public async execute(container: IContainer) {
+    const channel = (container.guildService.getChannel("weather_events")) as TextChannel;
+    const resp = await container.httpService.get("https://api.weather.gov/alerts/active?area=FL");
+    const data: IWeather_Event = resp.data;
+
+    if (!data.features || data.features.length == 0) {
+      return;
+    }
+
+    const relevantData: IFeaturesEntity[] = data.features.filter(rD => {
```
joey-colon, line 27: we don't really have a good reason for defining the `relevantData` variable name. i would suggest changing the variable name to `headlines` and chaining your `.map` call on line 34 to this chain of operations

also, for your filter function, naming it `rD` does not make sense. can you name it as `feature` ?

src/app/jobs/weatherevents.job.ts
```diff
@@ -0,0 +1,53 @@
...
+      const highSeverity = rD.properties.severity === 'Severe' 
+                          || rD.properties.severity === 'Extreme';
+
+      return highSeverity && rD.properties.senderName.includes("Orlando")
+    });
+
+    const relevantHeadlines: string[] = relevantData.map(rD => rD.properties.headline)
```
joey-colon, line 34: variable naming, see above

tanndlin:
Resolve your comments when you take care of them :)
I would do it for you, but i dont know which ones are resolved

src/app/jobs/weatherevents.job.ts
Clev7: Yeah there's no real reason to do so, since we're really only using them here. I'll put them on the top.

src/app/jobs/weatherevents.job.ts
Clev7: Sure thing!

src/app/jobs/weatherevents.job.ts
Clev7: Ah right that's true. I didn't realize it returned a promise. I'll await it then.

src/app/jobs/weatherevents.job.ts
```diff
@@ -0,0 +1,51 @@
...
+
+interface IProperties {
+  severity: string;
+  headline: string;
+  senderName: string;
+}
+
+export class WeatherEventsJob extends Job {
+  // Run once an hour
+  public interval: number = 1000 * 60 * 60;
```
tanndlin, line 21:
```suggestion
  public interval: number = moment.duration(1, 'hours').asMilliseconds();
```
More readable and does calculation for you

src/app/jobs/weatherevents.job.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  public name: string = 'weather_events';
```
tanndlin, line 23: Put the name above the interaval, and just use normal english lol
```
'Weather Events'
```

src/app/jobs/weatherevents.job.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  constructor() {
+    super();
+  }
+
+  public async execute(container: IContainer) {
+    const channel = (container.guildService.getChannel("weather_events")) as TextChannel;
+    const resp = await container.httpService.get("https://api.weather.gov/alerts/active?area=FL");
```
tanndlin, line 31: These should be single quotes, do you have `prettier` installed?

src/app/jobs/weatherevents.job.ts
Clev7: Yeah I like that a lot better. If I apply it, though, I get an error saying "moment refers to a UMD global, but the current file is a module. Consider adding an import instead".

src/app/jobs/weatherevents.job.ts
Clev7: I do, but for some reason, it's not formatting on save for me.

src/app/jobs/weatherevents.job.ts
Clev7: Will do!

src/app/jobs/weatherevents.job.ts
Clev7: oh I just realized I just need to import it.

Clev7: On second thought, I'm not sure if we want to import momentjs just to do this since moment is a very bloated library, but if we're using it somewhere else already then I have no issues with that.

src/app/jobs/weatherevents.job.ts
```diff
@@ -17,18 +17,16 @@ interface IProperties {
...
 
   constructor() {
     super();
   }
 
   public async execute(container: IContainer) {
-    const channel = (container.guildService.getChannel("weather_events")) as TextChannel;
-    const resp = await container.httpService.get("https://api.weather.gov/alerts/active?area=FL");
+    const channel = (container.guildService.getChannel('weather_events')) as TextChannel;
+    const resp = await container.httpService.get('https://api.weather.gov/alerts/active?area=FL');
```
tanndlin, line 29: Store this link as a class constant
`private _ENDPOINT: string = '...'`

Clev7:
Hmm it worked on my end, but it seems that we don't have ms in our pacakge.json?

src/app/jobs/weatherevents.job.ts
Clev7: Got it!

tanndlin (approved):
I think this is poggers

src/app/jobs/weatherevents.job.ts
```diff
@@ -0,0 +1,55 @@
...
+  severity: string;
+  headline: string;
+  senderName: string;
+}
+
+export class WeatherEventsJob extends Job {
+  public name: string = 'Weather Events';
+  public interval: number =  ms('1h');
+
+  private _ENDPOINT: string = 'https://api.weather.gov/alerts/active?area=FL';
```
suneettipirneni, line 23: this should be `readonly static`

suneettipirneni (changes requested):

src/app/jobs/weatherevents.job.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  constructor() {
+    super();
+  }
+
+  public async execute(container: IContainer) {
+    const channel = (container.guildService.getChannel('weather_events')) as TextChannel;
+    const resp = await container.httpService.get(this._ENDPOINT);
```
Armintron, line 31: Arent we stepping away from using `httpService`? I might be wrong on this.

suneettipirneni: In this context either way it doesn't matter. My PR needs to be split down before any changes can be enforced.

src/app/jobs/weatherevents.job.ts
Clev7: Are you referring to all three fields or just the last one?

src/app/jobs/weatherevents.job.ts
Clev7: In the event that that change can be enforced, someone can comment on this PR or open an issue about it and I'll change it back to what it needs to be.

src/app/jobs/weatherevents.job.ts
Clev7: I can't make name static since the Function object's prototype also contains a name and would cause a conflict, so I'll only make it readonly. I can make the others static and readonly.

Clev7: If we're going by that paradigm, though, then wouldn't it have to be implemented across the other jobs?

suneettipirneni: I'm referring to only changing `_ENDPOINT` to a static field since it doesn't change across any instance of a plugin.

Clev7: gotcha.

Clev7: made it to where only _ENDPOINT has static readonly so it's not confusing.

tanndlin (approved):

suneettipirneni (approved):
Lgtm, just need to resolve merge conflicts.

PR #505, opened 2021-06-25
Only allow wolfram alpha and equation plugin in channels under #help

Armintron:


src/app/plugins/equation.plugin.ts
```diff
@@ -19,6 +18,11 @@ export class EquationPlugin extends Plugin {
   }
 
   public async execute(message: IMessage, args: string[]): Promise<void> {
+    // Only allow in help channels
+    if (((message.channel as TextChannel).parent as CategoryChannel).name !== 'Help') {
```
tanndlin, line 22: Are you sure the capitalization is correct?
Should probably do `name.toLowerCase() !== 'help'`

src/app/plugins/wolframalpha.plugin.ts
```diff
@@ -24,6 +24,12 @@ export class WolframAlphaPlugin extends Plugin {
   }
 
   public async execute(message: IMessage, args: string[]) {
+    // Only allow in help channels
+    if (((message.channel as TextChannel).parent as CategoryChannel).name !== 'Help') {
```
tanndlin, line 28: saa

suneettipirneni (changes requested):
It would be preferable if a new `ChannelType` like `ChannelType.Help` was introduced so it could used in other plugins down the line.

tanndlin:
> It would be preferable if a new `ChannelType` like `ChannelType.Help` was introduced so it could used in other plugins down the line.

I thought of that too. Help should be under public as a subset. This PR would need a rename though. Out of scope for this as it stands

Armintron:
Ill leave this PR as is and make another for making channel subsets

tanndlin:
I've started coding this, and changing `Constants.ts` is just not worth the added complexity
We can just make a function to compare a string vs channel parent name if anything

tanndlin:
You can now set a `pluginCategoryName` as of #519
That's all you need to do to solve this PR

PR #509, opened 2021-06-26
Automatic plugin loading

suneettipirneni:
This allows for automatic plugin loading from the `plugins/` folder. This removes the previously needed step of adding your plugin manually to the `plugin.loader.ts`. 

Notable changes:
- new `Plugin` property, `commandName`, this property is the base command name the plugin uses.
- In order for plugins to be automatically loaded, they will need to be default exports.


src/app/bot.ts
```diff
@@ -36,18 +36,33 @@ export class Bot {
...
+            this.container.loggerService.error(`${file} has a default export, but it is not of type Plugin`);
+            return;
+          }
+        } catch(err) {
+          this.container.loggerService.warn(`${file} doesn't have a default export of type Plugin!`);
+          return;
+        }
+
+        // Register plugin.
+        this.container.pluginService.register(plugin);
```
joey-colon, line 63: can we move this to be within the try? and then remove the `return` within the error block

src/app/bot.ts
```diff
@@ -36,18 +36,33 @@ export class Bot {
...
-        .forEach((plugin) => this.container.pluginService.register(plugin, this.container));
-    } catch (e) {
-      this.container.loggerService.error(e);
-    }
+    const pluginFolder = './src/app/plugins';
+    fs.readdir(pluginFolder, (_err, files) => {
+      files.forEach(async file => {
+
+        // Import the class from the plugin file.
+        const thing = await import(`./plugins/${file}`);
```
joey-colon, line 44: can we use a more descriptive name? maybe `pluginImport` or rename what `plugin` is known as now to `pluginInstance` and move `thing` to be = `plugin`

tanndlin (commented):
Is creating a display name relevant to this topic?

src/common/plugin.ts
```diff
@@ -6,6 +6,8 @@ export abstract class Plugin implements IPlugin {
 
   public abstract get name(): string;
```
joey-colon, line 7: i think this property should still be referring to the plugin name. to have it refer to the command name is kind of confusing when we consider the base class and interface we're abiding to is called `Plugin`.

can we maybe refer to `name` as the name of plugin and change `displayName` to be `command` or `commandName`?

src/app/bot.ts
suneettipirneni: 👀

suneettipirneni:
> Is creating a display name relevant to this topic?

The `displayName` property is just the new name for the older `name` property.

tanndlin:
> > Is creating a display name relevant to this topic?
> 
> The `displayName` property is just the new name for the older `name` property.

Oh I see what you did now

tanndlin (approved):

Armintron (approved):
Looks good to me

PR #511, opened 2021-06-27
Remove `environment.ts` in favor of declaration merging.

suneettipirneni:
It's more common to merge declarations of `process.env` rather than create a wrapper. It seemed that the main use of the `Environment` class was to enable autocompletions of known env vars used in our project. A declaration file achieves this without needing the developer to have any extra imports like `Environment.ts` needed.

### What about the default values in the `Environment.ts` file?
I believe that providing default values for env vars right out of the gate can be seen as confusing and misleading. For example, one might expect `MY_COOL_VAR` to be undefined, when in reality it has a default value so it's always truthy. Or a developer may think that the program is modifying their env vars on their machine, because a value is always filled. Because of this defaults are now decided outside of the environment within the call/usage site.

_Regardless only 3 env vars did this, and one of them was already being provided a default value in the call site._


Armintron (approved):
Looks good to me

tanndlin (approved):

PR #513, opened 2021-06-28
Removed ESLint Indent in favor of Prettier

tanndlin:
ESLint and Prettier should not be used for formatting at the same time, and they conflict

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -31,14 +31,6 @@
     "no-unused-vars": "off",
     "@typescript-eslint/no-unused-vars": ["error", { "caughtErrors": "none" }],
     "@typescript-eslint/no-explicit-any": "error",
-    "@typescript-eslint/indent": [
-      "error",
-      2,
-      {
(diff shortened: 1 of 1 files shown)


suneettipirneni:
I recently updated the deps with a newer `eslint-prettier-config`, can you run `npm i --only=dev` and reopen the project to see if the issue still persists?

tanndlin:
> I recently updated the deps with a newer `eslint-prettier-config`, can you run `npm i --only=dev` and reopen the project to see if the issue still persists?

Does not fix it. The amount of indentation is not the same when a declaration is on a different line, but this is like the 3rd time we've had this issue with Eslint and Prettier

suneettipirneni:
> > I recently updated the deps with a newer `eslint-prettier-config`, can you run `npm i --only=dev` and reopen the project to see if the issue still persists?
> 
> 
> 
> Does not fix it. The amount of indentation is not the same when a declaration is on a different line, but this is like the 3rd time we've had this issue with Eslint and Prettier

Ok fair enough

PR #516, opened 2021-06-29
Fix auto plugin loader

suneettipirneni:


tanndlin (commented):
Is this tester in a docker instance?

tanndlin (approved):

PR #519, opened 2021-07-01
Create Option for Plugin Category and Fix Help Bug

### LINKED ISSUE #508: Help Plugin not showing all plugins available

tanndlin:
also fixes #508


tanndlin:
This is in reference to #505. Makes it so that the Wolffram Plugin can be ran in a specific category

PR #524, opened 2021-07-03
Fix Leaderboard Build

tanndlin:


tanndlin:
Not necessary anymore

PR #528, opened 2021-07-06
Remove http and logger service.

suneettipirneni:
This is sliced from my larger branch, bascially instead of doing:

```ts
// HTTP
this.container.httpService.get
this.container.httpService.post

// Logging
this.container.loggerService.warn
this.container.loggerService.error
```

The properties are accessed via their modules:
```ts
// HTTP
axios.get
axios.post

// Logging
winston.warn
winston.error
```


joey-colon (changes requested):
lets talk on discord

c650:
The reason for these services is if we ever want to change the backend (axios, winston) to some other lib. 

is it over-engineering?

idk

suneettipirneni:
> The reason for these services is if we ever want to change the backend (axios, winston) to some other lib. 
> 
> 
> 
> is it over-engineering?
> 
> 
> 
> idk

I think premature abstraction is a more accurate description. This coupled with over abstraction. The issue with these abstractions is that they currently are more of obstacles to use than the solution in this PR. An abstraction should make things more accessible not less. The purpose has been defeated.

**What about swapping out services?**

There are still better solutions to this problem than using instances and containers. For example, axios and Winston don't need to be nested. Just make a static object/class that wraps them. It makes them easy to access and reduces useless indirection.

c650:
> > The reason for these services is if we ever want to change the backend (axios, winston) to some other lib.
> > is it over-engineering?
> > idk
> 
> I think premature abstraction is a more accurate description. This coupled with over abstraction. The issue with these abstractions is that they currently are more of obstacles to use than the solution in this PR. An abstraction should make things more accessible not less. The purpose has been defeated.
> 
> **What about swapping out services?**
> 
> There are still better solutions to this problem than using instances and containers. For example, axios and Winston don't need to be nested. Just make a static object/class that wraps them. It makes them easy to access and reduces useless indirection.

is there any middle-ground way we can migrate away without a 35 file diff?

suneettipirneni:
> > > The reason for these services is if we ever want to change the backend (axios, winston) to some other lib.
> 
> > > is it over-engineering?
> 
> > > idk
> 
> > 
> 
> > I think premature abstraction is a more accurate description. This coupled with over abstraction. The issue with these abstractions is that they currently are more of obstacles to use than the solution in this PR. An abstraction should make things more accessible not less. The purpose has been defeated.
> 
> > 
> 
> > **What about swapping out services?**
> 
> > 
> 
> > There are still better solutions to this problem than using instances and containers. For example, axios and Winston don't need to be nested. Just make a static object/class that wraps them. It makes them easy to access and reduces useless indirection.
> 
> 
> 
> is there any middle-ground way we can migrate away without a 35 file

The only way I see to do would be to keep the existing services and just wrap them to use the new services but if I'm being honest I don't see much issue in the diff, especially if it's mostly red.

suneettipirneni:
> > > > The reason for these services is if we ever want to change the backend (axios, winston) to some other lib.
> > 
> > > > is it over-engineering?
> > 
> > > > idk
> > 
> > > 
> > 
> > > I think premature abstraction is a more accurate description. This coupled with over abstraction. The issue with these abstractions is that they currently are more of obstacles to use than the solution in this PR. An abstraction should make things more accessible not less. The purpose has been defeated.
> > 
> > > 
> > 
> > > **What about swapping out services?**
> > 
> > > 
> > 
> > > There are still better solutions to this problem than using instances and containers. For example, axios and Winston don't need to be nested. Just make a static object/class that wraps them. It makes them easy to access and reduces useless indirection.
> > 
> > 
> > 
> > is there any middle-ground way we can migrate away without a 35 file
> 
> The only way I see to do would be to keep the existing services and just wrap them to use the new services but if I'm being honest I don't see much issue in the diff, especially if it's mostly red.

I take this back any solution would create a diff of the same size.

tanndlin:
As stated above and in the Discord, this is an unnecessary change and if anything a step backwards. This was created incase we want to change services. If merged this, then wanted to change services we'd have another pain in the ass 35 file diff

PR #531, opened 2021-07-09
Fizz Buzz in counting

### LINKED ISSUE #529: Do fizzBuzz in #counting
If one of yall codes this wrong lol rip

tanndlin:
fixes #529

### DIFF
--- a/src/app/handlers/counting.handler.ts
+++ b/src/app/handlers/counting.handler.ts
@@ -1,13 +1,19 @@
-import { TextChannel } from 'discord.js';
+import { EmojiIdentifierResolvable, TextChannel } from 'discord.js';
 import Constants from '../../common/constants';
-import { IMessage, IContainer, IHandler } from '../../common/types';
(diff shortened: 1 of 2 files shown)


src/app/handlers/counting.handler.ts
```diff
@@ -41,4 +55,19 @@ export class CountingHandler implements IHandler {
 
     return isNextNumber && isOnlyNumber;
   }
+
+  private _getEmojis() {
+    // The weird chars 🇫 && 🇧 are converted to the letter emojis in discord
+    this._fizzEmoji =
```
c650, line 61: You could make a getEmojiOrDefault and call it twice instead of the copypasta

c650 (approved):

src/app/handlers/counting.handler.ts
tanndlin: I feel like this pattern is in the code a lot. I going to make a getEmoji under guildService

PR #536, opened 2021-07-10
Animals Plugin

CallumCoy:
Simple plugin similar to dog and cat, but it can look up different animals.

I decided to keep cat and dog separate as they both have a massive API behind them already (so the code would have never ended).  Also, who wants to type `!animal dog` when you can already do `!dog`

I've tested all the arguments types it takes; however, not in one run as I am still trying to figure out which channel is causing issues with my testing bot.

And as always go easy on me, I'm but an old man now :older_man:.

### DIFF
--- a/src/app/plugins/animal.plugin.ts
+++ b/src/app/plugins/animal.plugin.ts
@@ -2,9 +2,10 @@ import Constants from '../../common/constants';
 import { Plugin } from '../../common/plugin';
 import { ChannelType, IContainer, IHttpResponse, IMessage, Maybe } from '../../common/types';
 import { MessageEmbed } from 'discord.js';
-import moment from 'moment';
+import ms from 'ms';
 
-export class AnimalPlugin extends Plugin {
+export default class AnimalPlugin extends Plugin {
+  public commandName: string = 'animal';
   public name: string = 'Animal Plugin';
   public description: string = 'Get wild with a fraction of gods creatures';
   public usage: string =
@@ -22,7 +23,7 @@ export class AnimalPlugin extends Plugin {
   private _speciesEmbed: Maybe<MessageEmbed>;
   private _subspeciesEmbed: Maybe<MessageEmbed>;
 
-  private _ANIMAL_UPD_THRESH: number = moment.duration(1, 'days').asMilliseconds();
+  private _ANIMAL_UPD_THRESH: number = ms('1d');
   private _LAST_UPD_TIME: number = 0;
 
   constructor(public container: IContainer) {
--- a/src/app/plugins/animal.plugin.ts
+++ b/src/app/plugins/animal.plugin.ts
@@ -2,6 +2,7 @@ import Constants from '../../common/constants';
 import { Plugin } from '../../common/plugin';
 import { ChannelType, IContainer, IHttpResponse, IMessage, Maybe } from '../../common/types';
 import { MessageEmbed } from 'discord.js';
+import moment from 'moment';
 
 export class AnimalPlugin extends Plugin {
   public name: string = 'Animal Plugin';
@@ -14,12 +15,15 @@ export class AnimalPlugin extends Plugin {
 
   private _API_URL: string = 'https://api.fetchit.dev/';
 
-  private _validAnimals: Set<string> = new Set([]);
-  private _species: string[] = [];
-  private _subspecies: ISubSpecies[] = [];
+  private _validAnimals: Set<string> = new Set([]); // All animals and species
+  private _species: string[] = []; // All species.
+  private _subspecies: ISubSpecies[] = []; // All subspecies housed by their species (not all species show here).
 
   private _speciesEmbed: Maybe<MessageEmbed>;
   private _subspeciesEmbed: Maybe<MessageEmbed>;
+
+  private _ANIMAL_UPD_THRESH: number = moment.duration(1, 'days').asMilliseconds();
+  private _LAST_UPD_TIME: number = 0;
 
   constructor(public container: IContainer) {
     super();
@@ -27,9 +31,11 @@ export class AnimalPlugin extends Plugin {
   }
 
   public async execute(message: IMessage, args?: string[]) {
+    this._updateAnimalAPIData();
     const input = args?.length ? args : [this._getRandomAnimal()];
 
-    if (input[0].includes('list')) {
+    // Checks if they want to know what species/ subspecies are available.
+    if (input[0].startsWith('list')) {
       if (!input[1] || input[1].startsWith('species')) {
         message.reply(this._makeSpeciesEmbed());
         return;
@@ -44,6 +50,7 @@ export class AnimalPlugin extends Plugin {
       return;
     }
 
+    // Preps the lookup data, and checks if it's valid.
     const animalLookUp: string = input.join('_');
     if (!this._validAnimals.has(animalLookUp)) {
       message.reply(`\`${input}\`is a invalid species/ sub-species.`);
@@ -51,34 +58,25 @@ export class AnimalPlugin extends Plugin {
     }
 
     let searchTerms: string = `name=${animalLookUp}`;
-
     const speciesIndex = this._subspecies.findIndex((species) => species.species === animalLookUp);
 
+    // Checks if it's a sub-species, it retrieves its species.
     if (!this._species.includes(animalLookUp)) {
-      const species = this._subspecies.reduce((targetSpecies, curSpecies) => {
-        if (targetSpecies) {
-          // we already found the species
-          return targetSpecies;
-        }
+      const species = this._subspecies.find((species) => species.subSpecies.includes(animalLookUp))
+        ?.species;
 
-        if (curSpecies.subSpecies.includes(animalLookUp)) {
-          return curSpecies;
-        }
-
-        return curSpecies;
-      });
-
-      searchTerms = `name=${species.species}&sub-species=${animalLookUp}`;
+      searchTerms = `name=${species}&sub-species=${animalLookUp}`;
     } else if (speciesIndex > -1) {
+      // Detects if the species has subspecies.  Then applies a random sub-species accordingly.
       const subSpeciesList: string[] = this._subspecies[speciesIndex].subSpecies;
       const randomSubspecies = subSpeciesList[Math.floor(Math.random() * subSpeciesList.length)];
 
       searchTerms = `name=${animalLookUp}&sub-species=${randomSubspecies}`;
     }
 
-    this.container.httpService
+    await this.container.httpService
       .get(`${this._API_URL}species/?${searchTerms}`)
-      .then((response: IHttpResponse) => {
+      .then(async (response: IHttpResponse) => {
         // Notifies the user if there was a problem contacting the server
         if (Math.floor(response.status / 100) !== 2) {
           message.reply(
@@ -88,10 +86,11 @@ export class AnimalPlugin extends Plugin {
         }
 
         const [animalData] = response.data.animal;
-
-        message.reply('', { files: [animalData.Image], name: 'image.jpg' });
+        await message.reply('', { files: [animalData.Image], name: 'image.jpg' });
       })
-      .catch();
+      .catch((err) => {
+        this.container.loggerService.warn(err);
+      });
   }
 
   private _getRandomAnimal(): string {
@@ -140,6 +139,15 @@ export class AnimalPlugin extends Plugin {
   }
 
   private _updateAnimalAPIData() {
+    // Checks if the API has passed it refresh time
+    const lastAnimalUpdate: number = Date.now() - this._LAST_UPD_TIME;
+
+    if (lastAnimalUpdate < this._ANIMAL_UPD_THRESH) {
+      return;
+    }
+
+    this._LAST_UPD_TIME = Date.now();
+
     this.container.httpService
       .get(`${this._API_URL}species/allspecies/`)
       .then((response: IHttpResponse) => {
@@ -168,6 +176,7 @@ export class AnimalPlugin extends Plugin {
       })
       .catch((err) => this.container.loggerService.warn(err));
 
+    // Clears out the saved embeds so that they can be reset upon the next call.
     this._speciesEmbed = undefined;
     this._subspeciesEmbed = undefined;
   }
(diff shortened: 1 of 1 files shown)


src/app/plugins/animal.plugin.ts
```diff
@@ -0,0 +1,189 @@
...
+      searchTerms = `name=${species}&sub-species=${animalLookUp}`;
+    } else if (speciesIndex > -1) {
+      // Detects if the species has subspecies.  Then applies a random sub-species accordingly.
+      const subSpeciesList: string[] = this._subspecies[speciesIndex].subSpecies;
+      const randomSubspecies = subSpeciesList[Math.floor(Math.random() * subSpeciesList.length)];
+
+      searchTerms = `name=${animalLookUp}&sub-species=${randomSubspecies}`;
+    }
+
+    await this.container.httpService
```
Armintron, line 78: Why use `await` and `.then`?

src/app/plugins/animal.plugin.ts
```diff
@@ -0,0 +1,189 @@
...
+  public pluginChannelName: string = Constants.Channels.Public.Pets;
+
+  private _API_URL: string = 'https://api.fetchit.dev/';
+
+  private _validAnimals: Set<string> = new Set([]); // All animals and species
+  private _species: string[] = []; // All species.
+  private _subspecies: ISubSpecies[] = []; // All subspecies housed by their species (not all species show here).
+
+  private _speciesEmbed: Maybe<MessageEmbed>;
+  private _subspeciesEmbed: Maybe<MessageEmbed>;
```
Armintron, line 24: Why use class members for the embeds? They are always written over anyways

src/app/plugins/animal.plugin.ts
```diff
@@ -0,0 +1,189 @@
...
+    // Checks if the API has passed it refresh time
+    const lastAnimalUpdate: number = Date.now() - this._LAST_UPD_TIME;
+
+    if (lastAnimalUpdate < this._ANIMAL_UPD_THRESH) {
+      return;
+    }
+
+    this._LAST_UPD_TIME = Date.now();
+
+    this.container.httpService
```
Armintron, line 152: use `await`

Armintron (changes requested):
On top of the requested changes, the bot seems to be spotty with throwing errors and responding correctly when using the plugin. This is regardless of arguments. See below:
image
Also, when an error occurs, Lion doesnt respond with anything. There should be a catch-all error message.

CallumCoy:
> On top of the requested changes, the bot seems to be spotty with throwing errors and responding correctly when using the plugin. This is regardless of arguments. See below:
> image
> Also, when an error occurs, Lion doesnt respond with anything. There should be a catch-all error message.

Yeah not too sure why, but this issue is for the first 20 seconds the list hasn't been generated.  In production this shouldn't be a problem as it unlikely someone will call it within that time frame.  Although if you want I can make a special case message for when the data hasn't been loaded.

src/app/plugins/animal.plugin.ts
CallumCoy: But it isn't, It only rewrites them if they are more than a day old, as the API is currently been worked on.

src/app/plugins/animal.plugin.ts
CallumCoy: ^ looking at this, this is probably why that first bug is a thing thx.

tanndlin (changes requested):

src/app/plugins/animal.plugin.ts
```diff
@@ -0,0 +1,189 @@
...
+    // Checks if it's a sub-species, it retrieves its species.
+    if (!this._species.includes(animalLookUp)) {
+      const species = this._subspecies.find((species) => species.subSpecies.includes(animalLookUp))
+        ?.species;
+
+      searchTerms = `name=${species}&sub-species=${animalLookUp}`;
+    } else if (speciesIndex > -1) {
+      // Detects if the species has subspecies.  Then applies a random sub-species accordingly.
+      const subSpeciesList: string[] = this._subspecies[speciesIndex].subSpecies;
+      const randomSubspecies = subSpeciesList[Math.floor(Math.random() * subSpeciesList.length)];
```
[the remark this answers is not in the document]
CallumCoy: Yeah but one is pulling from a set and the other is pulling from an array, so it would just make function longer than it needs.

and the reason I have one for animal and not sub-species, is the line gets quite long for the random animal when it is inline, but sub-species is an okay length.

Open to changing it though, as right now it is not consistently handled.

src/app/plugins/animal.plugin.ts
```diff
@@ -0,0 +1,189 @@
...
+    'animal <Species (optional)> <subSpecies(Optional)> | animal list (species or <species> or subSpecies)';
+  public pluginAlias = ['wild', 'species'];
+  public permission: ChannelType = ChannelType.Public;
+  public pluginChannelName: string = Constants.Channels.Public.Pets;
+
+  private _API_URL: string = 'https://api.fetchit.dev/';
+
+  private _validAnimals: Set<string> = new Set([]); // All animals and species
+  private _species: string[] = []; // All species.
+  private _subspecies: ISubSpecies[] = []; // All subspecies housed by their species (not all species show here).
```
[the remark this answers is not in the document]
CallumCoy: Not sure what you mean?  there is already a map for subspecies

`
this._subspecies = this._species
          .filter((species) => speciesData[species].length > 0 && species != '')
          .map((species: string) => {
            return {
              species: species,
              subSpecies: speciesData[species],
            };
          });
`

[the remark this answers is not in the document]
tanndlin: I meant like a Hashmap or a record
[SO link](https://stackoverflow.com/questions/42211175/typescript-hashmap-dictionary-interface)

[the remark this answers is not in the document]
tanndlin: So one species will map to an array of subspecies

src/app/plugins/animal.plugin.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  private _speciesEmbed: Maybe<MessageEmbed>;
+  private _subspeciesEmbed: Maybe<MessageEmbed>;
+
+  private _ANIMAL_UPD_THRESH: number = ms('1d');
+  private _LAST_UPD_TIME: number = 0;
+
+  constructor(public container: IContainer) {
+    super();
+    this._updateAnimalAPIData();
```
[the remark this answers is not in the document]
CallumCoy: Since the API takes so long to load I'm going to leave it in, so you don't have to wait 30-40 seconds for it to load after you call it for the first time.

src/app/plugins/animal.plugin.ts
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  public execute(message: IMessage, args?: string[]) {
+    this._updateAnimalAPIData();
+    const input = args?.length ? args : [this._getRandomAnimal()];
+
+    // Checks if they want to know what species/ subspecies are available.
+    if (input[0].startsWith('list')) {
```
tanndlin, line 38: array destructuring

src/app/plugins/animal.plugin.ts
```diff
@@ -94,6 +84,21 @@ export default class AnimalPlugin extends Plugin {
       });
   }
 
+  private async _pickListType(message: IMessage, input?: string) {
```
tanndlin, line 84: use a better name than input

src/app/plugins/animal.plugin.ts
```diff
@@ -94,6 +84,21 @@ export default class AnimalPlugin extends Plugin {
...
+    if (input === undefined || input.startsWith('species')) {
+      message.reply(this._makeSpeciesEmbed());
+      return;
+    }
+
+    if (input && input.startsWith('subspecies')) {
```
tanndlin, line 90: if you use 
```ts
if(!input || ...)
```
at the top, then you wont have to check if its true down here

src/app/plugins/animal.plugin.ts
```diff
@@ -0,0 +1,195 @@
...
+    const embed = new MessageEmbed();
+    embed.setColor('#0099ff').setTitle('Subspecies');
+
+    this._animals
+      .filter((animal) => !!animal.subSpecies)
+      .forEach((animal) => {
+        embed.addField(animal.species, animal.subSpecies.join('\n'), true);
+      });
+
+    return (this._subspeciesEmbed = embed);
```
tanndlin, line 130: its not expensive computationally to create this embed, it would simplify it a lot if you did not cache it

tanndlin: and same for the rest of the embeds

tanndlin (changes requested):
Sorry I forgot about this

src/app/plugins/animal.plugin.ts
CallumCoy: Not sure what you getting at.  did you want the destructuring removing?

tanndlin: use array destructuring for `input`

src/app/plugins/animal.plugin.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    // Preps the lookup data, and checks if it's valid.
+    const animalLookUp: string = input.join('_');
+    if (!this._validAnimals.has(animalLookUp)) {
+      message.reply(`\`${input}\`is a invalid species/ sub-species.`);
+      return;
+    }
+
+    let searchTerms: string = `name=${animalLookUp}`;
+    const animalIndex = this._animals.findIndex((animal) => animal.species === animalLookUp);
```
tanndlin, line 48: why not use this._animals.find(...)
and check if thats undefined

src/app/plugins/animal.plugin.ts
```diff
@@ -0,0 +1,181 @@
...
+    const species = this._findAnimalsFromSubspecies(animalLookUp)?.species;
+
+    // Checks if it's a sub-species, if so it retrieves its species.
+    if (species !== undefined) {
```
tanndlin, line 52:
```suggestion
    if (species) {
```

src/app/plugins/animal.plugin.ts
```diff
@@ -0,0 +1,181 @@
...
+      searchTerms = `name=${species}&sub-species=${animalLookUp}`;
+    } else if (this._animals[animalIndex].subSpecies.length > 0) {
+      // Detects if the species has subspecies.  Then applies a random sub-species accordingly.
+      const subSpeciesList: string[] = this._animals[animalIndex].subSpecies;
+      const randomSubspecies = subSpeciesList[Math.floor(Math.random() * subSpeciesList.length)];
+
+      searchTerms = `name=${animalLookUp}&sub-species=${randomSubspecies}`;
```
tanndlin, line 59: instead of having this pattern in multiple places, it could be cool to make a function that builds that url for us
```ts
const createSearchTerm = (species: string, sub?: string) => {
    let searchTerm = `name=${species}`;
    if(sub){
        searchTerm += `&sub-species=${sub}`
    }

    return searchTerm;
}
```

tanndlin, line 59: all of this 'getting the searchTerm' should be its own function. having all these if's makes it really complex and makes the main function too complicated

tanndlin (changes requested):

src/app/plugins/animal.plugin.ts
```diff
@@ -78,6 +65,27 @@ export default class AnimalPlugin extends Plugin {
       });
   }
 
+  private createSearchTerm(data: string, sub: boolean){
```
tanndlin: use a more useful name than `sub`. Maybe `isSubSpecies`

src/app/plugins/animal.plugin.ts
```diff
@@ -0,0 +1,189 @@
...
+        return data;
+      } 
+      if (species.subSpecies.length > 0) {
+        return this._getRandomSubSpecies(species);
+      } 
+    };
+
+    return `name=${species.species}&subspecies=${subSpecies}`;
+
+  }
```
tanndlin, line 82: I dont understand this. This could use either comments or a good refactor
or both

src/app/plugins/animal.plugin.ts
```diff
@@ -0,0 +1,189 @@
...
+      return;
+    }
+
+    if (listType.startsWith('subspecies')) {
+      message.reply(this._makeSubspeciesEmbed());
+      return;
+    }
+
+    await message.reply(this._makeSingleSubSpeciesEmbed(listType));
+    return;
```
tanndlin, line 96: useless return

src/app/plugins/animal.plugin.ts
```diff
@@ -0,0 +1,189 @@
...
+    // Preps the lookup data, and checks if it's valid.
+    const animalLookUp: string = input.join('_');
+    if (!this._validAnimals.has(animalLookUp)) {
+      message.reply(`\`${input}\`is a invalid species/ sub-species.`);
+      return;
+    }
+
+    const searchTerms: string = this._createSearchTerm(animalLookUp, !!this._findAnimalsFromSubspecies(animalLookUp)?.species);
+
+    this.container.httpService
```
tanndlin, line 49: await

src/app/plugins/animal.plugin.ts
tanndlin: bump

tanndlin (changes requested):

CallumCoy:
https://github.com/cs-discord-at-ucf/lion/pull/536#discussion_r694424994
I don't speak git at all.  If you mean also add an await here just did that.  Or if you mean rebase, then mb I will make sure to do that this time before I push.

But if you mean update version which is was google says it means, then you're gonna have to elaborate since I am not aware of us using versions at all.

Also, no I didn't die, just had to leave this to age like fine cheese.

src/app/plugins/animal.plugin.ts
CallumCoy: Had a go at commenting and changing the name of the function. You'll have to let me know if that is good enough, cause I'm too lazy to refactor the first time around.

tanndlin:
> [#536 (comment)](https://github.com/cs-discord-at-ucf/lion/pull/536#discussion_r694424994)
> I don't speak git at all. If you mean also add an await here just did that. Or if you mean rebase, then mb I will make sure to do that this time before I push.
> 
> But if you mean update version which is was google says it means, then you're gonna have to elaborate since I am not aware of us using versions at all.
> 
> Also, no I didn't die, just had to leave this to age like fine cheese.

bump means like +1
and `await` means to await that promise yeah

PR #538, opened 2021-07-19
Don't make a mongodb connection required

suneettipirneni:


Armintron (commented):
Shouldn't we instead not instantiate the storage service in general and just check if it's null? This checks if you specifically dont have env vars, but doesnt check if you have a valid mongo connection

suneettipirneni:
> Shouldn't we instead not instantiate the storage service in general if you dont have the vars and just check if it's null? This checks if you specifically dont have env vars, but doesnt check if you have a valid mongo connection

If you don't have the env vars set you can't have a valid connection. MongoDB was never required until my changes to plugin state.

Armintron (commented):
Isn't this logic specific to the function? Why not put it in the function?

Armintron (approved):

src/services/plugin.service.ts
```diff
@@ -21,6 +21,11 @@ export class PluginService {
 
   public async initPluginState(container: IContainer): Promise<void> {
 
+    if (!process.env.MONGO_DB_NAME || !process.env.MONGO_URL ||
+      !process.env.MONGO_USER_NAME || !process.env.MONGO_USER_PASS) {
+        return;
+    }
```
tanndlin, line 27: Does this default to on?

src/services/plugin.service.ts
```diff
@@ -21,6 +21,11 @@ export class PluginService {
 
   public async initPluginState(container: IContainer): Promise<void> {
 
+    if (!process.env.MONGO_DB_NAME || !process.env.MONGO_URL ||
```
c650, line 24: prefer using `Environment`, not `process.env`

src/services/plugin.service.ts
```diff
@@ -21,6 +21,11 @@ export class PluginService {
...
+      !process.env.MONGO_USER_NAME || !process.env.MONGO_USER_PASS) {
+        return;
+    }
+
     if (!mongoose.connection.readyState) {
       await container.storageService.connectToDB();
     }
```
c650, line 31: cant we just try/catch instead

src/services/plugin.service.ts
suneettipirneni: `Environment` was removed

c650: seethe

src/services/plugin.service.ts
suneettipirneni: This is for connecting to the database if not already connected. If we get to this point, we should be able to connect, and no errors should be thrown.

c650 (approved):

CallumCoy (approved):
lgtm, and Thank you.

Armintron (approved):

PR #551, opened 2021-08-15
V13

tanndlin:
https://discordjs.guide/additional-info/changes-in-v13.html#guildmember

I dont know what to do about [this last error](https://github.com/cs-discord-at-ucf/lion/blob/15b28da24face937f2553fe2a665fb33ecddcf11/src/__tests__/animal.test.ts#L13) bc I dont know jest
image
```
src/__tests__/animal.test.ts:13:35 - error TS2493: Tuple type '[options: string | MessagePayload | ReplyMessageOptions]' of length '1' has no element at index '1'.

const arg = spy.mock.calls[0][1];
```


tanndlin:
do i yolo

PR #554, opened 2021-08-22
moma

VijayStroup:


tanndlin:
why

src/app/handlers/moma.handler.ts
```diff
@@ -0,0 +1,11 @@
+import { IContainer, IHandler, IMessage } from '../../common/types';
+
+export class MomaHandler implements IHandler {
+  constructor(public container: IContainer) {}
+
+  public async execute(message: IMessage): Promise<void> {
```
c650, line 6: Do the If-not-return pattern at least

c650 (changes requested):

VijayStroup:
> why

image

tanndlin:
What branch is this based off of?

tanndlin (changes requested):
Maybe atleast generate a random number so 1/100 times or so Lion will do it

VijayStroup:
> What branch is this based off of?
master like 10min ago.

> Maybe atleast generate a random number so 1/100 times or so Lion will do it
I checked to see what would constitute a good message to reply to it with, and from my messages, it seems I normally reply with that gif if someone begins their message with "who".  This is also only done roughly 2-3 times per day with some days having none.  Would you still like me to implement a random chance, or is this quantity low enough?

c650:
Limit this to certain channels.

tanndlin:
> > What branch is this based off of?
> 
> master like 10min ago.
> 
> > Maybe atleast generate a random number so 1/100 times or so Lion will do it
> 
> I checked to see what would constitute a good message to reply to it with, and from my messages, it seems I normally reply with that gif if someone begins their message with "who". This is also only done roughly 2-3 times per day with some days having none. Would you still like me to implement a random chance, or is this quantity low enough?

how about 1/4 - 1/10 so its not predictable and annoying

VijayStroup:
> how about 1/4 - 1/10 so its not predictable and annoying

50/50 compromise?

tanndlin:
> > how about 1/4 - 1/10 so its not predictable and annoying
> 
> 50/50 compromise?

fine
i may end up reducing it if it gets annoying

src/app/handlers/moma.handler.ts
```diff
@@ -0,0 +1,33 @@
...
+    Constants.Channels.Public.Vehicles,
+    Constants.Channels.Public.Fitness,
+    Constants.Channels.Public.Food,
+    Constants.Channels.Public.VideoGames
+  ];
+
+  constructor(public container: IContainer) {}
+
+  public async execute(message: IMessage): Promise<void> {
+    if (!message.content.toLowerCase().startsWith('who ') || !message.content.toLowerCase().startsWith('who')) {
```
tanndlin, line 18: `!message.content.toLowerCase().startsWith('who')` is redudant

src/app/handlers/moma.handler.ts
```diff
@@ -0,0 +1,33 @@
...
+      return;
+    }
+
+    const chan = message.channel as TextChannel;
+    if (!this._whitelistedChannels.includes(chan.name.toLowerCase())) {
+      return;
+    }
+
+    const coinFlip: number = Math.floor(Math.random() * 2);
```
tanndlin, line 27: 50% of the time is pretty obnoxious

tanndlin (changes requested):

PR #579, opened 2021-09-05
Prevent nick names from including pings

### LINKED ISSUE #573: Pings in nick name changes arent prevented
#  The problem
Quite simply, a user can ping moderators/admins (or really anyone with access to the channel nickname changes are logged).

This is done by changing your nickname to something that has a valid ping in between two '\`' characters.

## An example: 
```
hello `<@userid/roleid>`
``` 
The above will ping the user or role specified by the id in the log change channel.

## Image Examples:
image
The above nick name change results in the following being logged:
image

TheIceCreamBear:
Removes the  ` character from showing up in nicknames logs, and replaces it with ' 

This prevents the creation of links to channels and pings of roles/users with the id of that channel/role/user.

Example of pings:
```
<#channel_id> - this links to the channel specified by channel_id
<@user_id>    - this pings the user specified by user_id
<@&role_id>   - this pings the role specified by role_id
```

This fixes #573

### DIFF
--- a/src/app/handlers/user_update.handler.ts
+++ b/src/app/handlers/user_update.handler.ts
@@ -15,8 +15,8 @@ export class UserUpdateHandler implements IHandler {
 
     this.container.messageService.sendBotReport(
       `User ${shouldPing ? newUser.user : newUser.user.tag} changed their name from \`${
-        oldUser.displayName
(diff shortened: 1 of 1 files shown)


tanndlin (approved):
lit, thanks bruh

PR #581, opened 2021-09-15
added bot log for messages longer than 199 Characters

Zachmayers:
Only two lines, just added a quick check for messages that are 200 characters or longer at the request of one of the Moderators.


src/app/handlers/blacklist.handler.ts
```diff
@@ -58,8 +58,9 @@ export class BlacklistHandler implements IHandler {
     const isClassChannel = this.container.classService.getClasses(ClassType.ALL).has(channel.name);
     const hasBackticks = message.content.toLowerCase().match(/```/);
     const hasAttachment = message.attachments.size;
+    const messageIsLong = message.content.length >= 200 ? true : false;
```
tanndlin, line 61: ternary is redundant `message.content.length >= 200` is already a boolean

tanndlin (changes requested):

Issue #588, opened 2021-09-28
V13 Update issues


tanndlin:
Keeping track of all v13 upgrading related issues here

- [ ] No image in bot logs from class channels

tanndlin:
The only job that runs every hour is weatherevents

PR #594, opened 2021-09-29
Fix market errors

tanndlin:
For some reason `message.deletable` and `message.deleted` no longer reliably work
also removed check for if a user is in the server because it was creating false negatives causing the market list to be empty
added a check for when the market list is empty too


tanndlin:
fixes #593 
refs #588

PR #604, opened 2021-10-14
add mod name to warnings

vijaystroup:
allow for more transparency for end users.


Kurtoid:
> moderating less fun
_should_ moderating be fun?

tanndlin:
It shouldn't be torture 😃

PR #633, opened 2022-01-06
Points

### LINKED ISSUE #630: Point System
- Users are awarded a point every time they do something
    - Sending messages
    - Getting reactions? Like Reddit karma 
- A Plugin that a user can play with the gamble their points
    - ie: a coinflip to double or lose betted amount
- Leaderboard to see who has the most points
    - Similar to the C4 and TTT leaderboards

tanndlin:
New Point System
fixes #630


tanndlin:
This was all reviewed in previous PRs with no changes 
so merging now

PR #640, opened 2022-01-16
Added catch to client.login

landoniwnl:
No comment

### DIFF
--- a/src/services/client.service.ts
+++ b/src/services/client.service.ts
@@ -14,7 +14,7 @@ export class ClientService extends Client {
         Intents.FLAGS.GUILD_VOICE_STATES,
       ],
     });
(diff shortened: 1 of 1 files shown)


src/services/client.service.ts
```diff
@@ -14,7 +14,7 @@ export class ClientService extends Client {
         Intents.FLAGS.GUILD_VOICE_STATES,
       ],
     });
-    this.login(process.env.DISCORD_TOKEN);
+    this.login(process.env.DISCORD_TOKEN).catch(e => console.log(e));
```
tanndlin, line 17: I was just using console.log to debug
But since this wont make it into production I dont think it matters? @c650

c650: Fine w me

tanndlin (approved):

PR #644, opened 2022-01-22
Fixed quote indentation issue on mobile

landoniwnl:
There is an indent before the author name on mobile.  This should fix it.

### DIFF
--- a/src/app/plugins/quote.plugin.ts
+++ b/src/app/plugins/quote.plugin.ts
@@ -56,7 +56,7 @@ export default class QuotePlugin extends Plugin {
       .setTitle(tag === 'Famous-quotes' ? 'Famous-Quotes' : tag)
       .setFooter('Pulled from api.quotable.io')
       .setTimestamp(new Date())
-      .setDescription(content + '\n\n- ' + author);
+      .setDescription(`${content}\n\n- ${author}`);
   }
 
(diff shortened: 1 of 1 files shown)


src/app/plugins/quote.plugin.ts
```diff
@@ -56,11 +56,7 @@ export default class QuotePlugin extends Plugin {
       .setTitle(tag === 'Famous-quotes' ? 'Famous-Quotes' : tag)
       .setFooter('Pulled from api.quotable.io')
       .setTimestamp(new Date())
-      .setDescription(
-        `${content}
-
-        - ${author}`
-      );
+      .setDescription(content + '\n\n- ' + author);
```
suneettipirneni, line 59: String interpolation should be used here.

```suggestion
      .setDescription(`${content}\n\n-${author}`);
```

suneettipirneni (changes requested):

src/app/plugins/quote.plugin.ts
landoniwnl: is there a specific reason to use it over a regular string?

suneettipirneni: It's more concise, and allows easier modification of the string, if needed, down the line.

PR #645, opened 2022-01-22
fix inconsistencies in quote plugin

suneettipirneni:
Makes the following changes:
- Marks non-mutated fields as `static` and `readonly`
- Caches strings that were the always the same value as static fields
- Uses `await` + assignment rather than `.then` to avoid nesting.
- Adds typing to the quote API responses

### DIFF
--- a/src/app/plugins/quote.plugin.ts
+++ b/src/app/plugins/quote.plugin.ts
@@ -62,7 +62,9 @@ export default class QuotePlugin extends Plugin {
     }
 
     const [tagInput] = args;
-    return tagInput === 'list' || QuotePlugin._VALID_TAGS.includes(tagInput.toLowerCase());
+    return (
+      tagInput.toLowerCase() === 'list' || QuotePlugin._VALID_TAGS.includes(tagInput.toLowerCase())
+    );
   }
 
   private _createEmbed(content: string, author: string, tag: string) {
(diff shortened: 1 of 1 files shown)


src/app/plugins/quote.plugin.ts
```diff
@@ -47,15 +61,16 @@ export default class QuotePlugin extends Plugin {
       return true;
     }
 
-    return args[0] === 'list' || this._VALID_TAGS.includes(args[0].toLowerCase());
+    const [tagInput] = args;
+    return tagInput === 'list' || QuotePlugin._VALID_TAGS.includes(tagInput.toLowerCase());
```
tanndlin, line 65: I should've caught this in the first PR for this plugin,
but this should be lowercased
```suggestion
    return tagInput.toLowerCase() === 'list' || QuotePlugin._VALID_TAGS.includes(tagInput.toLowerCase());

```

src/app/plugins/quote.plugin.ts
```diff
@@ -68,31 +83,30 @@ export default class QuotePlugin extends Plugin {
...
     this.validate(message, args);
+    const [inputTag] = args;
+    const url: string =
+      args.length === 1
+        ? `${QuotePlugin._DEFAULT_URL}?tags=${inputTag.toLowerCase()}`
+        : QuotePlugin._DEFAULT_URL;
 
-    if (args[0] === 'list') {
-      await message.reply('**Valid tags:** ' + this._VALID_TAGS.join(', '));
+    if (inputTag === 'list') {
```
tanndlin, line 93: saa
```suggestion
    if (inputTag.toLowerCase() === 'list') {

```

tanndlin (changes requested):

tanndlin:
Oh also should be on dev branch

tanndlin (approved):

PR #647, opened 2022-01-22
use more concise logic for ta plugin

suneettipirneni:


tanndlin:
Lol was this really that important

Either way, this doesnt work, the point of the commit was to allow this command in threads for `!ta ask` but not register/remove

PR #653, opened 2022-02-06
add external logger

joeydotdev:


tanndlin (approved):
I trust you lol

scripts/logger.js
```diff
@@ -0,0 +1,38 @@
...
+}
+
+function join() {
+  ensureLogger();
+  return through((data) => {
+    logger.info(data);
+  });
+}
+
+process.stdin.pipe(split()).pipe(join()).pipe(process.stdout);
```
c650, line 33: What is split

c650 (commented):
Please also capture stderr 

otherwise Makes sense to me.

id like a comment on what the split function is 

also we might as well strip down the logger service

Dockerfile
```diff
@@ -5,4 +5,4 @@ COPY package*.json ./
 RUN npm install
 COPY . .
 RUN npm run build
-CMD ["node", "./dist/index.js"]
+CMD "node ./dist/index.js | node ./scripts/logger.js"
```
c650, line 8: Put a 2 >& 1 before the pipe

c650 (approved):
LGTM as long as you ran it in docker and it works

tanndlin:
can u change the branch so its on develop

PR #661, opened 2022-03-02
Remove market sold feature

tanndlin:
Bug introduced in #312 allowed users to mark others' posts as sold.

I have decided to remove this feature, as when I tried to reintroduce it the bot takes a long time to await the reactions' users.
This feature does the exact same as removing the `!market add` from your message, since #312 made it so that sold listings are no longer visible.
The DM from the bot has been changed to reflect this.


tanndlin:
Ill test this in a bit

PR #670, opened 2022-03-15
Allow ID's for CheckClasses Plugin

tanndlin:
Does what the title says.

I made a new method in Moderation to convert a form of identification into a `GuildMember`.
Used that to parse the args for the CheckClasses Plugin

And added unit test for the regex to be safe


tanndlin:
This would probably be a good time to make a `is valid user identification string` method or something of the sort

PR #673, opened 2022-03-20
Exit early in blacklist handler

tanndlin:
When the message is deleted it **could??** cause a crash if we try to access its attributes later. So exit the loop early in that case.

I'm not sure of a better way to do this that is still obvious what's trying to be accomplished.
Technically using a `.every` would exit early when you `return false` but that just feels unintuitive for what I'm trying to do

### DIFF
--- a/src/app/handlers/blacklist.handler.ts
+++ b/src/app/handlers/blacklist.handler.ts
@@ -66,6 +66,10 @@ export class BlacklistHandler extends Handler {
       }
     });
 
+    if (deleted) {
+      return;
+    }
+
     const isClassChannel = this.container.classService.getClasses(ClassType.ALL).has(channel.name);
     const hasBackticks = message.content.toLowerCase().match(/```/);
     const hasAttachment = message.attachments.size;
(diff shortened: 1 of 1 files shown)


src/app/handlers/blacklist.handler.ts
```diff
@@ -44,8 +44,9 @@ export class BlacklistHandler extends Handler {
       return;
     }
 
+    let deleted = false;
     this._expressions.forEach(async ({ regex, label }) => {
```
c650, line 48: Let’s change this forEach. It’s making this code messy. Instead, 

let patternsFound = expressions.filter()

…

if patternsFound is not empty, file report, delete message, etc.

This way, finding the pattern doesn’t leave side effects.

c650 (changes requested):
See comment

src/app/handlers/blacklist.handler.ts
tanndlin: thx

tanndlin:
Upon making that change, I feel like the crash could have been an unawaited DM, which would crash if the user has DMs turned off. Either way, I do appreciate refactoring this handler

Issue #672, opened 2022-03-15
Convert !ta ask to an embed


tanndlin:
make the question appear in an embed or something because its super ugly right now.
The ping to the TA's cannot be in the embed or it wont *actually* ping

c650:
@bquach2

PR #681, opened 2022-04-17
Suggestion Feature for Issue#655

### LINKED ISSUE #655: Suggestions/feedback plugin
Create a plugin that will allow for a user to submit an anonymous suggestion for the server. The Moderation/Admin team should ideally review the submitted suggestions once a month or cycle and implement suggestions at their discretion. Suggestions can be displayed in a read-only channel.

Invocation could be something like:
`!suggest <suggestion>`

ex. !suggest Creating an NFT channel.

bquach2:
Implemented Suggestion plugin requested in issue #655 and created a test case for it. Fixes #655


src/app/plugins/suggestions.plugin.ts
```diff
@@ -0,0 +1,23 @@
+import { MessageEmbed} from 'discord.js';
+import { Plugin } from '../../common/plugin';
+import { IContainer, IMessage, ChannelType } from '../../common/types';
+
+export default class SuggestionPlugin extends Plugin {
+  public commandName: string = 'suggest';
+  public name: string = 'Suggestions Plugin';
+  public description: string = 'Create suggestions for lion to output';
+  public usage: string = 'suggest <string>';
```
c650, line 9: Only suggestions channel

src/app/plugins/suggestions.plugin.ts
```diff
@@ -0,0 +1,23 @@
...
+  public permission: ChannelType = ChannelType.All;
+  private _embed: MessageEmbed = new MessageEmbed();
+  constructor(public container: IContainer) {
+    super();
+  }
+
+  public async execute(message: IMessage, args: string[]) {
```
c650, line 16: Can we rate limit to prevent abuse

c650 (changes requested):

c650:
I think this is good for #672 too/instead

c650:
I think this is good for #672 too/instead

bquach2:
Updated to only work in Suggestions channel, thanks!

src/app/plugins/suggestions.plugin.ts
```diff
@@ -0,0 +1,25 @@
...
+import Constants from '../../common/constants';
+
+export default class SuggestionPlugin extends Plugin {
+  public commandName: string = 'suggest';
+  public name: string = 'Suggestions Plugin';
+  public description: string = 'Create suggestions for lion to output';
+  public usage: string = 'suggest <string>';
+  public override pluginChannelName: string = Constants.Channels.Public.ServerRequests;
+  public permission: ChannelType = ChannelType.All;
+  private _embed: MessageEmbed = new MessageEmbed();
```
tanndlin, line 13: I dont see a reason to store the embed as a property of the class when nothing is carried over from each use.

You could add a footer to the embed to justify this, but even then `new MessageEmbed` is not an expensive operation.
I would just make a new one each time

src/app/plugins/suggestions.plugin.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(public container: IContainer) {
+    super();
+  }
+
+  public async execute(message: IMessage, args: string[]) {
+    const suggestion = args.join(' ');
+    // Combines input
+    this._embed.setTitle('Suggestion:');
+    this._embed.setDescription(suggestion);
+    await message.reply({ embeds: [this._embed] });
```
tanndlin, line 23: Would be nice to have Lion react with a thumbs up then thumbs to down
to imply to the user that you should react to vote

tanndlin (changes requested):

bquach2:
I updated to make a new embed object every time, thanks!

PR #683, opened 2022-04-23
Force image extensions in pet plugins

### LINKED ISSUE #613: Cat (and probably dog) plugins giving a .false file
This was fixed at one point [here](https://github.com/cs-discord-at-ucf/lion/commit/e15c0f99cbc5a0bd55ef9e51ff234058aa27ba7d#diff-bd24a41506774a43f65a64f7bbf134a214cf460283b44236e1a0dd698238b950) but had to be removed with the djs13 update as it was not a valid option in `message.reply()` anymore

So we need a new way of forcing it to be an image again

LambdaPaul:
Fix #613.


tanndlin (approved):
Appreciate this!

This was fixed before, then after DiscordJS 13 it got put back in and I didn't remember the fix

c650 (approved):

Issue #668, opened 2022-03-11
Taco Tuesday


tanndlin:
Give 2x (or maybe a random multiplier per week) on Tuesdays to taco gains.
Could also do a happy hour or something like that.

Would just check the time when adding tacos
inject it [here](https://github.com/cs-discord-at-ucf/lion/blob/5ecae6c2c9aa6169a3e18203040489bc1004eb46/src/services/point.service.ts#L13-L20)

PR #684, opened 2022-04-30
Added taco multiplier on Tuesdays for issue #668

landoniwnl:
I want more tacos

### DIFF
--- a/src/services/point.service.ts
+++ b/src/services/point.service.ts
@@ -15,10 +15,6 @@ export class PointService {
       return;
     }
 
-    if (new Date().getDay() === 2) {
-      amount = amount * 2;
-    }
-
     const userDoc = await this.getUserPointDoc(id);
     await PointsModel.updateOne(userDoc, { $inc: { numPoints: amount } });
   }
--- a/src/services/point.service.ts
+++ b/src/services/point.service.ts
@@ -15,6 +15,10 @@ export class PointService {
       return;
     }
 
+    if (new Date().getDay() === 2) {
+      amount = amount * 2;
+    }
+
     const userDoc = await this.getUserPointDoc(id);
     await PointsModel.updateOne(userDoc, { $inc: { numPoints: amount } });
   }
--- a/src/app/handlers/point_message.handler.ts
+++ b/src/app/handlers/point_message.handler.ts
@@ -14,6 +14,9 @@ export class PointMessageHandler extends Handler {
       return;
     }
 
-    await this.container.pointService.awardPoints(message.author.id, 1);
+    // Sets amount to 2 on tuesdays
+    const amount = new Date().getDay() === 2 ? 2 : 1;
+
+    await this.container.pointService.awardPoints(message.author.id, amount);
   }
 }


src/services/point.service.ts
```diff
@@ -15,6 +15,10 @@ export class PointService {
       return;
     }
 
+    if (new Date().getDay() === 2) {
+      amount = amount * 2;
+    }
```
Armintron, line 20:
```suggestion
   // give bonus points on tuesdays
    if (new Date().getDay() === 2) {
      amount *= 2;
    }
```

Armintron (changes requested):
Add comments to your code and use a big boi operator for self multiplcation

src/services/point.service.ts
c650: Is day zero based or one based? Does it start with Sunday or Monday?

Armintron: > 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDay
It starts with Sunday = 0

tanndlin (changes requested):
This should be separate from how gambling awards points in some way.
As is, it would make gambling print money

tanndlin:
> This should be separate from how gambling awards points in some way. As is, it would make gambling print money

Like double the amount in the `point_message.handler.ts` and/or `point_react.handler.ts`

Armintron:
> This should be separate from how gambling awards points in some way. As is, it would make gambling print money

That is inherient to any sort of multiplier. Maybe 2x is too much? Just 1.5x? At the end of the day its just the fun of it - its alright if you can more easily make tacos.

tanndlin:
> > This should be separate from how gambling awards points in some way. As is, it would make gambling print money
> 
> That is inherient to any sort of multiplier. Maybe 2x is too much? Just 1.5x? At the end of the day its just the fun of it - its alright if you can more easily make tacos.

Its not really gambling if ur guaranteed to make money

Armintron:
> > > This should be separate from how gambling awards points in some way. As is, it would make gambling print money
> > 
> > 
> > That is inherient to any sort of multiplier. Maybe 2x is too much? Just 1.5x? At the end of the day its just the fun of it - its alright if you can more easily make tacos.
> 
> Its not really gambling if ur guaranteed to make money

In the long run and in the limit to infinity yes, but you can still lose a lot with a big bet. Yes, you could make that back if you sit and spam but at the end of the day, you are driving attraction towards to plugin - which is almost always the end goal.

tanndlin:
> In the long run and in the limit to infinity yes, but you can still lose a lot with a big bet. Yes, you could make that back if you sit and spam but at the end of the day, you are driving attraction towards to plugin - which is almost always the end goal.
 
 Being able to min bet and get back 4x the min bet just ruins the leaderboard, thats inflation man

Armintron:
> > In the long run and in the limit to infinity yes, but you can still lose a lot with a big bet. Yes, you could make that back if you sit and spam but at the end of the day, you are driving attraction towards to plugin - which is almost always the end goal.
> 
> Being able to min bet and get back 4x the min bet just ruins the leaderboard, thats inflation man

So then how does your proposed change fix this? I didnt understand that comment

tanndlin:
> So then how does your proposed change fix this? I didnt understand that comment

Just dont affect the gambling amount

Armintron:
> > So then how does your proposed change fix this? I didnt understand that comment
> 
> Just dont affect the gambling amount

But thats the whole point of this PR. Was your note just on the correct way to alter an award amount? Im not following

tanndlin:
Just change [this](https://github.com/cs-discord-at-ucf/lion/blob/c75a9a0239e7ace53ad2d6863db23ee911a6b9e6/src/app/handlers/point_message.handler.ts#L17) lol

The gamble uses the point service `awardPoints`, so just don't change that

Armintron:
> Just change [this](https://github.com/cs-discord-at-ucf/lion/blob/c75a9a0239e7ace53ad2d6863db23ee911a6b9e6/src/app/handlers/point_message.handler.ts#L17) lol
> 
> The gamble uses the point service `awardPoints`, so just don't change that

I had no clue you were trying to say "we should double the amount users get for sending messages, not for gambling". I get it now. Yes that is a good middle ground

tanndlin:
> Like double the amount in the point_message.handler.ts and/or point_react.handler.ts

Armintron:
> > Like double the amount in the point_message.handler.ts and/or point_react.handler.ts

I understand but as a non-contributer idk what those things do until i look at them

src/services/point.service.ts
Armintron: we not doubling the amount for gambling

landoniwnl:
I didn't realize that what I did affected the gambling amount too. It should only give bonus points from messages now.

tanndlin (approved):
lgtm
we can change this number later if we feel it doesn't have enough impact

PR #691, opened 2022-05-02
Move report input system to IDs

tanndlin:
Makes the default type of identification for the report system a snowflake

This standardization, will make the Alt Tracker a more seamless implementatio

### DIFF
--- a/src/app/plugins/modreport.plugin.ts
+++ b/src/app/plugins/modreport.plugin.ts
@@ -105,9 +105,15 @@ export default class ModReportPlugin extends Plugin {
   }
 
   private async _handleFullList(message: IMessage, id: string) {
+    const member = await Moderation.Helpers.resolveUser(this.container.guildService.get(), id);
+    if (!member) {
+      await message.reply('Could not get member');
+      return;
+    }
+
     try {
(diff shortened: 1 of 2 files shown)


src/app/plugins/modreport.plugin.ts
```diff
@@ -70,69 +79,55 @@ export default class ModReportPlugin extends Plugin {
...
+      await this.container.modService.getModerationSummary(this.container.guildService.get(), id)
     );
   }
 
-  private async _handleFullList(message: IMessage, user_handle: string) {
+  private async _handleFullList(message: IMessage, id: string) {
     try {
       await message.reply({
-        content: `Full Report for ${user_handle}`,
+        content: `Full Report for ${id}`,
```
c650, line 110: can we pass both `user_handle` and `id` just for this ??

c650 (changes requested):

src/app/plugins/modreport.plugin.ts
tanndlin: Can fetch the user handle from the ID, I just want all these methods to be standardized

PR #703, opened 2022-05-08
db and log-related bug fixes

c650:


src/services/controller.service.ts
```diff
@@ -16,6 +17,12 @@ export class ControllerService {
       return;
     }
 
+    try {
+      await this._storageService.connectToDB();
```
tanndlin, line 21: Do we need to call this again?

It's already called in bot.ts, which precedes this

c650: i changed `connectToDB` to return a cached promise. that way we only do the work once but we can also enforce that it's been done

c650:
in addition to this PR i updated the mongo conn string in Secrets (in Settings on the repo) to make sure it didnt have quotes. we were getting conn issues (only in docker,,, so clearly some version differences) 

i also updated the Discord conn string because we asked joey to refresh it.

PR #716, opened 2022-08-01
Auto lose c4 after not making move in x time

### LINKED ISSUE #598: Auto lose c4 after x time
After y amount of moves have been made, if player z does not move within x time, make them automatically lose to prevent score abuse.

msabur:
Fixes #598 
The time limit is hard-coded to 2 minutes.

### DIFF
--- a/src/app/plugins/connectfour.plugin.ts
+++ b/src/app/plugins/connectfour.plugin.ts
@@ -145,7 +145,7 @@ class ConnectFourGame {
   private _aiDepth = 4;
 
   // messy initialization since typescript won't let me use -1 or null
-  private _timeoutId: ReturnType<typeof setTimeout> = setTimeout(function(){});
+  private _timeoutId: ReturnType<typeof setTimeout> = setTimeout(function () {});
 
   private _moveTimeLimit = 2 * 60 * 1000; // milliseconds
 
--- a/src/app/plugins/connectfour.plugin.ts
+++ b/src/app/plugins/connectfour.plugin.ts
@@ -144,6 +144,11 @@ class ConnectFourGame {
   private _playingLion: boolean;
   private _aiDepth = 4;
 
+  // messy initialization since typescript won't let me use -1 or null
+  private _timeoutId: ReturnType<typeof setTimeout> = setTimeout(function(){});
+
+  private _moveTimeLimit = 2 * 60 * 1000; // milliseconds
+
   private _winner: Maybe<number> = null;
   private _tie: boolean = false;
   private _gameOver: boolean = false;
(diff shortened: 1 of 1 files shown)


src/app/plugins/connectfour.plugin.ts
```diff
@@ -144,6 +144,11 @@ class ConnectFourGame {
   private _playingLion: boolean;
   private _aiDepth = 4;
 
+  // messy initialization since typescript won't let me use -1 or null
```
joeydotdev, line 147: `ReturnType<...> | null` doesnt work here?

src/app/plugins/connectfour.plugin.ts
```diff
@@ -144,6 +144,11 @@ class ConnectFourGame {
...
+  private _timeoutId: ReturnType<typeof setTimeout> = setTimeout(function () {});
+
+  private _moveTimeLimit = 2 * 60 * 1000; // milliseconds
```
joeydotdev, line 150: what's this time limit in seconds? it'd be more helpful to have that number as a comment IMO

src/app/plugins/connectfour.plugin.ts
```diff
@@ -351,9 +367,14 @@ class ConnectFourGame {
     );
   }
 
-  private async _updateGameState(msg: IMessage): Promise<void> {
+  private async _updateGameState(msg: IMessage, timedOut: boolean): Promise<void> {
```
joeydotdev, line 370: we can make the second argument be optional and have it as an object. either named as `options` or `params` and have `timedOut` be an optional property in this object

this will allow you to simplify the invocation for consumers of this function that aren't coming from a `timedOut` source

joeydotdev (commented):
looks good overall but just some minor things that can help clean this up

src/app/plugins/connectfour.plugin.ts
tanndlin: We have a module in the package called [ms](https://github.com/cs-discord-at-ucf/lion/blob/03a650e26f1e4afc07dee0f93d1ec0a289d7d807/src/app/handlers/command.handler.ts#L74)

Make it a lot more readable

src/app/plugins/connectfour.plugin.ts
msabur: Good idea. Used `undefined` instead of `null`, since that matches `clearTimeout()`'s signature.

src/app/plugins/connectfour.plugin.ts
msabur: Definitely. I changed it to use ms instead.

joeydotdev (approved):
🚀

tanndlin (approved):
I'll leave this up for just a little longer in case anyone else wants to chime in

PR #726, opened 2022-08-25
Tanndlin/force single server

### LINKED ISSUE #723: Idiot proof server assertion
Since Lion is only supposed to run in **one** server, I propose that we enforce that.

We should error out instead of running if more than one guild is detected to prevent commands being run in the wrong server

@c650 whatcha think

tanndlin:
fixes #723

### DIFF
--- a/src/app/listener.ts
+++ b/src/app/listener.ts
@@ -31,11 +31,13 @@ export class Listener {
     this.container.clientService.on('ready', async () => {
       // Enforce only one server
       if (this.container.clientService.guilds.cache.size > 1) {
-        throw new Error(
+        this.container.loggerService.error(
           'More than one server detected. This bot is intented to run in only one server, ' +
             'and can cause issues when in multiple.\n Please remove this bot from other ' +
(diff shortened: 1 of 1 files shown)


tanndlin:
This would kill the client on ready if in multiple guilds

src/app/listener.ts
```diff
@@ -29,6 +29,18 @@ export class Listener {
...
     this.container.clientService.on('ready', async () => {
+      // Enforce only one server
+      if (this.container.clientService.guilds.cache.size > 1) {
+        this.container.loggerService.error(
+          'More than one server detected. This bot is intented to run in only one server, ' +
+            'and can cause issues when in multiple.\n Please remove this bot from other ' +
+            'servers, or create a new application for Lion.'
+        );
+
+        this.container.clientService.destroy();
```
joeydotdev, line 40: why not just throw Error? the program shouldnt be running if we're in this state

tanndlin: Destroy kills the client, I cant throw an unhandled error and crash it instead if you want

Issue #724, opened 2022-08-25
Send link to class after registering for it


tanndlin:
When using `!register aaa1234_hello` the bot should respond with a link to that channel `#aaa1234_hello` to make accessing easier

PR #727, opened 2022-08-25
Send links to registered classes

layyne:
oo prettyy

### DIFF
--- a/src/app/plugins/register.plugin.ts
+++ b/src/app/plugins/register.plugin.ts
@@ -86,7 +86,6 @@ export default class RegisterPlugin extends Plugin {
   }
 
   private async _giveResultsToUser(results: string[], args: string[], message: IMessage) {
-    // let numSuccessfulClasses = 0;
     const validClasses: string[] = [];
     const invalidClasses: string[] = [];
 
--- a/src/app/plugins/register.plugin.ts
+++ b/src/app/plugins/register.plugin.ts
@@ -100,6 +100,7 @@ export default class RegisterPlugin extends Plugin {
     });
 
     const embedAuthorData = this.container.messageService.getEmbedAuthorData(message);
+    const shouldShowAuthorOnRegister = true;
 
     if (validClasses.length > 0) {
       // List of channel links, one per line
@@ -113,7 +114,7 @@ export default class RegisterPlugin extends Plugin {
         .setTitle('Successfully registered')
         .setDescription(validChannels)
         .setColor('#a3be8c');
-      if (Constants.ShouldShowAuthorOnRegister) {
+      if (shouldShowAuthorOnRegister) {
         embed.setAuthor(embedAuthorData);
       }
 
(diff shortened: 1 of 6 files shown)


tanndlin:
fixes #724 kek

tanndlin (commented):
Does it also send a link when accepting a fuzzy search recommendation?

layyne:
> Does it also send a link when accepting a fuzzy search recommendation?

It already does that. It's not in an embed currently, though.

src/app/plugins/register.plugin.ts
```diff
@@ -99,24 +99,30 @@ export default class RegisterPlugin extends Plugin {
...
         .map((validClass) => {
           return this.container.classService.findClassByName(validClass);
         })
         .join('\n');
+
+      const embed = new MessageEmbed()
+        .setTitle('Successfully registered')
+        .setDescription(validChannels)
+        .setColor('#a3be8c');
+      if (shouldShowAuthor) {
```
tanndlin, line 120:
```suggestion
      if (Constants.ShouldShowAuthorOnRegister) {
```

Just use the constant here instead of making a redundant variable

src/app/plugins/register.plugin.ts
```diff
@@ -134,18 +140,31 @@ export default class RegisterPlugin extends Plugin {
...
       embedMessages.map((embedData) => {
+        const cutoffEmbed = new MessageEmbed()
+          .setTitle('Successfully registered')
+          .setDescription(String(embedData.emojiData[0].args.classChan) || 'N/A')
+          .setColor('#a3be8c');
+        const closingEmbed = new MessageEmbed()
+          .setTitle('Closed registering offer')
+          .setDescription(String(embedData.emojiData[0].args.classChan) || 'N/A')
+          .setColor('#bf616a');
+        if (shouldShowAuthor) {
```
tanndlin, line 151: saa

src/services/class.service.ts
```diff
@@ -58,6 +59,12 @@ export class ClassService {
       const embeddedMessage: discord.MessageEmbed = new discord.MessageEmbed();
 
       embeddedMessage.setColor('#0099ff').setTitle(`${invalidClass} Not Found`);
+      if (Constants.ShouldShowAuthorOnRegister) {
+        embeddedMessage.setAuthor({
+          name: message.member?.displayName ?? '',
+          iconURL: message.author.avatarURL() ?? '',
```
tanndlin, line 65: might want to consider making a function that returns a `MessageEmbedAuthor` instead of having this everywhere

src/app/plugins/register.plugin.ts
```diff
@@ -99,24 +99,30 @@ export default class RegisterPlugin extends Plugin {
       }
     });
 
+    const authorData = {
+      name: message.member?.displayName ?? '',
+      iconURL: message.author.avatarURL() ?? '',
+    };
+    const shouldShowAuthor = Constants.ShouldShowAuthorOnRegister;
```
tanndlin, line 106:
```suggestion
```

tanndlin (changes requested):

src/common/constants.ts
```diff
@@ -102,7 +102,7 @@ const Constants = {
     Moderator: 'Moderator',
     Suspended: 'Suspended',
   },
-  ShouldShowAuthorOnRegister: true,
+  ShouldShowAuthorOnRegister: false,
```
tanndlin, line 105: Was this on purpose?

joeydotdev: i don't think this constant should live here since its just a tweak for how the internal of one plugin/service should run. is it possible to have it live in either of those files?

tanndlin: we can move it for now, this idea sparked #729 so eventually it will be moved there

tanndlin (approved):

PR #731, opened 2022-08-27
Enable equation plugin in math_lounge

### LINKED ISSUE #730: Allow equation plugin in math_lounge
As title. Requested by maximo#5992 in the server.

layyne:
Resolves #730.


src/common/plugin.ts
```diff
@@ -18,8 +18,11 @@ export abstract class Plugin implements IPlugin {
 
   public pluginAlias?: string[];
 
+  // Only provide ONE of these!
   public pluginChannelName?: string;
+  public pluginChannelNames?: string[];
```
c650, line 23: Why special case for 1? Array of one should work

src/common/plugin.ts
```diff
@@ -45,15 +48,38 @@ export abstract class Plugin implements IPlugin {
     const channelName = channel.name;
     const categoryName = channel.parent?.name.toLowerCase();
 
-    if (this.pluginChannelName && this.pluginChannelName !== channelName) {
+    if (this.pluginCategoryName) {
+      // If category matches, move on
+      if (this.pluginCategoryName.toLowerCase() !== categoryName) {
+        // If channel name matches, move on
+        if (this.pluginChannelNames) {
+          if (!this.pluginChannelNames.includes(channelName)) {
```
c650, line 56: Why is this so complicated

c650 (changes requested):

src/common/plugin.ts
suneettipirneni: As for some constructive feedback, this code could simplify some of a lot its branches by using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).

layyne: I'm also disappointed in how complicated this turned out; however, I couldn't figure out how to avoid it. I couldn't simply combine the null checks and equality checks, as it's necessary to know whether each field is defined, as well as whether the current channel matches it if it is. Without that granularity, there's no way I could see to prevent certain checks from occurring when they shouldn't, such as checking whether the channel name matched when the category name was already determined to match.
The short explanation for the complexity of this code is that there are 5 possible outputs depending on which restrictions the developer places on the plugin.
A solution I can think of is to scrap the optional list of channel names, or otherwise get rid of the old single channel name in favor of it (which would involve lots of refactoring).

src/common/plugin.ts
```diff
@@ -45,12 +48,31 @@ export abstract class Plugin implements IPlugin {
     const channelName = channel.name;
     const categoryName = channel.parent?.name.toLowerCase();
 
-    if (this.pluginChannelName && this.pluginChannelName !== channelName) {
-      const id = this.container.guildService.getChannel(this.pluginChannelName).id;
-      return `Please use this command in the <#${id}> channel.`;
+    let allowedChannels: string[] | undefined;
+    if (this.pluginChannelName) {
```
c650, line 52: This is much better. I still think it can be a bit cleaner than all these cases though

c650 (changes requested):

src/common/plugin.ts
layyne: Cleaned up more. Plugin developers can now provide allowed channels either using the single string field, or any number of strings in the array field.

src/common/plugin.ts
```diff
@@ -45,12 +47,25 @@ export abstract class Plugin implements IPlugin {
...
-    if (this.pluginChannelName && this.pluginChannelName !== channelName) {
-      const id = this.container.guildService.getChannel(this.pluginChannelName).id;
-      return `Please use this command in the <#${id}> channel.`;
+    const allowedChannels = [...(this.pluginChannelNames ?? [])];
+    if (this.pluginChannelName) {
+      allowedChannels.push(this.pluginChannelName);
     }
 
-    if (this.pluginCategoryName && this.pluginCategoryName.toLowerCase() !== categoryName) {
+    if (allowedChannels.length !== 0) {
```
c650, line 55: So where are you actually checking that the channelName is in the allowedChannels?

layyne: oh LOL that line escaped somehow, added back

src/common/plugin.ts
```diff
@@ -45,12 +47,27 @@ export abstract class Plugin implements IPlugin {
...
+      if (!allowedChannels.includes(channelName)) {
+        const validChannelNames = allowedChannels
+          .map((name) => this.container.guildService.getChannel(name))
+          .join(', ');
+        return `Please use this command in ${
```
c650, line 60: I think you need to do the category check before denying the command because it’s not in the channel list

src/common/plugin.ts
```diff
@@ -20,6 +20,8 @@ export abstract class Plugin implements IPlugin {
 
   public pluginChannelName?: string;
```
c650, line 22: Can you make pluginChannelName a string | string[] 

that would cut some of this down

c650 (changes requested):
I’m suspicious of how you tested this. And how you are testing it between changes

src/common/plugin.ts
layyne: Yep, realized this as soon as I pushed the last change(s).

src/common/plugin.ts
layyne: Done; I'm just not a huge fan of this naming scheme.

layyne:
> I’m suspicious of how you tested this. And how you are testing it between changes

I slacked on testing everything before these last 2 commits, but I'm fairly certain everything works now. I tried plugins using one of each possible combination of `pluginChannelName`, `pluginChannelNames`, and `pluginCategoryName`, both in channels the plugin was allowed in and blocked in. After the second-to-last commit, the channel/category checking worked exactly as before, as far as I could tell.

If you'd like, I can draft up a comprehensive test plan for these changes.

src/common/plugin.ts
```diff
@@ -67,7 +71,7 @@ export abstract class Plugin implements IPlugin {
             : `one of the following channels: ${validChannelNames}.`
         }`;
       }
-    } else if (this.pluginCategoryName && this.pluginCategoryName !== categoryName) {
+    } else if (this.pluginCategoryName && this.pluginCategoryName.toLowerCase() !== categoryName) {
       return `Please use this command in the \`${this.pluginCategoryName}\` category.`;
```
c650, line 73: i dont think you need this branch anymore

c650 (changes requested):

c650:
in the initial post of this PR, please describe how you tested this and which cases you covered

src/common/plugin.ts
layyne: It's definitely necessary for plugins that define a category but no channels. The category can fail to match whether the plugin also has channels defined or not.

tanndlin:
This really should be 2 PRs

The majority of the work here is enabling more than one designated channel name for a plugin
And the other is just implementing that on one plugin.

If you just want to accomplish your goal of `Enable equation plugin in math_lounge` there's not really a reason to convert the `pluginChannelName` to an array. Just put math_lounge and you might have to change one line in the command handler to check if the category **oor** the channel is valid

tanndlin:
deyeet?

PR #739, opened 2022-09-19
add slash commands

suneettipirneni:
Plugins that implement `ISlashCommand` will support slash commands.


.env.sample
```diff
@@ -12,5 +12,6 @@ WARNINGS_THRESH=x
 PAPERTRAIL_HOST=x
 PAPERTRAIL_PORT=x
 TWITTER_BEARER_TOKEN=x
+GUILD_ID=852656268090802263=x
```
c650, line 15: Nix

src/app/handlers/react.handler.ts
```diff
@@ -42,7 +49,7 @@ export class ReactHandler extends Handler {
     await this.container.warningService.deleteChan(user.id);
   }
 
-  private _handleClassChannelPinRequest(message: IMessage, channel: TextChannel) {
+  private _handleClassChannelPinRequest(message: IMessage | PartialMessage, channel: TextChannel) {
```
c650, line 52: Why do you need this

src/app/jobs/poll.job.ts
```diff
@@ -21,7 +21,7 @@ export class PollJob extends Job {
         .map(async (poll) => {
           const embed = container.pollService.createResultEmbed(poll);
 
-          await poll.msg.channel.send({embeds:[embed]}).then(() => {
```
c650, line 24: Nix

src/app/listener.ts
```diff
@@ -42,6 +45,38 @@ export class Listener {
...
+        })
+        .map((entry) => {
+          const [name, plugin] = entry;
+          const options = isSlashCommand(plugin)
+            ? (plugin as unknown as ISlashPlugin).parameters
+            : undefined;
+
+          return {
+            name: name,
+            description: plugin.description.substring(0, 99),
```
c650, line 61: 99?

.env.sample
suneettipirneni: This is used for local development of slash commands. Since slash commands need to be deployed, as developer will enter the guild ID they are developing in and lion will deploy commands in said guild so they can test them,

src/app/listener.ts
```diff
@@ -42,6 +45,38 @@ export class Listener {
...
+      // We only want to use guild commands for development, because global app commands
+      // can take over an hour to propogate.
+      if (process.env.NODE_ENV === Mode.Development) {
+        if (!process.env.GUILD_ID) {
+          throw new Error('You need to set the GUILD_ID in your .env file!');
+        }
+
+        await this.container.clientService.guilds.cache
+          .get(process.env.GUILD_ID)
+          ?.commands.set(commands);
```
c650, line 75: Why do the docs have all this REST garbage If you can do this natively

src/app/listener.ts
suneettipirneni: Slash commands have a description string length limit of 99 chars.

src/app/plugins/broadcast.plugin.ts
```diff
@@ -17,7 +23,7 @@ export default class BroadcastPlugin extends Plugin {
   public override minRoleToRun: RoleType = RoleType.Admin;
   public override commandPattern: RegExp = /((message|classes)\s.+|attach|confirm|cancel)/;
 
-  private _CHANS_TO_SEND: GuildChannel[] = [];
+  private _CHANS_TO_SEND: (GuildChannel | ThreadChannel)[] = [];
```
c650, line 26: Y

src/common/types.ts
```diff
@@ -95,7 +96,7 @@ export interface IPlugin extends IRunnable {
   usableInDM?: boolean;
   usableInGuild?: boolean;
   validate(message: IMessage, args: string[]): boolean;
-  hasPermission(message: IMessage): true | string;
+  hasPermission(message: IMessage | discord.CommandInteraction): true | string;
```
c650, line 99: I thought you could put permissions in the configuration for slash commands? Then we don’t need to do our own validation ?

src/services/guild.service.ts
```diff
@@ -9,7 +9,7 @@ export class GuildService {
     [Constants.Roles.Unverifed]: undefined,
   };
 
-  private _channelCache: Record<string, Maybe<GuildChannel>> = {
+  private _channelCache: Record<string, Maybe<GuildChannel | ThreadChannel>> = {
```
c650, line 12: Seems unrelated? Or is it because of the recent change … ugh. Should be a separate PR

c650 (changes requested):

src/app/listener.ts
suneettipirneni: ¯\_(ツ)_/¯

src/common/types.ts
suneettipirneni: Those validations are client-side validation checking the format. In our case we'd need to use lion's system for validation since it provides custom messages in chat for permissions errors.

.env.sample
c650: why not loop over the guilds like i do in https://github.com/cs-discord-at-ucf/lion/pull/738/files#diff-13c50b4cc67456d5ba9f4608265a1f770f856e4ab28225bf35ba2e4bfec7bea6

src/common/types.ts
```diff
@@ -223,6 +224,21 @@ export type ServerInfoType = 'MemberCount';
...
 export type Maybe<T> = T | undefined | null;
 
+export type MessageSendData =
+  | string
+  | discord.MessagePayload
+  | (discord.ReplyMessageOptions & { split?: false }); // Discord v13 change
+
+export type MessageEditData = string | discord.MessagePayload;
+
+export function isSlashCommand(plugin: unknown): boolean {
```
c650, line 234: does this function belong in types.ts ?

src/app/handlers/command.handler.ts
```diff
@@ -14,30 +30,40 @@ export class CommandHandler extends Handler {
     super();
   }
 
-  public async execute(message: types.IMessage): Promise<void> {
-    const command = this.build(message.content);
+  public async execute(message: types.IMessage | CommandInteraction): Promise<void> {
```
c650, line 33: instead of a union type, we should (upstream) populate some CommandPayloadCommon so that it has the same data whether it's from a '!' message or a slash command invocation

.env.sample
c650, line 15: i prefer that we automatically get the guild from the guild service on startup, de-register all commands, and then re-register whichever ones still exist.

c650 (changes requested):

c650:
just wondering where we wanna go with this?

src/app/handlers/command.handler.ts
suneettipirneni: I was thinking about this, and it feels like that would actually introduce more complexity. The reason being is that typescript unions automatically allow you access common properties. Also this is the only function where there's unioned command types. After this point they're separated and narrowed to their specific types. So another abstracted class would essentially be doing the same de-narrowing this function does but it would only be used in a single place.

src/app/handlers/command.handler.ts
```diff
@@ -1,8 +1,24 @@
...
+  | {
+      type: 'message';
+      message: types.IMessage;
+      plugin: types.IPlugin;
+      command: types.ICommand;
+      isDM: boolean;
+    }
+  | {
+      type: 'interaction';
+      message: CommandInteraction;
```
c650, line 18: do we want to have a `types.ICommandInteraction`?

src/app/handlers/command.handler.ts
```diff
@@ -14,30 +30,40 @@ export class CommandHandler extends Handler {
...
-    // checks to see if the user is actually talking to the bot
-    if (!command) {
-      return;
-    }
-
-    if (message.mentions?.everyone) {
-      message.reply('You cannot use a plugin, while pinging everyone.');
-      return;
-    }
+    let plugin;
```
c650, line 37: do you need to declare this here? it seems like you can tightly scope it to each branch of the if/else

src/app/handlers/command.handler.ts
```diff
@@ -114,25 +146,32 @@ export class CommandHandler extends Handler {
...
+        return;
+      }
+    }
+
+    let user;
+
+    if (message instanceof CommandInteraction) {
+      user = message.user.tag;
+    } else {
+      user = message.author.tag;
```
c650, line 174: can we just make this a const ... = ternary

src/app/handlers/command.handler.ts
```diff
@@ -143,13 +182,26 @@ export class CommandHandler extends Handler {
...
+      // If it's a regular command, use the regular executor.
+      if (commandData.type === 'message') {
+        await plugin.execute(commandData.message, commandData.command.args);
+      }
+
+      // If it's a slash command, use the slash runner.
+
+      // This should never happen.
+      if (!types.isSlashCommand(plugin)) {
+        throw new Error('Cannot invoke slash command on a non-slash-command plugin.');
```
c650, line 201: let's just log and return instead of throwing

src/app/handlers/command.handler.ts
```diff
@@ -143,13 +182,26 @@ export class CommandHandler extends Handler {
...
+      }
+
+      await (plugin as unknown as ISlashPlugin).run(message as CommandInteraction);
```
c650, line 204: ?

src/app/index.d.ts
```diff
@@ -23,6 +25,7 @@ declare global {
       TWITTER_BEARER_TOKEN?: string;
       WOLFRAM_APPID?: string;
       TESTER_TOKEN?: string;
+      GUILD_ID?: Snowflake;
```
c650, line 28: you shouldnt need these changes anymore.

src/app/listener.ts
```diff
@@ -42,6 +45,29 @@ export class Listener {
...
 
+      const commands = Object.entries(this.container.pluginService.plugins)
+        .filter((entry) => {
+          const [, plugin] = entry;
+          return isSlashCommand(plugin);
+        })
+        .map((entry) => {
+          const [name, plugin] = entry;
+          const options = isSlashCommand(plugin)
+            ? (plugin as unknown as ISlashPlugin).parameters
```
c650, line 56: since you filter, do you need to do this?

src/app/listener.ts
```diff
@@ -42,6 +45,29 @@ export class Listener {
...
+            : undefined;
```
c650, line 57: use null instead for user-specified "undef"

src/app/listener.ts
```diff
@@ -42,6 +45,29 @@ export class Listener {
...
+          return {
+            name: name,
+            description: plugin.description.substring(0, 99),
+            options,
+          };
+        });
+
+      // Register commands for all guilds.
+      for (const guild of this.container.clientService.guilds.cache.values()) {
+        guild.commands.set(commands);
```
c650, line 68: is this synchronous ? 

also do we have to do a pass of unregistering? otherwise a command that gets removed could be left dangling.

src/common/plugin.ts
```diff
@@ -84,12 +85,18 @@ export abstract class Plugin implements IPlugin {
     const totalChannels = channels.length;
     channels.splice(this._numChannelsShown);
 
+    if (message instanceof CommandInteraction) {
+      this.container.guildService.getChannel('test')
```
c650, line 89: ?

src/common/types.ts
```diff
@@ -223,6 +224,21 @@ export type ServerInfoType = 'MemberCount';
 export type RoleTypeKey = keyof typeof RoleType;
 export type Maybe<T> = T | undefined | null;
 
+export type MessageSendData =
+  | string
+  | discord.MessagePayload
+  | (discord.ReplyMessageOptions & { split?: false }); // Discord v13 change
```
c650, line 230: can this go in a smaller change since it's a v13 thing?

tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "types": ["node", "jest"],
     "esModuleInterop": true,
     "noImplicitOverride": true,
+    "skipLibCheck": true,
```
c650, line 13: what's this do ?

src/app/listener.ts
tanndlin: Don't think TS updates types of an array after using a filter

src/app/listener.ts
tanndlin: Set should be a strict equals to, not add.
So any slash command that was not in the set will be removed

tanndlin:
Would it make more sense to have 2 separate handlers, one for slash one for legacy? Instead of binding them both to the same class and having to type check everywhere

joeydotdev:
> Would it make more sense to have 2 separate handlers, one for slash one for legacy? Instead of binding them both to the same class and having to type check everywhere

+1 to this. Especially if we intend to do any enhancements on either of the types of commands, it's easier way to introduce regressions if the code is tightly coupled with each other. I think it's fine to duplicate code in this scenario if it's required

src/app/listener.ts
c650: cleaner to do a map after the filter then, if filter_map doesnt exist in ts

tsconfig.json
rob-3: https://www.typescriptlang.org/tsconfig#skipLibCheck

src/common/plugin.ts
rob-3: I'm removing this in my branch; I think it was just for testing

src/app/handlers/command.handler.ts
rob-3: What would be the idea here? Just a wrapper around the `CommandInteraction` type? If so, seems like something we can add later if we need it.

src/app/handlers/command.handler.ts
rob-3: Doen

src/app/handlers/command.handler.ts
rob-3: There's some trickiness going on with how exactly we want to tag slash command plugins vs classic plugins. I can rework some of this if you don't like how it is currently.

src/app/listener.ts
rob-3: This breaks the typings on line 68

src/app/listener.ts
rob-3: I'm adding await just to be sure here, +1 on tanndlin's comment

src/common/types.ts
rob-3: Sure

PR #740, opened 2022-09-25
Link #class_invite channel in class channel welcome message

msabur:
Makes "#class_invite" clickable in the welcome message of a class channel.

### DIFF
--- a/src/app/plugins/addclasschans.plugin.ts
+++ b/src/app/plugins/addclasschans.plugin.ts
@@ -128,7 +128,7 @@ export default class AddClassChannelsPlugin extends Plugin {
           })
           .then(async (newChan: GuildChannel) => {
             await (newChan as TextChannel).send({
-              embeds: [this._createFirstMessage(newChan.name)],
+              embeds: [await this._createFirstMessage(newChan.name)],
             });
           });
       } catch (e) {
@@ -139,10 +139,22 @@ export default class AddClassChannelsPlugin extends Plugin {
     this._STATE = [];
   }
 
-  private _createFirstMessage(chanName: string): MessageEmbed {
-    const invChan = this.container.guildService
-      .get()
-      .channels.cache.find((c) => c.name === Constants.Channels.Info.ClassInvite);
+  private async _createFirstMessage(chanName: string): Promise<MessageEmbed> {
+    const getInvChan = async () => {
+      let ret;
+      ret = this.container.guildService
+        .get()
+        .channels.cache.find((c) => c.name === Constants.Channels.Info.ClassInvite);
+      if (!ret) {
+        ret = (await this.container.guildService
+          .get()
+          .channels.fetch())
+          .find((c) => c.name === Constants.Channels.Info.ClassInvite);
+      }
+      return ret;
+    }
+
+    let invChan = await getInvChan();
     const embed = new MessageEmbed();
     embed.setTitle(`Welcome to ${chanName}!`);
     embed.setThumbnail(Constants.LionPFP);
--- a/src/app/plugins/addclasschans.plugin.ts
+++ b/src/app/plugins/addclasschans.plugin.ts
@@ -152,9 +152,9 @@ export default class AddClassChannelsPlugin extends Plugin {
           .find((c) => c.name === Constants.Channels.Info.ClassInvite);
       }
       return ret;
-    }
+    };
 
-    let invChan = await getInvChan();
+    const invChan = await getInvChan();
     const embed = new MessageEmbed();
     embed.setTitle(`Welcome to ${chanName}!`);
     embed.setThumbnail(Constants.LionPFP);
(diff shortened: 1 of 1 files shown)


c650:
It’s better if you look up the channel in the guild instead of hardcoding an ID

src/app/plugins/addclasschans.plugin.ts
```diff
@@ -140,10 +140,15 @@ export default class AddClassChannelsPlugin extends Plugin {
   }
 
   private _createFirstMessage(chanName: string): MessageEmbed {
+    const invChan = this.container.guildService
+      .get()
+      .channels.cache.find((c) => c.name === Constants.Channels.Info.ClassInvite);
```
c650, line 145: We probably have to fetch. I doubt the channel will be in the cache because the class invite channel is stale 99% of the time

src/app/plugins/addclasschans.plugin.ts
```diff
@@ -23,7 +23,7 @@ export default class AddClassChannelsPlugin extends Plugin {
   private _CATEGORIES: string[] = ['cs', 'it', 'ee', 'csgrad', 'eegrad', 'gened'];
   private _NEW_CHAN_MESSAGE =
     'Welcome to the class!\n\n' +
-    '**If it has not been done so already, please post the #class_invite ' +
+    '**If it has not been done so already, please post the [placeholder] ' +
```
c650, line 26: Put this in a method so that you can use a format string instead of relying on substitution

c650 (changes requested):

src/app/plugins/addclasschans.plugin.ts
msabur: Would you like `_NEW_CHAN_MESSAGE` to be declared in a method? Since we are fetching the channel ID, it would require an async method which seems messy.

c650:
```
private _getNewChanMessage(classInviteChanId) {
  return `.... ${classInviteChanID} ...`
}
```

src/app/plugins/addclasschans.plugin.ts
```diff
@@ -139,11 +141,28 @@ export default class AddClassChannelsPlugin extends Plugin {
     this._STATE = [];
   }
 
-  private _createFirstMessage(chanName: string): MessageEmbed {
+  private async _createFirstMessage(chanName: string): Promise<MessageEmbed> {
+    const getInvChan = async () => {
+      let ret;
```
c650, line 146:
```
    
     const  ret = this.container.guildService
        .get()
        .channels.cache.find((c) => c.name === Constants.Channels.Info.ClassInvite);
      if (!ret) {
        return (await this.container.guildService
          .get()
          .channels.fetch())
          .find((c) => c.name === Constants.Channels.Info.ClassInvite);
      }
```

c650: i prefer you move getInvChan to its own (private) method on the class and, keep _createFirstMessage non-async, and pass the channel id of class_invite (from getInvChan) into _createFirstMessage.

c650 (changes requested):

src/app/plugins/addclasschans.plugin.ts
msabur: Done. I also changed `getInvChan` to only return the channel id since that's all we use. If that's undesirable, I'll change it back.

src/app/plugins/addclasschans.plugin.ts
```diff
@@ -49,27 +49,41 @@ export default class AddClassChannelsPlugin extends Plugin {
 
   private _getNewChanMessage(id: String): string {
```
c650, line 50: You use capital String in only this place

src/app/plugins/addclasschans.plugin.ts
```diff
@@ -68,6 +47,42 @@ export default class AddClassChannelsPlugin extends Plugin {
...
+      .get()
+      .channels.cache.find((c) => c.name === Constants.Channels.Info.ClassInvite);
+    if (!ret) {
+      return (await this.container.guildService
+        .get()
+        .channels.fetch())
+        .find((c) => c.name === Constants.Channels.Info.ClassInvite);
+    }
+    return ret;
+  };
```
tanndlin, line 84: Sorry been gone all weekend so, just seeing this
but `GuildService` has a `getChannel` method that takes a string name of a channel

tanndlin:
I'm not really sure why this is such a big change, the only change really needed is the exact line where the channel should be linked.

```
\`**If it has not been done so already, please post the ${this.container.guildService.getChannel(Constants.Channels.Info.ClassInvite} \`
```
or just store the guild channel first idc

src/app/plugins/addclasschans.plugin.ts
msabur: True, I should use that instead. I was thinking of cache misses, but that shouldn’t happen since we have the GUILDS intent.

c650 (approved):

tanndlin (approved):

PR #741, opened 2022-10-01
Counting Leaderboard

tanndlin:
Only increments existing documents in the collection

I ran a one-time script to count up the messages and populate that collection


src/common/types.ts
```diff
@@ -229,3 +229,8 @@ export interface IUserPoints {
   numPoints: number;
   lastKingCrowning: Date;
 }
+
+export interface ICountingEntry {
```
c650, line 233: no guildID ?

src/app/handlers/counting.handler.ts
```diff
@@ -38,6 +39,9 @@ export class CountingHandler extends Handler {
         await message.react(this._buzzEmoji);
       }
 
+      // Increment count in counting leaderboard
+      const userDoc = await this._getUserCountingDoc(message.author.id);
+      await CountingLeaderboardModel.updateOne(userDoc, { $inc: { count: 1 } });
```
c650, line 44: is there no updateOrInsert type thing? ?

msabur (changes requested):
There’s a typo in filename `coutingLeaderboard.plugin.ts`.

src/app/handlers/counting.handler.ts
tanndlin: I tried, but I don't think I can do it with an insert too

PR #762, opened 2023-07-04
Deleted some plugins, jobs, and a service

bluffdev:
I don't think we need any of these


tanndlin:
I would leave the price plugin that one gets used a bit in the finance channel

bluffdev:
> I would leave the price plugin that one gets used a bit in the finance channel

It hasn’t been used in 3 months + people can just Google this info in 5 seconds

bluffdev:
> pls include a test plan
> 
> for sometihng as simple as this, all that really needs to be is "i ran the bot and it came up"

I am just trying to make the bot leaner by removing things that we don’t need. This will make it easier to migrate to discordjs 14 and add slash commands which both require modifying a lot of files

bluffdev:
> He's asking if you have at least set up the bot and got it to run in a test environment.
> Just to make sure its not bricking the bot with some weird side effect

Oh yes it’s running fine

c650 (approved):

c650:
in the future, include your test plan in the top message of the PR

PR #768, opened 2023-08-18
Suneet's slash commands v2

rob-3:


rob-3:
I'm going to push some changes tomorrow for this and then should be good for review again.

tanndlin:
I'll take a look when I get back to Orlando

rob-3:
Ok, this is ready for a quick review. I only really did the minimum to get it working again and fix the low-hanging fruit. I have a few remaining concerns:
* There's some serious hacks around how slash commands are invoked, see the casting to unknown
* Do we even want plugins to be able to support both slash and classic commands? Maybe we should require new plugins to supersede the old ones

src/app/handlers/command.handler.ts
```diff
@@ -114,42 +145,57 @@ export class CommandHandler extends Handler {
...
-      await plugin.execute(message, command.args);
+
+      // If it's a regular command, use the regular executor.
+      if (commandData.type === 'message') {
+        await plugin.execute(commandData.message, commandData.command.args);
+      }
+
+      // If it's a slash command, use the slash runner.
+
+      // This should never happen.
```
joeydotdev, line 192: I don't understand this comment. Why shouldn't it ever happen?

src/common/types.ts
```diff
@@ -221,6 +222,21 @@ export type ServerInfoType = 'MemberCount';
...
 export type Maybe<T> = T | undefined | null;
 
+export type MessageSendData =
+  | string
+  | discord.MessagePayload
+  | (discord.ReplyMessageOptions & { split?: false }); // Discord v13 change
+
+export type MessageEditData = string | discord.MessagePayload;
+
+export function isSlashCommand(plugin: unknown): boolean {
```
joeydotdev, line 232: You can use a type guard and this will simplify the type casting you're doing in all the consumers that are calling this function

`export function isSlashCommand(plugin: unknown): plugin is ISlashPlugin`

joeydotdev:
> * There's some serious hacks around how slash commands are invoked, see the casting to unknown

This should be resolved when we use type guards



> * Do we even want plugins to be able to support both slash and classic commands? Maybe we should require new plugins to supersede the old ones

I think Discord as a platform is trying to move away from the classic commands and get people on board to slash commands. As time progresses, I think it will be pointless to be supporting Slash commands when customer expectations change on how they can interact with the bot.

I think it might make sense to just have a series of other PRs that are targeted towards porting existing plugins to be using Slash commands. We could introduce a new file extension that signals that this is a slash plugin and deprecate the existing extension (and force fail any PR that contains a file add with this extension). Up for discussion on path to migrating existing plugins, but I don't think it's worth the added complexity to support both types of interactivity

src/app/handlers/command.handler.ts
rob-3: Basically because of lines 186-188 only slash commands should fall through. So if we get something that doesn't look like a slash command, something is wrong. I agree this logic is laid out in a bit of a strange way.

src/common/types.ts
rob-3: True

rob-3:
> I think it might make sense to just have a series of other PRs that are targeted towards porting existing plugins to be using Slash commands.

This was my thought too.

> We could introduce a new file extension that signals that this is a slash plugin and deprecate the existing extension (and force fail any PR that contains a file add with this extension). Up for discussion on path to migrating existing plugins, but I don't think it's worth the added complexity to support both types of interactivity

The other thing that is possible is discriminating on whether a method name exists. I think rn the difference is basically `plugin.run()` and `plugin.execute()`. We could allow existing plugins to add support for the slash way and then eventually start deprecating the other way with warnings. That way we wouldn't have to copy-paste plugin code across files.

Happy to rewrite this either way though

joeydotdev:
> The other thing that is possible is discriminating on whether a method name exists. I think rn the difference is basically `plugin.run()` and `plugin.execute()`. We could allow existing plugins to add support for the slash way and then eventually start deprecating the other way with warnings. That way we wouldn't have to copy-paste plugin code across files.

It'll be harder to track the overall progress of the migration if you're using the same file extension and just creating a new method on each plugin. It might be confusing to have it in this state as I would think migration would take awhile since we don't have anyone grinding on the project rn.

I think it becomes more confusing than it is beneficial when you add method names to the original plugin interface for this scenario. Especially if we were to do something like ` .run()` / `.execute()` to differentiate between original/slash commands. I feel being explicit in stating its a slash command (somewhere) is better than trying to modify the original contract

@tanndlin @c650 thoughts on this?

rob-3:
> It'll be harder to track the overall progress of the migration if you're using the same file extension and just creating a new method on each plugin. It might be confusing to have it in this state as I would think migration would take awhile since we don't have anyone grinding on the project rn.

This makes sense. Should I take the opportunity to clean up/rewrite the interface then?

> I think it becomes more confusing than it is beneficial when you add method names to the original plugin interface for this scenario. Especially if we were to do something like  .run() / .execute() to differentiate between original/slash commands. I feel being explicit in stating its a slash command (somewhere) is better than trying to modify the original contract

Agreed.

tanndlin:
> It'll be harder to track the overall progress of the migration if you're using the same file extension and just creating a new method on each plugin. It might be confusing to have it in this state as I would think migration would take awhile since we don't have anyone grinding on the project rn.

I mean, for now we could have a readme with checkboxes for the status of migration for each plugin, and each PR would check the plugins it converts. That seems like it would be the least invasive way to keep track of the progress without doubling our file count.

> I think it becomes more confusing than it is beneficial when you add method names to the original plugin interface for this scenario. Especially if we were to do something like ` .run()` / `.execute()` to differentiate between original/slash commands. I feel being explicit in stating its a slash command (somewhere) is better than trying to modify the original contract

Maybe we just change the name of run to something more explicit like you said. Just check to see if a method called `runSlashCommand` (or maybe something a little less on the nose) and call it if it exists on the plugin.

_Ideally_ we just overwrite the current execute function for all the plugins, but that would require a mega-pr to do all the plugins at once. I think for now having the duplicate `run` method is fine, and once all plugins are migrated we either:

1. Remove all the `execute` methods and keep the `run` methods as is
2. Or Remove all `execute` methods and rename the `run` methods back to `execute`

src/app/handlers/command.handler.ts
```diff
@@ -114,42 +144,57 @@ export class CommandHandler extends Handler {
...
     }
 
+    const user = message instanceof CommandInteraction ? message.user.tag : message.author.tag;
+
     const pEvent: types.IPluginEvent = {
       status: 'starting',
       pluginName: plugin.name,
-      args: command.args,
-      user: message.author.tag,
+      args: [],
```
tanndlin, line 177:
```suggestion
      args: 
          message instanceof CommandInteraction ? 
              commandData.command.args : commandData.command.options,
```
Would need a good way of `toString`ing the options though, since its basically an unsorted collection

Probably something like this
```ts
commandData.command.options.map((option) => `${option.name}: ${option.value}`)
```
I'm not 100% familiar slash commands yet so not sure if thats fully correct

c650:
> > The other thing that is possible is discriminating on whether a method name exists. I think rn the difference is basically `plugin.run()` and `plugin.execute()`. We could allow existing plugins to add support for the slash way and then eventually start deprecating the other way with warnings. That way we wouldn't have to copy-paste plugin code across files.
> 
> It'll be harder to track the overall progress of the migration if you're using the same file extension and just creating a new method on each plugin. It might be confusing to have it in this state as I would think migration would take awhile since we don't have anyone grinding on the project rn.
> 
> I think it becomes more confusing than it is beneficial when you add method names to the original plugin interface for this scenario. Especially if we were to do something like ` .run()` / `.execute()` to differentiate between original/slash commands. I feel being explicit in stating its a slash command (somewhere) is better than trying to modify the original contract
> 
> @tanndlin @c650 thoughts on this?

i like this plan of the intermediate separate interface to roll plugins over slowly. let's do it!

rob-3:
> i like this plan of the intermediate separate interface to roll plugins over slowly. let's do it!

Current `IPlugin` is this:
```ts
export interface IPlugin extends IRunnable {
  description: string;
  usage: string;
  // ...
  hasPermission(message: IMessage | discord.CommandInteraction): true | string;
  execute(message: IMessage, args?: string[]): Promise<void> | void;
}
```

(1) Do we want to have an interface like this:
```ts
interface SlashCommand {
  description: string;
  usage: string;
  // ...
  hasPermission(message: IMessage | discord.CommandInteraction): true | string;
  executeSlashCommand(some: Some, args: Args): Promise<void> | void;
}
```
or do we want a brand new interface written?

(2) Do we want migrated plugins to have a different file extension/location?

joeydotdev:
The SlashCommand interface works with me though one nitpick is to rename `executeSlashCommand` to just `execute`. feels a bit redundant to include `slash` 


Btw @tanndlin why are we returning `true | string` for `hasPermission` ? that is confusing to me

> (2) Do we want migrated plugins to have a different file extension/location?

Yeah i think i would prefer another directory and whenever someone migrates a legacy command plugin, they can just delete the file from the legacy folder. Super simple to keep count if we separate it out through directories

joeydotdev:
And also, do slash commands have incoming messages that trigger? if not it doesn't make sense for `hasPermission` to accept an `IMessage`

rob-3:
> And also, do slash commands have incoming messages that trigger? if not it doesn't make sense for `hasPermission` to accept an `IMessage`

They take an `Interaction` of some form, so not quite.

rob-3:
> The SlashCommand interface works with me though one nitpick is to rename `executeSlashCommand` to just `execute`. feels a bit redundant to include `slash`

This can be done, but then it's not as simple to dynamically tell if a plugin is slash or not. We would need to have a different file extension or a flag on the plugin or something.

> Yeah i think i would prefer another directory and whenever someone migrates a legacy command plugin, they can just delete the file from the legacy folder. Super simple to keep count if we separate it out through directories

Ok sounds good; I will make a separate directory and put slash commands in there.

rob-3:
I will try to minimize the amount of deviations from the `IPlugin` interface, but obviously some things like the message-based stuff will have to be reworked. Lmk ASAP if anyone disagrees with the discussion above and wants to do things another way.

suneettipirneni:
hasPermission(), for running a slash command doesn’t really fit. Slash command permissions are controlled server side and a slash command can only specify the default permissions (which can be overridden by server admins). It can’t conditionally be executed from based on guild permissions code.

joeydotdev:
I think we're all in agreement on the high level details. I'll leave it to you for the rest! Feel free to include a small slash plug-in command as poc in this diff. Excited to see it 😁

joeydotdev:
> hasPermission(), for running a slash command doesn’t really fit. Slash command permissions are controlled server side and a slash command can only specify the default permissions (which can be overridden by server admins). It can’t conditionally be executed from based on guild permissions code.

If i want a slash command to only be executed in certain channels or by certain groups, how does that look?

suneettipirneni:
> > hasPermission(), for running a slash command doesn’t really fit. Slash command permissions are controlled server side and a slash command can only specify the default permissions (which can be overridden by server admins). It can’t conditionally be executed from based on guild permissions code.
> 
> If i want a slash command to only be executed in certain channels or by certain groups, how does that look?

You can edit it within the discord settings or specify the default permissions for guild channels when creating a slash command.

https://support.discord.com/hc/en-us/articles/4644915651095-Command-Permissions

tanndlin:
> Btw @tanndlin why are we returning `true | string` for `hasPermission` ? that is confusing to me

I believe its true if the user has permission, or a string reason why they do not

suneettipirneni:
I was looking back at this branch after a while, and imo the best thing to do to enforce a separation between slash commands and text commands is to use a tagged union with a `type` discriminator. This allows type narrowing where the command type is ambiguous without the need for a dedicated file extension or some form of duck-typing the object shape. This also would require no retroactive changes to existing plugins.

We can refactor `IPlugin` to be a union between a text command and a slash command:
```ts
export interface TextCommand extends IRunnable {
  // For backwards compatability, if `type` is missing
  // we assume it's a text command.
  type?: 'text';
  description: string;
  usage: string;
  commandName: string;
  pluginAlias?: string[];
  permission: ChannelType;
  pluginChannelName?: string;
  usableInDM?: boolean;
  usableInGuild?: boolean;
  validate(message: IMessage, args: string[]): boolean;
  hasPermission(message: IMessage): true | string;
  execute(message: IMessage, args?: string[]): Promise<void> | void;
}

export interface SlashCommand {
  type: 'slash';
  commandName: string;
  description: string;
  options: ApplicationCommandOptions;
  execute(interaction: Interaction): Promise<void> | void;
}

export type IPlugin = TextCommand | SlashCommand;
```

Because the union is tagged, typescript can automagically disambiguate the command kind when needed. For example:

```ts
function doSomethingWithPlugin(plugin: IPlugin) {
	
	// Shared properties don't need any disambiguation.
	console.log(`plugin name is ${plugin.commandName}`);

	if (plugin.kind === 'slash') {
		// `plugin` auto-inferred to be `SlashCommand`

		// Slash command specific logic
		return;
	}

	// `plugin` auto-inferred to be `TextCommand`
	
	// Text command-specific logic
}
```

rob-3:
By using a separate folder, we can sidestep all of the runtime type-checking. The only thing other I plan to implement now is validating the shape of the plugins. Are the core contributors ok with adding zod or something similar?

tanndlin:
> We can refactor `IPlugin` to be a union between a text command and a slash command:

I like this, but slash commands should still implement `IRunnable` so that feature flags still work for it

tanndlin:
> By using a separate folder, we can sidestep all of the runtime type-checking. The only thing other I plan to implement now is validating the shape of the plugins. Are the core contributors ok with adding zod or something similar?

That's good with me

src/app/listener.ts
```diff
@@ -69,6 +80,17 @@ export class Listener {
       notificationChannel.send({ embeds: [embed] });
     });
 
+    // Used to handle slash commands.
+    this.container.clientService.on('interactionCreate', (interaction: Interaction) => {
+      // If it's not a command, we don't care.
+      if (!interaction.isCommand()) {
```
suneettipirneni, line 86: Commands can be user or message context menu commands, which isn't supported in lion yet. So we should specifically check against those cases just to be safe.
```suggestion
      if (!interaction.isChatInputCommandInteraction()) {
```

src/common/slash.ts
```diff
@@ -0,0 +1,20 @@
...
+  parameters?: ApplicationCommandOptionData[];
+  execute({
+    interaction,
+    container,
+  }: {
+    interaction: CommandInteraction;
+    container: IContainer;
+  }): void | Promise<void>;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any;
```
suneettipirneni, line 19: Is there a specific reason this is needed?

src/app/listener.ts
```diff
@@ -42,6 +44,15 @@ export class Listener {
 
       this.container.loggerService.info(`Loaded ${this.container.jobService.size()} jobs...`);
 
+      const slashCommandUploads = Object.entries(slashCommands).map(([key, command]) => {
+        return {
+          name: key,
+          description: command.description.substring(0, 99),
+        };
+      });
```
suneettipirneni, line 52: This should take into account `SlashCommand#parameters`
```suggestion
      const slashCommandUploads = Object.entries(slashCommands).map(([key, command]) => {
        return {
          name: key,
          description: command.description.substring(0, 99),
          options: command.parameters,
        };
      });
```

suneettipirneni (changes requested):

src/app/listener.ts
rob-3: This doesn't exist in Discord.js v13.8 which we're on currently. Not sure if we want to roll an update in this PR or not

src/app/listener.ts
```diff
@@ -69,6 +80,17 @@ export class Listener {
...
 
+    // Used to handle slash commands.
+    this.container.clientService.on('interactionCreate', (interaction: Interaction) => {
+      // If it's not a command, we don't care.
+      if (!interaction.isCommand()) {
+        return;
+      }
+
+      // We only need the slash command handler.
+      slashCommands[interaction.commandName].execute({ interaction, container: this.container });
```
suneettipirneni, line 91: The types for accessing a prop on a `Record<_,T>` using an arbitrary key are misleading, it should give `T | undefined` but `T` is given instead. Beyond the types, there also may be situations where the deployed commands on discord are de-synced with our local commands defined in the source which may trigger an undefined access error.

```suggestion
      slashCommands[interaction.commandName]?.execute({ interaction, container: this.container });
```

suneettipirneni (changes requested):

src/app/listener.ts
rob-3: Yep, will do

src/app/listener.ts
rob-3: I redid the Record type differently to fix this

src/common/slash.ts
suneettipirneni: I now understand the issue you're pointing out, however, I don't think this is an optimal solution. Using `[key: string]: any` causes the return type of methods within an object to always return `any`. For example, in the 8Ball plugin `this._createEmbed()` returns `any` despite being typed to return `MessageEmbed`.

Alternatively, there's two things we can do to avoid this:

- Continue to use plain objects, and have a separate export statement:
```ts
const replyOptions = [...];

const 8ballPlugin: ISlashCommand = {
	name: 'Magic Eight Ball',
	commandName: '8ball',
	...
	execute({ interaction }) {
    	interaction.reply({ embeds: [createEmbed()] });
  	},
};

function createEmbed() {
  return new MessageEmbed()
      .setTitle(plugin.name)
      .setThumbnail('https://publicdomainvectors.org/photos/Jarno_8_ball.png')
      .setDescription(getRandom(replyOptions));
}

export default 8ballPlugin;
```

- Use classes with `implements`
```ts
export default class Plugin implements ISlashCommand {
  commandName = 'Magic Eight Ball';
  name = '8ball';
  description = 'Get answers from a pseudorandom number generator.';
  execute({ interaction }: { interaction: CommandInteraction<CacheType>; container: IContainer; }): void | Promise<void> {
    interaction.reply({ embeds: [createEmbed()] });
  }

  createEmbed() {
    return new MessageEmbed()
        .setTitle(this.name)
        .setThumbnail('https://publicdomainvectors.org/photos/Jarno_8_ball.png')
        .setDescription(getRandom(replyOptions));
  }
}
```

I don't have a strong opinion on either of these, but both are more type-safe than the current approach.

src/app/listener.ts
suneettipirneni: While the updated type makes the types correct in this scenario it makes the incorrect in others (like when iterating over entries, the value can never be nullish). I think using a `Map<K,V>` instead of an object would avoid all these issues.

src/common/slash.ts
suneettipirneni: I believe the class error is due to the method params when using `implements` need explicit types.

src/app/listener.ts
suneettipirneni: Oh yeah I forgot this is still on v13. We can keep this as-is as it's logically equivalent to v14's `isChatInputCommandInteraction()`.

src/app/slash_plugins/8ball.plugin.ts
```diff
@@ -0,0 +1,35 @@
...
+    interaction.reply({ embeds: [createEmbed()] });
+  },
+};
+
+function createEmbed() {
+  return new MessageEmbed()
+    .setTitle(plugin.name)
+    .setThumbnail('https://publicdomainvectors.org/photos/Jarno_8_ball.png')
+    .setDescription(getRandom(replyOptions));
+}
```
suneettipirneni, line 35:
```suggestion
}

export default plugin;
```

suneettipirneni (changes requested):

src/app/listener.ts
rob-3: Ok, I changed this

src/common/slash.ts
```diff
@@ -0,0 +1,27 @@
...
+import { z } from 'zod';
+
+export const slashCommands: Map<string, ISlashCommand> = new Map();
+
+export const SlashCommand = z.object({
+  commandName: z.string(),
+  name: z.string(),
+  description: z.string(),
+  parameters: z.array(z.any()),
+  execute: z.function(),
```
suneettipirneni, line 12: We can make this a bit more specific in order to prevent the cast on line 89 in `src/common/bot.ts`
```suggestion
  execute: z
    .function()
    .args(z.any()) // Add more specific check later?
    .returns(z.union([z.void(), z.promise(z.void())])),
```

suneettipirneni (changes requested):

src/app/listener.ts
```diff
@@ -42,6 +44,15 @@ export class Listener {
 
       this.container.loggerService.info(`Loaded ${this.container.jobService.size()} jobs...`);
 
+      const slashCommandUploads = Array.from(slashCommands.entries()).map(([key, command]) => {
+        return {
+          name: key,
+          description: command!.description.substring(0, 99),
```
suneettipirneni, line 50:
```suggestion
          description: command.description.substring(0, 99),
```

suneettipirneni (changes requested):

rob-3:
This is pretty good as is. I'm not super happy with the parameters just being passed straight through (it's not easy to know how they work without reading the Discord docs), but this PR has to stop somewhere and I think it's in a good enough state now.

suneettipirneni (approved):

suneettipirneni:
> This is pretty good as is. ~~I'm not super happy with the parameters just being passed straight through (it's not easy to know how they work without reading the Discord docs)~~, but this PR has to stop somewhere and I think it's in a good enough state now.
> 
> 
> 
> Edit: I'm just going to remove the parameters code for now; I think someone should think about what, if any abstractions we want to provide over the native Discord API for these.

Personally, I don't think an abstraction is needed here. The payload options are pretty straightforward and readable as-is. I don't think there would be any benefit to be had if they were abstracted.

rob-3:
I don't necessarily disagree, I just haven't thought carefully about it.

rob-3:
I restored the field, but renamed it to `options` for more consistency with Discord.

src/app/bot.ts
```diff
@@ -69,6 +71,30 @@ export class Bot {
...
+        if (result.success) {
+          slashCommands.set(plugin.commandName, result.data as ISlashCommand);
+        } else {
+          this.container.loggerService.warn(
+            `${file} does not \`export default\` a plugin with type ISlashCommand!`
+          );
+          this.container.loggerService.warn(result.error);
+        }
+      });
+    });
```
tanndlin, line 97: This needs to be done somewhere else, or change the code flow of the instantiation of the `Listener`

The `Listener` is called in the [constructor](https://github.com/cs-discord-at-ucf/lion/blob/8ade0f3a100be08e4edf81acff8c484ee2e81b99/src/app/bot.ts#L25) of `Bot`
Which is before `_loadAndRun` is called by the `run` function

Listener is created [here](https://github.com/cs-discord-at-ucf/lion/blob/8ade0f3a100be08e4edf81acff8c484ee2e81b99/src/index.ts#L7) which tries to loop over the (currently empty) `slashCommands` map
`_loadAndRun` is called [here](https://github.com/cs-discord-at-ucf/lion/blob/8ade0f3a100be08e4edf81acff8c484ee2e81b99/src/index.ts#L8C1-L8C1) which populates the `slashCommands` map

tanndlin (changes requested):

src/app/bot.ts
rob-3: Hmm, where would you prefer? The code works in its current form. I bypassed several important-sounding classes with `Handler` and `Listener` in them because I just couldn't understand what they were for.

rob-3: I thought this would be the right place because this is where the regular plugins get set up.

tanndlin: the latter would be easier, if possible

rob-3: This is odd, because I just tested this and the upload appears to work as intended. I deleted 8ball (the only slash command rn) and reran and there was nothing. Then, I added 8ball back and reran and it was there at `/8ball`

rob-3: I wonder if it's a permissions thing? You need to [grant the slash commands permission](https://discordjs.guide/preparations/adding-your-bot-to-servers.html#creating-and-using-your-invite-link) to the bot in your server

rob-3: My reproduction:
```
git switch suneettipirneni/feature/slash-commands
npm ci
npm run watch
```

src/app/listener.ts
```diff
@@ -42,6 +44,20 @@ export class Listener {
 
       this.container.loggerService.info(`Loaded ${this.container.jobService.size()} jobs...`);
 
+      const slashCommandUploads = Array.from(slashCommands.entries()).map(([key, command]) => {
```
tanndlin, line 47: On second thought this is better off being right after loading from the file system, [here](https://github.com/rob-3/lion-clean-room/blob/360f1c35f24ea8b3690f12cdb9a5af94a5328eff/src/app/bot.ts#L77-L99)

This doesn't really have anything to do with the listener per se and would fix the race condition problem

tanndlin (changes requested):

src/common/slash.ts
```diff
@@ -0,0 +1,30 @@
...
+  name: z.string(),
+  description: z.string(),
+  options: z.optional(z.array(z.any())),
+  execute: z
+    .function()
+    .args(z.any())
+    .returns(z.union([z.void(), z.promise(z.void())])),
+});
+
+export interface ISlashCommand {
```
joeydotdev, line 18: is there a reason why we couldn't just infer the typing from the zod object?

src/app/listener.ts
rob-3: Ok, I will move it.

src/common/slash.ts
rob-3: The Zod typings for the `execute` method and `options` use Discord.js types that aren't in Zod and would be a pain to move over, and aren't fully descriptive. I could change to an inference + an intersection type, but I'm not sure that's better.

joeydotdev: Nah don't worry about moving over. that sounds like too much effort lmao

PR #769, opened 2023-08-30
Simplify initialization code

rob-3:
There are several methods and files (most notably `Kernel.ts`) that don't provide much value and just complicate the boot startup process. This PR removes a few files and moves almost all of the startup logic into `index.ts` to make it easier to figure out what order things happen in.


src/index.ts
```diff
@@ -1,8 +1,136 @@
...
+      container.loggerService.info('Webserver is now running')
+    );
+
+    webServer.get('/health', (_, res) => res.send('OK'));
+
+    container.loggerService.info('Bot loaded.');
+    while (true) {
+      const waiting = new Promise((resolve) => setTimeout(resolve, 1_000_000_000));
+      await waiting;
+    }
```
suneettipirneni, line 132: Is this still needed even though we're in an `async` context? If so, there probably needs to be some code refactored so it can be `await`ed in this block.

tanndlin (approved):
if its tested and runs the same, then its good with me

we could use some trimming of old abstractions

src/index.ts
rob-3: Not sure. I didn't touch it because 99% of this PR is just refactoring without any logic changes. The only thing I did remove was some code that purportedly restarted the web server in case of something, but never actually could be called (`91e9bfb`)

rob-3:
> if its tested and runs the same, then its good with me
> 
> we could use some trimming of old abstractions

It appears to work the same for me. The main thing I did was just inline functions

src/index.ts
```diff
@@ -1,8 +1,136 @@
...
+    // Register commands for all guilds.
+    await Promise.all(
+      container.clientService.guilds.cache.map((guild) => guild.commands.set(slashCommandUploads))
+    );
+
+    // register jobs
+    container.jobService.reset();
+
+    const jobs = container.jobService.jobs;
+    for (const job of jobs) {
```
c650, line 110: can you make this loop consistent w the stores loop? one is for-of the other is forEach

rob-3: Done

tanndlin:
deepscan

rob-3:
If this can be prioritized for review I would appreciate it, because keeping up with y'all's changes to `bot.ts` when I refactored it across files is going to get old fast :)

tanndlin:
Sorry it was failing a deepscan for an unused import

rob-3:
No worries!

src/index.ts
```diff
@@ -1,8 +1,138 @@
...
+  const webServer = express();
+  try {
+    container.loggerService.info('Loading and running Bot...');
+
+    container.clientService.on('ready', async () => {
+      container.pluginService.reset();
+
+      // load classic plugins
+      const pluginFolder = path.join(__dirname, '/plugins');
+      const pluginFiles = await readdir(pluginFolder);
```
tanndlin, line 30: My checkout is crashing on this line
```
error: Could not read plugins folder: Error: ENOENT: no such file or directory, scandir 'C:\Users\Tanner\Desktop\Projects\Github\lion\src\plugins'
```
The `__dirname` mustve changed since we are in a different file now. The path should be `lion\src\app\plugins`

tanndlin (changes requested):

src/index.ts
rob-3: Oh good catch, I didn't notice this when remerging it

rob-3: I wish there was a "run a command" e2e test that would catch stuff like this. All the existing tests passed despite this

tanndlin (approved):

PR #771, opened 2023-09-01
refactor: use vitest instead of jest

suneettipirneni:
Removes jest in favor of vitest.

Jest is consdirably more bloated and slower than vitest. Vitest offers almost the exact API as jest so the migration is straightforward.


rob-3:
I can vouch for vitest but don't have time to review this unfortunately. I'm not a maintainer either

c650 (approved):

tanndlin (approved):

PR #775, opened 2023-09-03
Migrate the pet plugins to slash commands

tanndlin:
Almost none of the logic changed. I only moved functions around to get out of the OOP class system from before.

Tested and works


tanndlin:
Waiting for #775

src/app/slash_plugins/dog.plugin.ts
```diff
@@ -0,0 +1,151 @@
...
+  options: [
+    {
+      name: 'breed',
+      description: 'The breed of dog to generate. Or "list" to list all breeds.',
+      type: 'STRING',
+      required: false,
+    },
+  ],
+  async execute({ interaction, container }) {
+    if (allBreeds.size === 0) {
```
suneettipirneni, line 24: Discord commands by default only have a window of 3 seconds to respond or else the response becomes invalid. In the case of this plugin, due to unexpected network latency, we may not be able to guarantee this. Instead we can defer `await interaction.defer()` to tell discord to extend the response window to 15 minutes.

You'll also need to update your `interaction.reply`s to `interaction.followUp` since we're telling discord we're updating a deferred interaction rather than a normal one.

```suggestion
  async execute({ interaction, container }) {
  	await interaction.defer();
    if (allBreeds.size === 0) {
```

suneettipirneni (changes requested):

src/app/slash_plugins/dog.plugin.ts
tanndlin: Ok awesome good to know

src/app/slash_plugins/dog.plugin.ts
```diff
@@ -0,0 +1,154 @@
...
+        breed.subBreed.forEach((subBreed) => allBreeds.add(subBreed));
+      });
+
+      // Sorting the subBreed list so the embed looks better
+      subBreeds.sort((a: IDogSubBreed, b: IDogSubBreed) => a.subBreed.length - b.subBreed.length);
+    })
+    .catch((err) => container.loggerService.warn(err));
+};
+
+interface IDogSubBreed {
```
c650, line 149: move this declaration to the top of the file

c650 (approved):
its funny how the coding style of our slash commands is so different from everything else.

PR #780, opened 2023-09-05
Convert `tictactoe` plugin to a slash command

suneettipirneni:
We no longer need reactions to handle interactions from the user, we can now use buttons. In addition, invalid moves are reported via an ephemeral message.

This has been tested and works.

Screenshot 2023-09-05 at 3 05 34 PM
Screenshot 2023-09-05 at 3 06 26 PM


suneettipirneni:
CI Failing for reasons outside this PR

tanndlin (approved):

PR #781, opened 2023-09-05
convert `connectfour` to a slash command

suneettipirneni:
Tested locally and works as expected.


tanndlin (approved):
I like this. a lot less jank

joeydotdev (approved):

PR #782, opened 2023-09-06
fix error logging

suneettipirneni:
Often whenever an error was hit in lion, the program would just exit and not provide any stack trace or context as to what happened. This was due to unhandled promise rejections being ignored. In NodeJS, `EventEmitter`s by default don't handle rejected promises which causes them to go into the void. Luckily there's an option to avoid this behavior [as documented here](https://nodejs.org/api/events.html#capture-rejections-of-promises). This re-routes all rejected errors to an `'error'` event which we can use to log promise rejections. 

This PR also improves some of the aesthetics of the console output to make it easier to read.


tanndlin (approved):
pog

PR #783, opened 2023-09-07
support option autocomplete support for slash commands

suneettipirneni:
This adds a new method `autocomplete` to `ISlashCommand`. This method is invoked whenever an autocomplete-able option is being typed by a user. The method simply responds with the autocomplete results for the user to use when typing a command option. 

More information on this feature can be found in the [docs](https://discordjs.guide/slash-commands/autocomplete.html#autocomplete).


tanndlin (approved):

c650 (commented):
can you add the fcn to one command so that reviewers can see how itll be implemented?

suneettipirneni:
> can you add the fcn to one command so that reviewers can see how itll be implemented?

An implementation of this can be found [here](https://github.com/cs-discord-at-ucf/lion/blob/f86f8afca27e11ec29701ec83363c88879212f69/src/app/slash_plugins/register.plugin.ts#L148)

tanndlin:
conflicts btw

PR #784, opened 2023-09-08
Convert gamble to slash command

tanndlin:


src/app/slash_plugins/gamble.plugin.ts
```diff
@@ -0,0 +1,98 @@
...
+  description: 'Bet your Tacos on a coin flip',
+  options: [
+    {
+      name: 'amount',
+      description: 'The amount of Tacos to gamble',
+      type: 'STRING',
+      required: true,
+    },
+  ],
+  execute: async ({ interaction, container }) => {
```
suneettipirneni, line 19: Small nit, other plugins use the `async execute(…)` style.

src/app/slash_plugins/gamble.plugin.ts
```diff
@@ -0,0 +1,98 @@
...
+      return;
+    }
+
+    await interaction.deferReply();
+
+    const userDoc = await container.pointService.getUserPointDoc(interaction.user.id);
+    const pointsToGamble = arg.toLowerCase() === 'all' ? userDoc.numPoints : parseInt(arg);
+
+    if (userDoc.numPoints === 0) {
+      await interaction.followUp('You have no Tacos to gamble!');
```
suneettipirneni, line 32:
```suggestion
      await interaction.followUp({ content: 'You have no Tacos to gamble!', ephemeral: true });
```

suneettipirneni (changes requested):

src/app/slash_plugins/gamble.plugin.ts
tanndlin: oh ok yeah i want to stick to the same pattern. just didnt realize you could do that

src/app/slash_plugins/gamble.plugin.ts
tanndlin: I need to remember that is an option lol

suneettipirneni (approved):

PR #785, opened 2023-09-09
Convert class commands to slash commands

suneettipirneni:
This converts `unregister`, `register` and `listclasses` to slash commands.

For `/register` and `/unregister` all messages are ephemeral so other can't see what classes you are and both provide autocomplete-able options, for example `/register`:

https://github.com/cs-discord-at-ucf/lion/assets/77477100/fa880ce5-5117-43a5-96d9-4d45ecf23304

`/unregister` is similar except it only shows classes you have already registered for.

Because we now have ephemeral message I wanted to have `/listclasses` avoid needing to use DMs. I did this by splitting the message up by category. The message now will have a select menu at the bottom to switch between different class categories:

Screenshot 2023-09-08 at 10 56 44 PM

Because this PR uses autocomplete it is blocked by #783


c650:
>For /register and /unregister all messages are ephemeral so other can't see what classes you are 

do we want it to be this way? what do people think? probably the answer is yes, just pulling it out for discussion

src/services/message.service.ts
```diff
@@ -116,16 +145,20 @@ export class MessageService {
...
           return;
         }
 
         if (options.cutoffMessage) {
-          await msg.edit(options.cutoffMessage);
           if (typeof options.cutoffMessage === 'string') {
-            await msg.suppressEmbeds(true);
+            await buttonInteraction.update({ content: options.cutoffMessage, components: [] });
+          } else {
+            // @ts-ignore
```
c650, line 160: ?

src/services/class.service.ts
```diff
@@ -56,36 +56,37 @@ export class ClassService {
...
 
-      embeddedMessage.setDescription(
-        `Did you mean \`${similarClassID}\`?\n` +
-          `React with ✅ to ${action} for this class.\n` +
-          'React with ❎ to close this offering.'
-      );
+      embeddedMessage.setDescription(`Did you mean \`${similarClassID}\`?\n`);
 
       // EmojiData acts as a key.
       emojiData.push({
```
c650, line 80: can you demo what this looks like?

src/services/class.service.ts
```diff
@@ -56,36 +56,37 @@ export class ClassService {
...
   public getSimilarClasses(
-    message: types.IMessage,
+    interaction: discord.CommandInteraction,
     invalidClasses: string[],
     action: 'register' | 'unregister'
-  ): types.IEmbedData[] {
+  ): types.IInteractionEmbedData[] {
     return invalidClasses.map((invalidClass: string) => {
-      const emojiData: types.IEmojiTable[] = [];
+      const emojiData: types.IButtonLabelTable[] = [];
```
c650, line 64: does `emojiData` seem like it's still a good variable name ?

c650 (approved):
i only had nits

maybe paste more demo clips of other features; at least the "Did you mean X?" flow looks like it changed a bit.

src/services/class.service.ts
suneettipirneni: Here's the demo:
https://github.com/cs-discord-at-ucf/lion/assets/77477100/aa73a55b-5708-4591-99dc-8c0368c2445d

src/services/class.service.ts
```diff
@@ -56,40 +56,41 @@ export class ClassService {
...
       const [similarClassID] = this.findSimilarClasses(invalidClass);
 
       // TODO check if similarity is close then decide whether to return the guess or tell them to get a mod.
 
-      embeddedMessage.setDescription(
-        `Did you mean \`${similarClassID}\`?\n` +
-          `React with ✅ to ${action} for this class.\n` +
-          'React with ❎ to close this offering.'
-      );
+      embeddedMessage.setDescription(`Did you mean \`${similarClassID}\`?\n`);
```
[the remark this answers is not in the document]
suneettipirneni: Unfortunately yes, autocomplete doesn't force you to choose an option it just suggests, people can still put in any arbitrary values. It probably won't be likely they do this, but this flow is still here just in case that happens.

tanndlin (approved):

PR #788, opened 2023-09-14
convert taco leaderboard to slash command

suneettipirneni:


c650 (changes requested):
you and @tanndlin both did !give ?

tanndlin (approved):

PR #790, opened 2023-09-16
Convert Checkclasses plugin to Slash command

tanndlin:


c650 (approved):

tanndlin:
I think we do this once #792 is finished
then make the infrastructure for context menus

PR #791, opened 2023-09-16
Convert Add/Del roles to slash commands

tanndlin:
2 things:

1. I wasnt entirely sure how to multiple roles at once, but most people do them one by one anyways. also keeps the logic simpler

2. I couldnt not find a way to prevent certain roles from showing up in the autocomplete. The sensitive ones are still not able to be added since Lion's role in below them, but the user is still able to try, albeit fail


c650:
> 2. I couldnt not find a way to prevent certain roles from showing up in the autocomplete. The sensitive ones are still not able to be added since Lion's role in below them, but the user is still able to try, albeit fail

so long as we fail gracefully it's fine

c650 (approved):

PR #792, opened 2023-09-16
Add defaultMemberPermissions to slash commands

tanndlin:
Brings back the minRoleToRun functionality of legacy plugins

just a secondary precaution to prevent sensitive commands being used

### DIFF
--- a/src/common/slash.ts
+++ b/src/common/slash.ts
@@ -14,7 +14,7 @@ export const SlashCommand = z.object({
   name: z.string(),
   description: z.string(),
   options: z.optional(z.array(z.any())),
-  minRoleToRun: z.optional(z.any()),
+  defaultMemberPermissions: z.optional(z.any()),
   execute: z
     .function()
     .args(z.any())
@@ -38,7 +38,7 @@ export interface ISlashCommand {
   name: string;
   description: string;
   options?: ApplicationCommandOptionData[];
-  minPermissions?: PermissionResolvable;
+  defaultMemberPermissions?: PermissionResolvable;
   execute({
     interaction,
     container,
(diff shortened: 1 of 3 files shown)


suneettipirneni:
We can modify our slash command structure to have a defaultPermission field. This tells discord the minimum permissions a user needs to run a command, and prevents any unsafe window between a command being deployed and an admin setting the perms.

tanndlin:
> We can modify our slash command structure to have a defaultPermission field. This tells discord the minimum permissions a user needs to run a command, and prevents any unsafe window between a command being deployed and an admin setting the perms.

oh ok perfect, wasnt sure if that was implemented on the API side since the server can set them manually as well

src/common/slash.ts
```diff
@@ -36,6 +38,7 @@ export interface ISlashCommand {
   name: string;
   description: string;
   options?: ApplicationCommandOptionData[];
+  minPermissions?: PermissionResolvable;
```
suneettipirneni, line 41: We can probably use the same discord name for it as it's a bit more accurate.
```suggestion
  defaultMemberPermissions?: PermissionResolvable;
```

src/common/slash.ts
```diff
@@ -13,6 +14,7 @@ export const SlashCommand = z.object({
   name: z.string(),
   description: z.string(),
   options: z.optional(z.array(z.any())),
+  minRoleToRun: z.optional(z.any()),
```
suneettipirneni, line 17: This will need to be renamed to the new field name.

suneettipirneni (changes requested):

suneettipirneni (approved):

PR #793, opened 2023-09-17
add support for context menu commands

suneettipirneni:
This PR changes the types for commands in general. Instead of commands satisfying the `ISlashCommand` type they instead satisfy the `Command` type. This was done for a few reasons, firstly slash commands are specific version of a discord application command. Secondly, having a tagged union `Command` type gives better type inference for user's creating commands. For example, in the default export when `type: 'USER'` is specified typescript automatically narrows the type of `interation` in `execute` to be a `UserContextMenuInteraction` instead of a `CommandInteraction`. 

Creating context menu commands is almost exactly the same as creating slash commands. The only difference is the optional `type` field is `'USER'` for user context menus or `'MESSAGE'` for message context menus. Here's an example of a message context menu command that echoes the target message:

```ts
import { Command } from '../../common/slash';

export default {
  type: 'MESSAGE',
  name: 'echo',
  commandName: 'echo',
  async execute({ interaction }) {
    await interaction.reply({ content: interaction.targetMessage.content });
  },
} satisfies Command;
```

And here's how the command looks to the user:

https://github.com/cs-discord-at-ucf/lion/assets/77477100/cf409098-acf4-4b06-893f-5baca8d0d007

**One very important limitation of using context menu commands is that the bot is only limited to having 5 of them**


src/app/listener.ts
```diff
@@ -77,22 +77,38 @@ export class Listener {
...
-        ?.execute({ interaction, container: this.container });
+      if (interaction.isUserContextMenu() && command?.type === 'USER') {
+        command?.execute({ interaction, container: this.container });
+        return;
+      }
+
+      if (interaction.isMessageContextMenu() && command?.type === 'MESSAGE') {
+        command?.execute({ interaction, container: this.container });
+        return;
+      }
```
tanndlin, line 111: Do these need to be spread out? like whats the use case where we care where the command came from

src/index.ts
```diff
@@ -89,12 +90,13 @@ EventEmitter.captureRejections = true;
...
 
-      const slashCommandUploads = Array.from(slashCommands.entries()).map(([key, command]) => {
+      const slashCommandUploads = Array.from(commands.entries()).map(([key, command]) => {
         return {
+          type: command.type,
           name: key,
-          description: command.description.substring(0, 99),
-          options: command.options,
-        };
+          description: 'description' in command ? command.description.substring(0, 99) : undefined,
```
tanndlin, line 97: Does this not work with just 
```
command.description?.substring(0, 99)
```
does that not do the same, but return undefined if its not specified?

src/index.ts
```diff
@@ -89,12 +90,13 @@ EventEmitter.captureRejections = true;
...
+          options: 'options' in command ? command.options : undefined,
```
tanndlin, line 98: saa

src/index.ts
suneettipirneni: These are mainly here to make the ts compiler happy, doing `command.description` reports an error because a context menu has no description.

suneettipirneni: Actually I found a way around this.

tanndlin (approved):

PR #794, opened 2023-09-17
Convert Add/Del chans to slash commands

tanndlin:


tanndlin:
These are slightly different from before because of the confirmation flow

now its a button that appears on the message to confirm or cancel the addition/deletion

that way its still a 2 step process

src/app/slash_plugins/addclasschans.plugin.ts
```diff
@@ -0,0 +1,238 @@
...
+  description: 'creates a bunch of class channels',
+  defaultMemberPermissions: 'ADMINISTRATOR',
+  options: [
+    {
+      name: 'category',
+      description: 'The category of the class',
+      type: 'STRING',
+      required: true,
+      choices: [
+        { name: 'Computer Science', value: 'cs' },
```
c650, line 33: should this list be factored out? it's the same as in some other plugin

src/app/slash_plugins/addclasschans.plugin.ts
```diff
@@ -0,0 +1,238 @@
...
+    await promptUser(interaction, container, parsedClasses);
+    // if (args[0] === 'confirm') {
+    //   await _proceedToAddClasses(message);
+    // } else if (args[0] === 'cancel') {
+    //   await _proceedToCancel(message);
+    // } else {
+    //   await _promptUser(message, parsedClasses);
+    // }
+
+    // Add buttons to response to replace old functionality
```
c650, line 76: delete

c650 (approved):
you tested?

src/app/slash_plugins/addclasschans.plugin.ts
tanndlin: Yeah this was done in one other place. I changed the code to copy that.

I dont think its worth factoring out since its a one liner.

tanndlin:
> you tested?

yup works good

c650 (approved):

PR #800, opened 2023-09-18
tell users about slash command if prefix command is used

suneettipirneni:
This sends a message to the user if a prefix command (with the same name as a slash command) is used telling them to use the slash command version instead. Demo:

https://github.com/cs-discord-at-ucf/lion/assets/77477100/c81df0ce-c987-410e-9fe4-27a5958a4829

### DIFF
--- a/src/app/handlers/command.handler.ts
+++ b/src/app/handlers/command.handler.ts
@@ -44,14 +44,12 @@ export class CommandHandler extends Handler {
     if (slashCommand) {
       const cachedCommand = this.container.guildService
         .get()
-        .commands.cache.find((c) => c.name === command.name);
+        .commands.cache.find((c) => c.name === command.name)!;
 
-      if (cachedCommand) {
-        await message.reply(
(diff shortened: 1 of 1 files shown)


src/app/handlers/command.handler.ts
```diff
@@ -37,6 +38,22 @@ export class CommandHandler extends Handler {
...
 
+    // Check if a slash command of the same name exists, if so redirect to it
+    const slashCommand = commands.get(command.name);
+
+    if (slashCommand) {
+      const cachedCommand = this.container.guildService
+        .get()
+        .commands.cache.find((c) => c.name === command.name);
+
+      if (cachedCommand) {
```
tanndlin, line 49: Whats the use case where we would need to double check that the slash command is in the fs **and** thats its deployed?

Shouldnt every thing in that map be deployed?

tanndlin (commented):
this is pog

tanndlin (approved):

PR #802, opened 2023-09-18
convert alt and leaderboard to slash commands

suneettipirneni:


src/app/slash_plugins/leaderboard.plugin.ts
```diff
@@ -0,0 +1,103 @@
...
+  gameEnum: GameType
+) {
+  return container.gameLeaderboardService.createMatchupLeaderboardEmbed(
+    interaction.user,
+    opponent,
+    gameEnum
+  );
+}
+
+function getMatchUpEmbed(
```
tanndlin, line 21: one of these is marked async, one is not. 
you should just specify the result type to be Promise<MessageEmbed> for both instead

src/app/slash_plugins/leaderboard.plugin.ts
```diff
@@ -0,0 +1,103 @@
+import { CommandInteraction, User } from 'discord.js';
+import { Command } from '../../common/slash';
+import { IContainer } from '../../common/types';
+import { GameType } from '../../services/gameleaderboard.service';
+
+const gameTypes = ['TicTacToe', 'ConnectFour'];
```
tanndlin, line 6: We should probably make this an enum instead of an array to avoid the gymanstics of getting the index and using that

src/__tests__/alt.test-old.ts
```diff
@@ -0,0 +1,60 @@
...
+//       'add\n465026352777789441\n465026352777789441\nn465026352777789441',
+//       'add\n465026352777789441',
+//       'add\n465026352777789441\n',
+//     ],
+//   ])('Add Tests: Should be false when arg is: %s', (input) => {
+//     expect(
+//       new AltPlugin(getContainerMock()).validate(getMessageMock(), input.split(' '))
+//     ).toBeFalsy();
+//   });
+// });
```
tanndlin, line 60: Let's just delete this

if we need it later we can checkout an older commit

tanndlin (changes requested):
sorry for the delay

c650 (approved):

tanndlin (approved):

PR #803, opened 2023-09-19
Feat/anonreport slash

tanndlin:


c650 (changes requested):
this diff is huge. did you mess up a rebase?

c650:
also will anonreport still be a dm?

tanndlin:
> this diff is huge. did you mess up a rebase?

Good catch, I need to rebase the base branch with develop

> also will anonreport still be a dm
No I made it ephemeral for all responses when reporting

c650:
> > also will anonreport still be a dm
> 
> No I made it ephemeral for all responses when reporting

if it's not a dm and someone types it wrong, like putting a ! instead of a / , itll get sent to whatever channel, right? minor concern

src/app/slash_plugins/anonreport.plugin.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (subcommand === 'respond') {
+      if (!isModerator) {
+        return interaction.reply({
+          content: 'You must be a moderator to use this command.',
+          ephemeral: true,
+        });
+      }
+
+      const ticket_id = interaction.options.getString('ticket_id');
+      if (!ticket_id) {
```
c650, line 82: is this validation necessary, since we spec that "respond" has a `required` ticket_id param?

c650 (approved):
code lgtm. 

still have the concern about the fundamental change from this being a dm command to it being in a pub channel

src/app/slash_plugins/anonreport.plugin.ts
tanndlin: I can assert it non null, for whatever reason it comes as nullable from the API even though its required

tanndlin:
> if it's not a dm and someone types it wrong, like putting a ! instead of a / , itll get sent to whatever channel, right? minor concern

Hmm I'm not 100% sure on the way to go about this. It would be nice to completely move away from DMs since that was always a hassle

if we remove all DMs we dont have to worry about catching when the user has messages from non friends blocked, etc

c650 (approved):

joeydotdev (approved):
🚢

PR #806, opened 2023-09-21
store the static breed information for cat/dog plugins

suneettipirneni:
Instead of fetching the breed data whenever the plugin is loaded we store the data locally and reference it. This allows us to use autocomplete for breeds.


tanndlin (approved):

c650 (commented):
how will this stay up to date?

suneettipirneni:
> how will this stay up to date?

Either someone will ping us and tell us there's an issue or an issue ticket will be made. I don't think the breeds lists are updated very frequently at all.

c650 (approved):

PR #811, opened 2023-09-29
feat: Convert ta to slash command

tanndlin:
This isn't *really* a staff command but I'm editting the addclasschans welcome message to reflect the changes in this PR
so its easier to merge here


src/app/slash_plugins/ta.plugin.ts
```diff
@@ -0,0 +1,259 @@
...
+
+export interface ITAEntry {
+  userID: Snowflake;
+  chanID: Snowflake;
+  guildID: Snowflake;
+}
+
+export type TADocument = ITAEntry & Document;
+
+const ALLOWED_ROLES = [Constants.Roles.TeachingAssistant, Constants.Roles.Professor];
```
suneettipirneni, line 23: This gating should be configured by the server admins and not the developers. The admins can go into the integrations and set roles that are and aren't allowed to use this rather than it being hardcoded.

For the time being, we can just set the `defaultMemberPermissions` to `'ADMIN'` so no one untrusted can use the command in the meantime.

Ideally, in the future we don't have to hardcode roles at all.

suneettipirneni (changes requested):

c650 (approved):
lgtm pending suneet

PR #824, opened 2023-12-16
Fixed casing missmatch

CallumCoy:
Individual breeds were never found.  Due to a mismatch on the casing.  This sets the breeds to lowercase to match the user input.

### DIFF
--- a/src/app/commands/cat.command.ts
+++ b/src/app/commands/cat.command.ts
@@ -46,8 +46,8 @@ const command: Command = {
     let searchCom = '';
 
     const breedIn = interaction.options.getString('breed')?.toLowerCase() ?? 'random';
-    // checks if their was a bread was a breed, then if that breed is recognized
-    const breedEntry = ALL_BREEDS.find((breed) => breed.name === breedIn);
+    // checks if their was a breed was a breed, then if that breed is recognized
+    const breedEntry = ALL_BREEDS.find((breed) => breed.name.toLowerCase() === breedIn);
 
     if (breedEntry !== undefined) {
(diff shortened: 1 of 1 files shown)


src/app/commands/cat.command.ts
```diff
@@ -46,8 +46,8 @@ const command: Command = {
     let searchCom = '';
 
     const breedIn = interaction.options.getString('breed')?.toLowerCase() ?? 'random';
-    // checks if their was a bread was a breed, then if that breed is recognized
-    const breedEntry = ALL_BREEDS.find((breed) => breed.name === breedIn);
+    // checks if their was a breed was a breed, then if that breed is recognized
+    const breedEntry = ALL_BREEDS.find((breed) => breed.name.toLowerCase() === breedIn);
```
c650, line 50: there is a lower-cased breedNames list at the top of the file

c650 (changes requested):

src/app/commands/cat.command.ts
CallumCoy: That has list added to it though, which isn't a breed,  unless you want me to check against list specifically.

CallumCoy: Oh wait, disregards you're talking about query

c650: the `list` case is handled on 36

CallumCoy: Can't use breedNames or query.  they are arrays, we need the full set since we need the ID for the breed.

c650 (approved):

tanndlin (approved):

PR #846, opened 2024-11-16
Fix garage command not working

msabur:
Updated the `/garage` command for the current API. 
Info for the Libra garage is no longer included, and the occupancy count can sometimes exceed the total count.

An output
IMG_2692


c650 (approved):
tysm!

msabur:
> tysm!

You’re welcome!

PR #853, opened 2025-04-29
Fix pluralization in success message for /unregister command

msabur:
Adjusts success message for /unregister to use "class" or "classes" depending on count. Previously it would always say "classes".

image


tanndlin (approved):
Sick, thanks

Clev7:
I object to this change!

